PR #6, opened 2018-09-03
add web3-provider package

pedrouid:
[Work in Progress]
- Still not matching v0.6.0 spec changes
- webpack config needs to be replaced by rollup
- README is incomplete


jdkanani:
Here you need webpack to bundle. @shahbhavir @pedrouid  web3-provider is browser lib where one can use `<script src="web3-provider" />` and then use `web3 = new Web3(window.walletconnectProvider)`. Thoughts?

pedrouid:
Since we are using rollup, let's keep it consistent for this monorepo

PR #26, opened 2018-09-23
v0.7.0 [Monorepo]

pedrouid:
**Checklist:**
- [x] add sendTransaction
- [x] add signMessage
- [x] add signTypedData
- [x] refactor transaction requests to call requests
- [x] merge web3-subprovider
- [ ] improved error messages
- [ ] add APN support

**Requires:** 
- [ ] walletconnect/py-walletconnect-bridge#38
- [ ] walletconnect/py-walletconnect-push#1
- [ ] walletconnect/node-walletconnect-bridge#3
- [ ] walletconnect/node-walletconnect-push#1


pedrouid:
Requires #17

README.md
```diff
@@ -61,17 +61,57 @@ if (session.new) {
...
+]
 
-if (transactionStatus.success) {
-  const { txHash } = transactionStatus // Get transaction hash
+/**
+ *  Sign Typed Data
+ */
+try {
+  // Signed typed data
+  const result = await webConnector.signTypedData(msgParams)
```
jdkanani: Wouldn't that be great if we keep compatible to ETH rpc instead of creating new method for different method.

Example: 

```js
webConnector.sendCallRequest({
   id: 1,
   method: 'eth_sendTransaction',
   params: [{
      from: '0xab12...1cd',
      to: '0x0',
      nonce: 1,
      gas: 100000,
      value: 0,
      data: '0x0'
   }]
})
```

and
 
```js
const msgParams = [
  {
    type: 'string',
    name: 'Message',
    value: 'My email is john@doe.com'
  },
  {
    type: 'uint32',
    name: 'A number',
    value: '1537836206101'
  }
]

webConnector.sendCallRequest({
   id: 1,
   method: 'eth_signedTypedData',
   params: [msgParams]
})
```

That way we don't need to take care of each type of ETH request, standard will be more generic and open.

pedrouid: Absolutely! I'm currently changing it on the web3-provider branch as I quickly found it to be easier to match 1:1 the JSON RPC request spec.

I will be updating this PR soon. Thanks for the feedback 🙏

PR #27, opened 2018-09-26
add toJSON serialization to save session in app

jdkanani:


pedrouid:
@jdkanani I moved it to js-walletconnect-core because this is also shared with the browser-sdk and also added expires

PR #35, opened 2018-10-02
Fix approveSession data

jdkanani:
Approve session like - 
```
approveSession(["0x123...", ... ])
```

or need to change `getSessionStatus` in `walletconnect/walletconnect`

### DIFF
--- a/packages/walletconnect-react-native/src/index.js
+++ b/packages/walletconnect-react-native/src/index.js
@@ -11,7 +11,7 @@ export default class WalletConnector extends Connector {
   // approve session
   //
   async approveSession(data) {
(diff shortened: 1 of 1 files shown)


pedrouid:
I see what you mean but it's better to update `getSessionStatus` to make it consistent through encryptionPayloads

pedrouid:
Actually I just realised that is not necessary to expose the variable approved as well, because it's not used by the bridge at all

jdkanani:
There are multiple inconsistencies. Maybe we should fix the standard of data passing between client-bridge-rn and then follow it.

pedrouid:
That's what Im doing right now

pedrouid:
duplicate #36

PR #36, opened 2018-10-02
V0.7.3

pedrouid:


pedrouid:
Requires WalletConnect/py-walletconnect-bridge#45
Requires WalletConnect/node-walletconnect-bridge#8

pedrouid:
Check this PR @jdkanani 

Now it's more consistent on requests and passing encryptionPayloads and should fix the other two PRs you had submitted

There is definitely inconsistencies with the node-bridge and py-bridge implementations. I try to keep them up-to-date as much possible but I'm giving priority to the py-bridge which is more battle tested and easier to setup using the quick Docker setup

PR #40, opened 2018-10-12
Trying to fix web3 provider. 

digitaldonkey:
Part one...


digitaldonkey:
I got a different linter obviously it applies the eslinrc from my vue app.

Issue #38, opened 2018-10-10
How to include the source files in my project? (v0.8.x)


digitaldonkey:
When I use walletconnect-web3-provider as stated in the readme, I will get a compiled and minified version in my code. Which is:

* hard to debug
* created a overhead in my own module. I'm using webpack myself and there is the nice codeSplitting feature I want to take advantage of. E.g. the buffer module is part of Web3 and part of walletconnect. If I can include the sources in my build webpack would nicely put the chunks together and I wouldn't force the user to load the same code twice. 

I played around with installing walletconnect-monorepo directly and importing like 
`import WalletConnectProvider from 'walletconnect-monorepo/packages/walletconnect-web3-provider/src'`

Bu this seems pretty strange and I currently don't know how to automatically do the the stuff done by `npm run bootstrap` and `npm run check-packages`

digitaldonkey:
You de-minified the compiled code in 0.78 but that actually doesn't solve anything and could be archived in Crome console too by using pretty print. 

What you need is the actual source code in the subrepo created from the monorepo with sourcemaps pointing there. 

Otherwise debugging WalletConnect is a pain.

digitaldonkey:
The only way I can debug the code unminified within my app is by cloning `walletconnect-monorepo`, run the [dev-workflow](https://github.com/WalletConnect/walletconnect-monorepo/tree/v0.7.x#development-workflow) and applying the following. 
My actual app dependencies is supposed to be `npm install walletconnect-web3-provider` only as shown in [Readme](https://github.com/WalletConnect/walletconnect-monorepo/tree/v0.7.x#for-web3-provider-web3js).

```diff
diff --git a/package.json b/package.json
index 66389b6..1fdd69f 100644
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "walletconnect-monorepo",
+  "version": "0.7.8",
   "description": "Monorepo for WalletConnect Javascript Libraries",
   "private": true,
   "keywords": [
diff --git a/packages/walletconnect-web3-provider/src/index.js b/packages/walletconnect-web3-provider/src/index.js
index 7d48c32..9b2de54 100644
--- a/packages/walletconnect-web3-provider/src/index.js
+++ b/packages/walletconnect-web3-provider/src/index.js
@@ -1,6 +1,6 @@
 import ProviderEngine from 'web3-provider-engine'
[...]

PR #42, opened 2018-10-22
Feature  web3 provider

digitaldonkey:


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "walletconnect-monorepo",
+  "version": "0.7.8",
```
pedrouid: This needs to be removed, this package.json is not a package!!!

I've told you this repository is a wrapper to manage the other packages. There is no such thing a `walletconnect-monorepo` on NPM and there shouldnt be. Please look into monorepos further!

packages/walletconnect-web3-subprovider/src/index.js
```diff
@@ -88,7 +98,7 @@ export default class WalletConnectSubprovider extends Subprovider {
   }
   sendAsync(payload, callback) {
     const next = () => {
-      const sendAsync = this.engine.sendAsync.bind(this)
+      const sendAsync = this._walletconnect.sendAsync.bind(this)
```
pedrouid: Are you still getting an issue with this issue? I've successfully triggered JSON-RPC requests on mobile using the `walletconnect-web3-provider` package using the README instructions. If this issue is related with Walleth specifically, I've contacted @ligi and he said that hasn't updated to the latest spec.

digitaldonkey: Sorry. A leftover from #38

digitaldonkey: The above engine vs _walletconnect I will check.

digitaldonkey:
Now everything should work and be merged.
If it works with your previous code I'm pretty sure it will fail on a slower machine as session is not enforced but only available due to lucky timing.

packages/walletconnect/src/index.js
```diff
@@ -167,7 +167,18 @@ export default class WalletConnect extends Connector {
     const result = await this._getEncryptedData(`/session/${this.sessionId}`)
 
     if (result) {
-      if (result.data.approved) {
+      if (result.data && !result.data.approved) {
```
digitaldonkey: This part is just legacy to be able to use WallEth before upgrade. 
It's not important.

pedrouid:
> Now everything should work and be merged.
> If it works with your previous code I'm pretty sure it will fail on a slower machine as session is not enforced but only available due to lucky timing.

The reason it currently works is because I changed to implementation as you can read the documentation was updated on the [README](https://github.com/WalletConnect/walletconnect-monorepo/blob/master/packages/walletconnect-web3-provider/README.md) for this package.

digitaldonkey:
That might help 👍

What about a **fallback** until Liggi updates WallEth? 
My solution might be shitty, but it alt least allow me to use and test. 
https://github.com/WalletConnect/walletconnect-monorepo/pull/42/files#diff-a2d67c3933f19104ca576fd5707ca80f

Devcon is near and WalletConnect is currently only usable with iPhones - no Android supported.

digitaldonkey:
You might also want to add a try/catch wrapper to the readme. 

At some point the session times out and the user is left without any feedback.

https://github.com/digitaldonkey/web3multisign/blob/65406eeb5aedcc9907b5ed7df8fc1589311843dd/src/components/signerDialogs/WalletConnect.vue#L117-L128

It would be also nice if the arguments on would be respected and carried on.
I've never seen the defaults used, but at least I can't change interval and timeout. 
```listenSessionStatus(interval, timeout)```

pedrouid:
> That might help 👍
> 
> What about a **fallback** until Liggi updates WallEth?
> My solution might be shitty, but it alt least allow me to use and test.
> https://github.com/WalletConnect/walletconnect-monorepo/pull/42/files#diff-a2d67c3933f19104ca576fd5707ca80f
> 
> Devcon is near and WalletConnect is currently only usable with iPhones - no Android supported.

I'm going to work with Ligi this weekend on upgrading to the latest spec so it will be ready before DevCon



> You might also want to add a try/catch wrapper to the readme.
> 
> At some point the session times out and the user is left without any feedback.
> 
> https://github.com/digitaldonkey/web3multisign/blob/65406eeb5aedcc9907b5ed7df8fc1589311843dd/src/components/signerDialogs/WalletConnect.vue#L117-L128
> 
> It would be also nice if the arguments on would be respected and carried on.
> I've never seen the defaults used, but at least I can't change interval and timeout.
> `listenSessionStatus(interval, timeout)`

That's a good point! I will added it directly to the Listener class. Also the defaults are available there:
https://github.com/WalletConnect/walletconnect-monorepo/blob/master/packages/walletconnect-core/src/listener.js#L19

digitaldonkey:
Timeout of 6 sec? I don't think it's true.

It would be just great if I can pass poll interval and timeout on with listenSessionStatus(). 
You don't need to document it, but it would be great if you can override it.

pedrouid:
I see what you mean, that's my bad! You can override from `walletconnect` library but forgot to pass it on `walletconnect-web3-provider`, I will push an update in a second

W3stside:
Hi guys - popping in here without without a ton of background info or context on this PR but does this perhaps fix the `There was a problem connection to Example Dapp` issue? 

We (Gnosis) are currently integrating you guys and I am running into this issue using WallETH on Android. 

If I'm completely wrong just ignore me and continue on. Nice work anyways, seems really promising! :)

pedrouid:
Both me and Ligi got carried away with all the events happening in Prague, so currently WallETH is still on WalletConnect v0.5.x while the Example Dapp is on v0.7.x. Alternatively you can try Balance Wallet on iOS which has published an update today for v0.7.x.

pedrouid:
Several changes have been handled with v0.7.x changes on Example Dapp, WallETH, Balance Wallet and Tokenary Wallet. Web3 Provider future changes will re-visited with Issue #50

PR #54, opened 2018-11-30
Refactor Build Config

### LINKED ISSUE #38: How to include the source files in my project? (v0.8.x)

### LINKED ISSUE #48: Fix Babel Polyfill (v0.8.x)
When used together with libraries `eth-registry` and `dapparatus` the following error occurs on build
```
Error: only one instance of babel-polyfill is allowed
```

### LINKED ISSUE #51: Build Config Refactor (v0.8.x)
There are currently a few issues with the current config, it's sufficient for most cases but it needs improvement.

Required:
- [x] Fix babel polyfill duplicated instance issue (#48)
- [ ] Output unminified ES5 code for debugging (#38)
- [ ] Replace Webpack config with Babel config only

pedrouid:
Fix #51
Fix #48 
Fix #38


pedrouid:
PR will be re-visited with alternative solution. Current implementation breaks on React-Native

PR #64, opened 2018-12-06
Fix personal sign bug

GongHe007:
When send personal sign request, response is invalid, because had delete payload.id. 
```
# requestmanager.js
if (!Jsonrpc.isValidResponse(result)) {
    return callback(errors.InvalidResponse(result));
}
```
Then approveCallRequest in RN, return a object payload like this.
```
{
  result: "",
  approve: true
}
But web3.personal.sign should return a string of signedMessage.

Just solve my problem.

### DIFF
--- a/packages/walletconnect-core/src/connector.js
+++ b/packages/walletconnect-core/src/connector.js
@@ -446,11 +446,10 @@ export default class Connector {
     let payload = this.checkObject(data, 'payload')
 
     if (payload.id) {
-      delete payload.id
+      payload.id = this.randomId()
     }
 
     return {
-      id: this.randomId(),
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/walletconnect-core/src/connector.js [hunk 1]
```diff
@@ -446,11 +446,10 @@ export default class Connector {
     let payload = this.checkObject(data, 'payload')
 
     if (payload.id) {
-      delete payload.id
```
pedrouid: Actually we should just remove the `payload.id` check and always override the `payload.id`. Otherwise if the `payload.id` is missing, we end up with no ID at all

PR #81, opened 2019-03-18
Fix for when an error is returned

Arctek:
Otherwise it causes an error in the console, given payload.result is undefined


pedrouid:
I see what you mean, but let's call an empty `return` instead

```javascript
  if (error) {
    reject(error)
    return
  }
  if (payload.result) {
    resolve(payload.result)
  } 
```

PR #84, opened 2019-03-28
[core] change _handleDisconnect

yuanyu90221:
remove console.error in _handleSessionDisconnect
螢幕快照 2019-03-28 上午10 42 27

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0-beta-json.7",
+  "version": "1.0.0-beta.6",
   "packages": [
     "packages/*"
   ]
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0-beta.6",
+  "version": "1.0.0-beta-json.7",
   "packages": [
     "packages/*"
   ]
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -736,7 +736,6 @@ class Connector {
         event: 'disconnect',
         params: [{ message }]
       })
-      console.error(message) // tslint:disable-line
     }
     this._removeStorageSession()
     this._toggleSocketPing()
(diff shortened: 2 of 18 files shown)


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0-beta.6",
+  "version": "1.0.0-beta-json.7",
```
pedrouid: Don't worry about the versioning. I will handle it from my side. However I'm curious about why you called it `-beta-json.7`???

packages/core/src/index.ts
```diff
@@ -736,7 +736,6 @@ class Connector {
         event: 'disconnect',
         params: [{ message }]
       })
-      console.error(message) // tslint:disable-line
```
pedrouid: Awesome! Thanks

packages/core/src/index.ts
```diff
@@ -838,7 +837,7 @@ class Connector {
   private _subscribeToInternalEvents () {
     this.on('wc_sessionRequest', (error, payload) => {
       if (error) {
-        console.error(error) // tslint:disable-line
+        console.log('[wc_sessionRequest]',error) // tslint:disable-line
```
pedrouid: This isn't much of a change. We need to trigger an `error` event instead.

pedrouid:
Hi @yuanyu90221, thanks for your PR but there are too many unnecessary package-lock changes included on it. I've changed L739 on the `core` package like we discussed and also replaced other `console.error` calls with internal event `error` callbacks.

You can see these changes on the commit [907a98357d45bfe8a5bfb03a464983d17a13166b](https://github.com/WalletConnect/walletconnect-monorepo/commit/907a98357d45bfe8a5bfb03a464983d17a13166b0). I will be including these changes with other fixes coming this week on `beta.7`.

Thanks for reporting this issue and jumping on it  🙏 Appreciate it!

PR #100, opened 2019-04-10
Provider fixes (Fixes #96, Fixes #97)

### LINKED ISSUE #96: HandleRequest handles promise incorrectly on web3-provider
Change this
```js
const walletConnector = await engine.getWalletConnector()
const { error, result } = await walletConnector.sendCustomRequest(payload)
end(error, result)
```

To this

```js
const walletConnector = await engine.getWalletConnector()
try {
  const result = await walletConnector.sendCustomRequest(payload)
  end(null, result)
} catch (error) {
  end(error)
}
```

Check [L159](https://github.com/WalletConnect/walletconnect-monorepo/blob/73251a8c7239d5cf74bfb567f2a7d4394f2489b4/packages/web3-provider/src/index.js#L159)

### LINKED ISSUE #97: hooked-wallet-subprovider implemented incorrectly
Currently hooked-wallet-subprovider API is somewhat misleading as the method calls available as options do not necessarily relay the transactions correctly

Instead of using `sendTransaction`, it needs to handle `proccessTransaction` and consequently for other method calls

rmeissner:
- Closes #97 
  - There is no hook available for `processSignTransaction` (see https://github.com/MetaMask/web3-provider-engine/issues/308)
- Closes #96

### DIFF
--- a/packages/browser/package-lock.json
+++ b/packages/browser/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "@walletconnect/browser",
-	"version": "1.0.0-beta.9",
+	"version": "1.0.0-beta.10",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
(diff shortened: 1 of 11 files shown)


pedrouid:
Just added some changes but I will give it a try locally before merging.

Issue #105, opened 2019-04-16
Refactor exchangeKey flow


pedrouid:
Current implementation of exchangeKey is insecure. I first designed it as a pragmatic approach to recycle persisted sessions on browser's localstorage. However you can't securely exchange symmetric keys using previously compromised keys. This is something that I've been wanting to revisit and change for months but haven't prioritized in the whole scope of things that needed to be tackled.

A couple of weeks ago, me and @ligi brainstormed a better selection to ensure that the key exchange was secure and guaranteed to only be known by two parties. WalletConnect protocol is intended to only be used 1-on-1 therefore anyone else with access to the symmetry key exposed through URI (displayed by either a QR Code or deep link) shouldn't be able to decrypt these communications between these parties.

The first proposed change is that exchangeKey should only be triggered after the connection is established, therefore no unnecessary computation is used unless the session is approved by the counter-party (which currently is exchange prior approval).

The second proposed change is that the exchangeKey flow should be initiated by a signing challenge through a JSON-RPC request with the method `wc_signingChallenge`. For example, after connection approval the Dapp will request a signing challenge (which should be specific for WalletConnect to be auto-signed). This is very analogous to a similar proposal that I made on ETH Magicians [here](https://ethereum-magicians.org/t/automatic-authentication-signature/2429/8?u=pedrouid). The Wallet would receive this challenge and trigger an event for the signing challenge which can be handled with a callback as follows.

```javascript
walletConnector.on("signing_challenge", (error, payload) => {
  if (error) {
    throw error;
  }

  wallet.sign(payload.params[1])
    .then((signature) => {
      walletConnector.approveSigningChallenge(signature)
    })
    .catch((error) => {
      walletConnector.rejectSigningChallenge(error.message)
    });
});
```

The reason we included this signing challenge is to be used by the Dapp to recover the public key to be used to encrypted the contents of the following exchangKey request. Hence after the Wallet has successfully signed the challenged, the Dapp will follow with a new request that will contain an encrypted payload with the next symmetric key to be used in the communications. The JSON-RPC request should have a method of `wc_updateKey` and should include a single parameter that includes a hexadecimal encoded string of the next key encrypted with the public key recovered from the signing challenge. We won't reuse the `wc_exchangeKey` JSON RPC method in order to keep backwards compatibility with previous v1.0.0-beta versions. This key can be decrypted by subscribing to an event for the key update which can be handled with a callback as follows.

```javascript
[...]

PR #106, opened 2019-04-17
Decouple WebSocket transport from Core

pedrouid:
I've started a branch called `refactor-core` in order to start refactoring a decoupling components from the massive core library file. Looking for feedback and suggestions to improve the refactor.

The goal is to decouple WebSockets, EventEmitter, ExchangeKey, InternalEvent handlers, JSON-RPC handles, Subscribers, etc

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0-beta.9",
+  "version": "1.0.0-beta.10",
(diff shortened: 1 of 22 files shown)


pedrouid:
PS - I've added changes from the PR #104 to disable exchange key temporarily as it's also going to be refactored following the Proposal described on #105

PR #116, opened 2019-04-30
Fix signPersonalMessage Bug

GongDDEX:
In `signPersonalMessage` function, convert message to hex if message  is not a hex, but return a string without `'0x'` prefix. 

In https://test.walletconnect.org source code
```
export async function signMessage(message: any) {
  if (activeAccount) {
    const result = await activeAccount.signMessage(
      message.substring(0, 2) === "0x"
        ? ethers.utils.arrayify(message)
        : message
    );
    return result;
  } else {
    console.error("No Active Account"); // tslint:disable-line
  }
  return null;
}
```

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -571,7 +571,7 @@ class Connector {
     }
 
     if (!isHexStrict(params[1])) {
-      params[1] = convertUtf8ToHex(params[1])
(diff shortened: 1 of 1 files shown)


pedrouid:
Awesome! Thanks for reporting this. It may be better to fix the utils to support both prefixed and non-prefixed hex values.

PR #117, opened 2019-05-02
Add utils hex prefix argument

pedrouid:
Replace #116


pedrouid:
Hey @GongDDEX, going to take this approach instead

PR #160, opened 2019-07-06
Implement WCIP-6

pedrouid:


pedrouid:
Requires WalletConnect/node-walletconnect-bridge#32

PR #178, opened 2019-08-10
Channel Provider

pedrouid:


packages/channel-provider/README.md
```diff
@@ -0,0 +1,66 @@
...
+  nodeUrl: "<INSERT_NODE_URL>",
+  store
+});
+
+/**
+ *  Deposit
+ */
+channel.deposit({
+  amount: "0x3abc", // represented as bignumber
+  assetId: AddressZero // Use the AddressZero constant from ethers.js to represent ETH, or enter the token address
```
LayneHaber: You can enter this as just `amount` as a string, or leave the `assetId` as `undefined` if you are transferring eth.

the `string --> bignumber` and `assetId --> AddressZero` conversions will happen automatically

packages/channel-provider/README.md
```diff
@@ -0,0 +1,66 @@
...
+  amount: "0x3abc" // in Wei, represented as bignumber
+  assetId: AddressZero // represents ETH
+})
+
+
+/**
+ * Withdrawing
+ */
+await channel.withdraw({
+  recipient: // defaults to signer xpub but can be changed to withdraw to any recipient
```
LayneHaber: defaults to signer accounts[0], recipient should be an ETH address not an xpub

PR #182, opened 2019-08-26
updated the donation page to not use a very long url

julien51:
Hello!
This is a PR to improve the experience on the "Sponsor" button.
Right now it shows a very long URL because that URL includes a base64 encoded version of the configuration.
This PR adds the configuration from the repo which means that now the URL can be much cleaner.

Here is how it looks like:

image

Note: this is a purely cosmetic change. The older version still works so you can dismiss that PR if you think it is not that important.


pedrouid:
Thanks! 🙌

PR #200, opened 2019-11-15
Increase default z-index to fix overlapping

mirshko:
A lot of websites might have a header that is a higher `z-index` than 100, in the case of making sure that something is higher than the rest of the content.

It makes sense to have the z-index for the modal as high as possible.

The value its set to is the same value as Intercom uses to make sure the Intercom messenger is always on top.


pedrouid:
Thanks 🙏 I will have it out in the next release

Issue #193, opened 2019-11-12
Can't reconnect with stale localStorage.walletconnect data


Velenir:
Good day!

I have a problem with reestablishing connection to a mobile Wallet when **localStorage** in the browser has connection information saved but the session was terminated on the side of the Wallet.

I have a minimal example [here](https://codesandbox.io/s/walletconnect-setup-48ezu) but it also happens exactly the same at https://example.walletconnect.org/ . Wallets I tested with: Trust Wallet, Gnosis Safe and https://test.walletconnect.org/
Steps to reproduce:

1. Connect to a mobile wallet
2. Close the page
3. Disconnect on the mobile wallet (page closed -> **localStorage** isn't cleared)
4. Open the page again
5. Try to connect

Connection is established (or looks established) for a brief moment and then disconnect event happens as the session data in **localStorage** is obviously stale.

What would you recommend to handle this?
After briefly browsing the source code I saw that a socket is opened in WalletConnector constructor, which causes a disconnect event if connection data is stale, so my workaround has been to wait a second, but that is obviously not ideal.

I was thinking maybe a property like `connected` but one that would return a Promise which will resolve only when WalletConnector has tried and either failed or succeeded in establishing a connection.
So that I could do
```js
const provider  = new WalletConnect({
    infuraId: "***"
  });

// immediately resolved with false if nothing in localStorage
// resolved with true if connection could be reestablished with the data from localStorage
// resolved with false if connection failed
const connected = await provider.wc.connectedPromise

await provider.enable()
```

Current behavior makes it hard to setup _autoconnection_ to WalletConnect on page load as we can't be sure the session data in **localStorage** is valid. Have to introduce an artificial delay to be sure.

pedrouid:
@Velenir could you confirm this has been fixed with PR #206?

jklmli:
This is still an issue - this doesn't work:
```javascript
await activate(walletconnect, undefined, true);
```
will log `Warning: Suppressed stale connector activation` due to `StaleConnectorError`.

One workaround is this:
```javascript
await activate(walletconnect, undefined, true);
await (walletconnect.walletConnectProvider as WalletConnectProvider).onDisconnect();
await activate(walletconnect, undefined, true);
```

This is a little more robust:
```javascript
const provider = walletconnect.walletConnectProvider as WalletConnectProvider;
await Promise.any([
  new Promise((resolve) => provider.onConnect(resolve)),
  provider
    .onDisconnect()
[...]

PR #206, opened 2019-12-06
Fix wrong initial state of Web3Provider

Velenir:
When `web3-provider` is instantiated it assumes `chainId` to be either `1` (default) of whatever is provided as constructor parameter.
After WalletConnect is connected on the wallet side, `web3-provider` should sync its state to account and chain id of the wallet. But the [callback for wc.on('connect'](https://github.com/WalletConnect/walletconnect-monorepo/blob/master/packages/web3-provider/src/index.js#L301) currently has wrong signature, so the state is never updated.

Also if WalletConnector is already connected (`wc.connected === true`, which can be problematic in its own way #193), state is likewise not updated.

State is synced only on `session_update` event if that happens later.

As a result if we connect to, for example, a Rinkeby wallet, `web3-provider` still has `chainId = 1` and HttpConnection (`this.http`) makes requests to `mainnet.infura.io`

This PR solves that by correcting `wc.on'connected` callback and updating state with `wc.session` on initial `getWalletConnector` call.

### DIFF
--- a/packages/web3-provider/src/index.js
+++ b/packages/web3-provider/src/index.js
@@ -33,6 +33,7 @@ class WalletConnectProvider extends ProviderEngine {
 
     this.wc = new WalletConnect({ bridge: this.bridge })
     this.isConnecting = false
+    this.connected = false
     this.isWalletConnect = true
     this.connectCallbacks = []
     this.accounts = []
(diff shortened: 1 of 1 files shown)


pedrouid:
Thanks so much for this PR 🙌 I'm surprised this hasn't raised any issues until now. Probably just a coincidence that most implementations must be using `chainId=1`

PR #214, opened 2019-12-19
BUG: Can't read property "disableSessionCreation" of undefined

mikheevm:
I accidentally broke walletconnect web3-provider with this PR by not checking if options object is defined: https://github.com/WalletConnect/walletconnect-monorepo/pull/208

This PR fixes that
Screenshot 2019-12-19 at 11 51 00


pedrouid:
Thanks @mikheevm for this fix 🙌 I've published this fix v1.0.0-beta.42

PR #228, opened 2020-01-23
Add option to set `silent` flag true or false on requests

pedrouid:


packages/types/index.d.ts
```diff
@@ -258,4 +258,9 @@ declare module '@walletconnect/types' {
     rpc?: IRPCMap
     infuraId?: string
   }
+
+  export interface IRequestOptions {
+    _topic?: string
```
pedrouid: Remove custom topic option

packages/types/index.d.ts
```diff
@@ -258,4 +258,9 @@ declare module '@walletconnect/types' {
...
+    pushNotification?: boolean
```
pedrouid: Rename pushNotification to forcePushNotification

packages/types/index.d.ts
maricalucian: Moved topic back to separate param, and made options second optional param

packages/types/index.d.ts
maricalucian: Done

pedrouid:
Thanks @maricalucian, I've merged into a branch to tweak it before releasing. Check #230 for updates

PR #224, opened 2020-01-18
Add mobile-linking support

pedrouid:
- [x] Detect browser and OS environment
- [x] Display deep linking for mobile
- [ ] Display single deep link for Android
- [x] Display multiple deep link for iOS
- [ ] Register deep link on core client
- [ ] Trigger deep link on signing requests

PR #232, opened 2020-02-07
support mobile deeplinking

pedrouid:


pedrouid:
I've added these changes to the other PR @brunobar79, let's start working from the other one.

replaced by #224

PR #238, opened 2020-02-23
[web3-provider] Possibility to catch when session is created

Alirun:
This Pull Request allows to register callbacks to event when session is created in `web3-provider`

The idea came up when I wanted to improve UX and automatically trigger deeplink when QR code appears for scanning, but there was no way to catch that while using `provider.enable()`

Now it's simply possible to register callback on session creation, get `provider.wc.uri` and use it however you want


pedrouid:
There is a better way to do this actually

```typescript
const provider = new WalletConnectProvider({
    infuraId: "INSERT_INFURA_ID", // required
    qcorde: false, // disable qrcode modal
})

// display_uri will be triggered on session creation
provider.wc.on('display_uri', uri => console.log('uri', uri))

const web3 = new Web3(provider)
```

PR #267, opened 2020-05-10
Health check

miohtama:
Ready to merge. 

This is a health check script that does a basic ping pong message over a bridger server.

* Can be run from a command line
* Can be run as a Discord bot and alert on any problems

Later I plan to add:

* Can be run on a standalone web page for wallet support people to diagnose issues


.eslintignore
```diff
@@ -1 +1,2 @@
-example
\ No newline at end of file
+example
+**/node_modules/*
```
miohtama, line 3: Not sure why this was needed for health-check package, as other packages managed to run with the default eslint configuration without desdencing linting `node_modules`.

packages/operations/health-check-cli/src/healthcheck.ts
```diff
@@ -0,0 +1,271 @@
+import Connector from "@walletconnect/core";
+import * as cryptoLib from "@walletconnect/iso-crypto";
```
pedrouid, line 2: Did you have any issues using the `@walletconnect/client`? It would be great to replace it for these two libraries

.eslintignore
pedrouid: no worries, this is reasonable

packages/operations/health-check-cli/src/healthcheck.ts
miohtama: I did not try, but I can replace it with the client. Looks like it does the same job of few lines.

PR #277, opened 2020-06-02
make android work identically to iOS

brunobar79:
This PR removes any difference between Android & iOS in terms of mobile deeplinking.
Both platform work and behave the same way now (using AppLinks for Android and Universal Links for iOS)


pedrouid:
Is this really necessary? Android can just use a single button to prompt supporting Wallets available in the user's device, why are we changing this?

pedrouid:
Is this really necessary? Android can just use a single button to prompt supporting Wallets available in the user's device, why are we changing this?

brunobar79:
You

> Is this really necessary? Android can just use a single button to prompt supporting Wallets available in the user's device, why are we changing this?

You're right! forgot about that

PR #279, opened 2020-06-03
Fix event listener in web3-provider

Velenir:
There is a typo in event name. 
As a result provider never disconnects fully.

Also currently it calls `this.stop` on `L388`, which results in `'stop'` event from `web3-provider-engine`. But` 'close'` event is never emitted. Shouldn't it call `this.close()` instead of `this.stop()` or at least emit `'close'`?

### DIFF
--- a/packages/providers/web3-provider/src/index.ts
+++ b/packages/providers/web3-provider/src/index.ts
@@ -330,7 +330,7 @@ class WalletConnectProvider extends ProviderEngine {
 
   async subscribeWalletConnector() {
     const wc = await this.getWalletConnector();
-    wc.on("disconnetect", error => {
+    wc.on("disconnect", error => {
       if (error) {
         this.emit("error", error);
         return;


pedrouid:
Well the event `close` is custom to WalletConnect... it was still being debated on EIP-1193 what events it should include...

https://github.com/ethereum/EIPs/issues/2319

I need to include the new events `connect` and `disconnect`

Velenir:
I see you included `'disconnect'` event in 28788bc, which is great

But like it now is the event only emitted when `provider.disconnect() or provider.close()` is called client-side.
Shouldn't it also be emitted when a 'disconnect' event is received from the connected wallet, in `wc.on('disconnect'`?. By calling `this.close()` on [L347](https://github.com/WalletConnect/walletconnect-monorepo/blob/next/packages/providers/web3-provider/src/index.ts#L347)

pedrouid:
You're right, I need to add it there too 🙏 good catch

PR #282, opened 2020-06-04
move onclose up, make more clickable

21paradox:
walletconnect-modal__close__icon size too small.


pedrouid:
Thanks @21paradox  🙌  published under beta.102

PR #284, opened 2020-06-05
Fix read requests never resolving

Velenir:
`HttpConnection` never emits events, so [event listeners](https://github.com/WalletConnect/walletconnect-monorepo/blob/next/packages/providers/web3-provider/src/index.ts#L409L410) never fire and promises from`handleReadRequests` never resolve.

Maybe also remove  [event listeners](https://github.com/WalletConnect/walletconnect-monorepo/blob/next/packages/providers/web3-provider/src/index.ts#L409L410) if you see fit

Btw, `return this.http.send(payload);` was there previously and was removed at some point between `beta.47 -- beta.61`

### DIFF
--- a/packages/providers/web3-provider/src/index.ts
+++ b/packages/providers/web3-provider/src/index.ts
@@ -283,14 +283,7 @@ class WalletConnectProvider extends ProviderEngine {
       this.emit("error", error);
       throw error;
     }
-    this.http.send(payload);
-    return new Promise(resolve => {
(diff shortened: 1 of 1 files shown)


pedrouid:
This one failed because of the this.http.send returns type `unknown` you can just type cast it

Velenir:
> This one failed because of the this.http.send returns type `unknown` you can just type cast it

Done

PR #285, opened 2020-06-05
Allow provider.request to handle write requests

Velenir:
Currently in web3-provider `provider.request` doesn't handle write requests (`eth_sendTransaction` etc.)
Even `provider.send` handles write requests only when given a callback as the second argument, otherwise requests are passed to `handleOtherRequests -> handleReadRequests`

I propose adding a `handleWriteRequests`method for some request.methods.

I'm not sure if `personal_*` request.methods should be handled like write requests (with `this.sendAsync`) or are they supposed to go through [wc.sendCustomRequest](https://github.com/WalletConnect/walletconnect-monorepo/blob/6688db304b/packages/providers/web3-provider/src/index.ts#L276)?

Also maybe more methods should be added, like `eth_signTypedData`?

### DIFF
--- a/packages/providers/web3-provider/src/index.ts
+++ b/packages/providers/web3-provider/src/index.ts
@@ -253,6 +253,7 @@ class WalletConnectProvider extends ProviderEngine {
         case "personal_sign":
         case "personal_sendTransaction":
           response = await this.handleWriteRequests(payload);
+          break;
         default:
           response = await this.handleOtherRequests(payload);
       }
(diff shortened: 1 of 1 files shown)


pedrouid:
These are great 🙌 Thanks a lot of tackling them. We should add some unit testing for these too.

`personal_*` are also write requests because they require signing

Travis detected a missing `break` statement before the default case.. could you include this one?

Velenir:
> Travis detected a missing `break` statement before the default case.. could you include this one?

Oops, was careless of me

PR #286, opened 2020-06-05
web3-provider improvements

pedrouid:
I've moved your changes to this branch @Velenir as they were breaking some functionality...

These changes will need us to add some unit testing before they can merged into the default branch


Velenir:
What was breaking? Something in request handling?

pedrouid:
Yes, the changes to how provider.request was handling signing requests was breaking. Need to investigate further. However I've included fixes for HttpConnection missing payloads 🙌

PR #298, opened 2020-06-22
Fix add id and jsonrpc fields to EIP-1193 request payloads

kimpers:
This fixes the issue of EIP-1193 requests being invalid due to missing required fields.


pedrouid:
We should add it inside of send method instead of the request method also use payloadId() util for generating unique Ids

pedrouid:
Actually I will just merge it and publish it on the next tag

PR #305, opened 2020-07-02
Release v1.1.0

pedrouid:
- [x] add i18n support to qrcode-modal
- [x] add imToken Wallet links
- [x] add Pillar Wallet links
- [x] add Math Wallet links
- [x] update mobile links display
- [x] add ability to filter mobile links


packages/helpers/qrcode-modal/src/browser/assets/style.css
```diff
@@ -318,10 +318,19 @@
   font-size: 15px;
 }
 
+.walletconnect-connect__buttons__wrapper {
+  max-height: 44vh;
```
mirshko, line 346: Generally the content of the buttons should determine the max height, if the button change height based n the screen it can cause responsive issues or overlapping.

PR #345, opened 2020-09-03
Import EventEmitter as named import instead of default

geigerzaehler:
We import `EventEmitter` from `events` using a named import instead of using the default export. Before, this would break when using `@types/node@14`. Moreover this would break when using [rollup][1] together with [`rollup-plugin-node-polyfills`][2].

[1]: https://rollupjs.org/
[2]: https://github.com/ionic-team/rollup-plugin-node-polyfills

### DIFF
--- a/packages/helpers/http-connection/package.json
+++ b/packages/helpers/http-connection/package.json
@@ -48,7 +48,7 @@
     "@babel/register": "7.8.3",
     "@types/jest": "22.2.3",
     "@types/lodash.isnumber": "3.0.6",
-    "@types/node": "12.12.14",
+    "@types/node": "14.6.3",
     "npm-run-all": "4.1.5",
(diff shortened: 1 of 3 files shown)


pedrouid:
I'm still annoyed at this breaking change introduced by `@types/node` 😅 

Quick question: will this break compatibility for people who have previously integrated WalletConnect? what would be the best way to test this?

geigerzaehler:
> will this break compatibility for people who have previously integrated WalletConnect?

This won’t break compatibility if they are using the NodeJS version of the `events` module or a package that is compatible with it (e.g. [`events`](https://www.npmjs.com/package/events) on NPM which I believe is used by webpack).

> what would be the best way to test this?

I’m not familiar with the tests of this project. But if they still pass then this should be enough because the events module itself did not change.

PR #353, opened 2020-09-18
QRCode click to copy URI

miguelmota:
Not sure if there's already an existing way of doing this but when using WalletConnect where the dapp and wallet are on the same device and browser (e.g. interacting with dapp on mobile browser while using a browser based-wallet), there needs to be an easy way to view or copy the connection URI to the clipboard since the QRCode image can't be scanned.

This simple PR copies the connection URI to the clipboard when the QRCode image is clicked.


pedrouid:
Great idea! I actually think that could be further improved by:
* updating the copy to notify users of the copy to clipboard option
* display a text version with a copy to clipboard button next to it

Going to take a go at this before releasing. Thanks for the initiative @miguelmota! 🙏

miguelmota:
Awesome! Thanks Pedro

PR #362, opened 2020-09-27
fix(socket): add error handler for incoming connection error

brayden-ooi:
The error handler is added to catch errors such as 

`WebSocket connection to 'wss://bridge.walletconnect.org/' failed: Invalid frame header`

and `502` on the front end. However, more work will need to be done to pass the error down to the library users.


brayden-ooi:
I have tested the draft and it appears that the throw statement does not work in event listeners as with ordinary functions.

packages/clients/core/src/index.ts
```diff
@@ -1071,6 +1071,16 @@ class Connector implements IConnector {
...
+    this._transport.on("error", (e: Event)  => 
+      this._eventManager.trigger({
+        id: 0,
+        jsonrpc: '',
+        error: {
+          message: "Websocket connection failed"
+        },
+      })
+    );
+
```
brayden-ooi, line 1083: This hack will pass a `JsonRpcResponseError` object to eventManager and the error will be caught under the `call_request` listener.

pedrouid:
Great catch. Still need to look into how this error should be thrown and handled by integrators

brayden-ooi:
yup definitely, ordinary throw statements do not work with event listeners so it might be better to have an error callback within the classes or just let the users to define the callback and pass it to instantiate the WalletConnect class. We can also have an error event along with 'message', 'open' and 'close', a more complete approach than what I have done in this PR. This way we can refactor all the errors and let the 'error' event be responsible to all the errors that could happen during the communication phase between devices. Then we can perform some cleanups after the error event is triggered.

Anyhow, let me know if there's anything i can help with once you have made up your mind.

brayden-ooi:
I have attached a 502 for your reference
Screenshot 2020-10-02 at 11 09 44 AM

andylim0221:
is there any feedback for this PR?

pedrouid:
I'm going to make the change simply to include this error event to be labelled as transport_error instead of formatting it as a JSON-RPC response

PR #368, opened 2020-10-05
Add signature and improve implementation for provider method

geigerzaehler:
We add an explicit signature to `Web3Provider#enable` and make the implementation more idiomatic.


pedrouid:
Thanks 👍 this makes sense

PR #370, opened 2020-10-09
Allow reconnect by resetting 'pending'

geigerzaehler:
Before, calling `connect` after `killSession` did not open the modal and reconnect. The issue was that the value of `pending` was never reset to `false` after the connection was established.

### DIFF
--- a/packages/clients/core/src/index.ts
+++ b/packages/clients/core/src/index.ts
@@ -89,6 +89,7 @@ class Connector implements IConnector {
 
   private _handshakeId = 0;
   private _handshakeTopic = "";
+  private _pending = false;
 
   // -- session ----------------------------------------------------- //
 
@@ -319,11 +320,11 @@ class Connector implements IConnector {
   }
 
   set pending(value) {
-    // empty
+    this._pending = value;
   }
 
   get pending() {
-    return !!this._handshakeTopic;
+    return this._pending;
   }
 
   get session() {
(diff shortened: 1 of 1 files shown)


packages/clients/core/src/index.ts
```diff
@@ -319,11 +320,11 @@ class Connector implements IConnector {
   }
 
   set pending(value) {
-    // empty
+    this._pending = value;
```
pedrouid: this should still be empty

packages/clients/core/src/index.ts
```diff
@@ -413,25 +414,22 @@ class Connector implements IConnector {
...
-      }
 
+    if (this.connected) {
+      return {
+        chainId: this.chainId,
+        accounts: this.accounts,
+      };
+    }
+
+    this.pending = true;
```
pedrouid: this should be set on the private variable `this._pending`

packages/clients/core/src/index.ts
```diff
@@ -441,6 +439,8 @@ class Connector implements IConnector {
 
         resolve(payload.params[0]);
       });
+    }).finally(() => {
+      this.pending = false;
```
pedrouid: same as above

pedrouid (changes requested):

NunoAlexandre:
@pedrouid @geigerzaehler I have just tried running the latest released version, `1.3.5` and this issue seems to persist.

pedrouid:
tbh I think reconnecting is advised against

it's much easier to just create a new connector

WalletConnect v1.0 clients weren't really designed for connecting and reconnecting since the connectors were designed for a single session usage

NunoAlexandre:
Thanks for your reply, @pedrouid. That is the approach I went with @ https://github.com/radicle-dev/radicle-upstream/pull/1591. 

The specific issue we are experiencing is that once the QR Code modal is triggered if it is dismissed, follow-up attempts to trigger it again won't do anything whatsoever. 

That seems like a bug rather than a design decision tho. The workaround used in the linked PR involves detecting a dismissal of the modal and create a new instance of `WalletConnect` in that case, which also needs to happen when the user disconnects their wallet.

PR #417, opened 2021-01-21
Refactor Connector to support multi blockchain platforms

baerae-colin:
Replacing the hard-coded string "eth" with a variable `rpcNamespace`,
Connector can handle JSON-RPC methods having different namespaces.

For example, Klaytn has a namespace of "klay" instead of "eth",
"klay_sendTransaction" will be used instead of "eth_sendTransaction".

To preserve compatibility, the default value of the variable is "eth".

### DIFF
--- a/packages/clients/core/src/index.ts
+++ b/packages/clients/core/src/index.ts
@@ -110,6 +110,9 @@ class Connector implements IConnector {
   private _qrcodeModal: IQRCodeModal | undefined;
   private _qrcodeModalOptions: IQRCodeModalOptions | undefined;
 
+  // -- namespace prefix ----------------------------------------------------- //
+  private _namespacePrefix: string;
+
   // -- constructor ----------------------------------------------------- //
 
   constructor(opts: IConnectorOpts) {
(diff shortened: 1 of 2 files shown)


pedrouid:
Hey @kjhman21, this is unnecessary since you can just send your own custom request:

https://github.com/WalletConnect/walletconnect-monorepo/tree/next/packages/clients/client#send-custom-request

Additionally WalletConnect v2.0 will explicitly handle different namespaces

baerae-colin:
Hi, @pedrouid, with this change, we can use the same function calls (sendTransaction). This is what I wanted.

If it is not suitable for wallet connect, please let me know. 

Also, could you tell me when WalletConnnect v2.0 will be released?

pedrouid:
well technically this change wouldn't hurt per se

my point was that these changes are all done at the connector level when you can already send any JSON-RPC request you want without needing this change

this PR would be more useful if this be configured all the way up to the provider level like web3-provider

baerae-colin:
Okay. I got it. I will make another PR for this. Thanks.

One more thing to ask! Do you have any planned date for WalletConnect 2.0? or Is there any document or blogs for the WalletConnect 2.0 plan?

pedrouid:
WalletConnect v2.0 available on NPM: https://www.npmjs.com/package/@walletconnect/client/v/experimental

WalletConnect v2.0 documentation: https://docs.walletconnect.org/v/v2.0.0-alpha/

baerae-colin:
@pedrouid Thanks for the info. I've just seen the documentation, but I couldn't find the JSON-RPC API methods for Cosmos and Polkadot. When the documentation will be filled?

PR #438, opened 2021-02-10
fix(web3-provider): stop createSession after close

### LINKED ISSUE #436: Do not reopen modal when using disconnect()
Hey, when I call disconnect(), the modal reopens; any way to disable this?
I want to fully disconnect and use the app as a guest/unknown user

jrhender:
Fixes #436 
`enabled` flag stops `getWalletConnector` from creating session after `close`.
Otherwise, `getWalletConnector` call (from `handleRequest`) ends up calling `wc.createSession`.

I think `handleRequest` might be being continually triggered because of the block polling of `Web3ProviderEngine`:
https://github.com/MetaMask/web3-provider-engine/blob/c8d9a8e46703ab417aeeeba583694057f38cfdf7/index.js#L24

I think the block polling should be stopped by the call to `stop` in `onDisconnect`, but I think this call may not be effective...

An alternative solution could be to have `close` remove the providers added during `initialize` (as then `handleRequest` wouldn't get called). The providers could then be re-added if `enable` is called again.

### DIFF
--- a/packages/providers/web3-provider/src/index.ts
+++ b/packages/providers/web3-provider/src/index.ts
@@ -34,6 +34,7 @@ class WalletConnectProvider extends ProviderEngine {
   public chainId = 1;
   public networkId = 1;
   public rpcUrl = "";
+  private enabled = false;
 
   constructor(opts: IWalletConnectProviderOptions) {
     super({ pollingInterval: opts.pollingInterval || 8000 });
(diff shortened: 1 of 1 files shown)


jrhender:
I'm still relatively new to walletconnect/Web3ProviderEngine, so if I'm missing something here or there is an easier way, please let me know 😄

jrhender:
Not needed, implemented workaround instead, see issue #436

PR #439, opened 2021-02-11
Use uri deeplink for android

youngkidwarrior:
react-native-dapp helper package was using the iphone implementation for android, which supports agnostic deeplinks. Therefore I added a connect button for android OS using the connector uri

Testing Device: Galaxy S10
React Native version: 0.62.2

Steps to reproduce:
1. Install latest react-native-dapp: `yarn add @walletconnect/react-native-dapp@latest`

2. Wrap app in WalletConnectProvider from `@walletconnect/react-native-dapp
```javascript
import WalletConnectProvider from '@walletconnect/react-native-dapp';
import AsyncStorage from '@react-native-community/async-storage';

           <WalletConnectProvider
                bridge="https://bridge.walletconnect.org"
                clientMeta={{
                    description: 'Shenanigan Developer App',
                    url: 'https://she.energy',
                    icons: ['https://walletconnect.org/walletconnect-logo.png'],
                    name: 'Shenanigan'
                }}
                storageOptions={{
                    asyncStorage: AsyncStorage
                }}
            >
                            <App />
            </WalletConnectProvider>
```
3.Call connector.connect()
```javascript
import { useWalletConnect } from '@walletconnect/react-native-dapp';
const toggleWeb3 = async () => {
        if (!connector.connected) {
            await connector.connect();
        } else {
            await connector.killSession();
        }
    };
```

Choosing any wallet option opens `Warning` `Error: Unable to open url`


cawfree:
Hey @youngkidwarrior, unfortunately I was unable to replicate this issue. The current branch appears to open the selected wallet application on Android as expected (tested on qemu).

Would it be possible for you to provide replication instructions?

youngkidwarrior:
@cawfree updated my original post.

I'm curious about testing on an emulator. Usually they are not allowed to have any wallet apps to download from an app store

cawfree:
@youngkidwarrior

Thanks a lot, I'll give this another try now. I had installed [Rainbow](https://rainbow.me) from source.

cawfree:
Hey @youngkidwarrior, I was able to successfully replicate. Thanks a ton for raising this. 🙏 

As it currently stands, I'm not sure the proposed fix will be compatible with expected behaviour; this is because we expect the library to have some knowledge of which wallet was selected so that we can navigate into it later on; but I'm not yet sure we can get this information from the native `BottomSheet` that is presented by Android OS. So we just need to find some way to inform the app of what the selection was, so I'm going to experiment with `Linking.addListener` to see if any useful information can be ascertained.

I can still _navigate_ into an expected wallet as expected, but the connection prompt does not occur unless we navigate using the `universalLink`.

The error `Unable to open url.` error should only occur when a selected wallet is not installed. I've seen this happen once, which implies there could be a rendering synchronization issue with the library i.e. the WalletConnect `uri` is malformed or uninitialized.

youngkidwarrior:
@cawfree nice! Glad we caught it.

I see yes that explains some problems I was having yesterday actually with WalletService.

cawfree:
Enhancements and corrections have been applied in #442.

PR #452, opened 2021-03-05
Update Aktionariat wallet universal link

muratogat:
Updating UL because iOS doesn't trigger it when linked from a page under the same domain.
Thank you.


pedrouid:
Fixed on the new registry with commit: https://github.com/WalletConnect/walletconnect-registry/commit/bbd9c64952d54bd5770a8615f4a8f2f0f0ab3651

PR #459, opened 2021-03-11
add hashkey me to regitry

MrWude:


pedrouid:
This app doesn't look like it's currently available on the app stores only through test flight


Please make this submission on the new registry here: https://github.com/WalletConnect/walletconnect-registry

Issue #334, opened 2020-08-18
estimateGas returns undefined result and no error


yann300:
calling `eth_estimateGas` on a failing transaction returns:
```
jsonrpc: "2.0"
result: undefined
```
seems like the `error` property is missing.

I am using the provider 
`import WalletConnectProvider from "@walletconnect/web3-provider";`
from `this.provider = await web3Modal.connect()`

The call to `estimateGas` succeed if the transaction doesn't fail.

```
jsonrpc: "2.0"
method: "eth_estimateGas"
params: Array(1)
0:
 data: "0xd98471b2"
 from: "0xdfb0c20131328c90cc58650a3231695042a28875"
 to: "0x47e4783eaf21343166651f2fd98d1d9dfb0929de"
 value: "0x0"
```

```
curl https://ropsten.infura.io/v3/83d4d660ce3546299cbe048ed95b6fad \
    -X POST \
    -H "Content-Type: application/json" \
    -d '{"jsonrpc":"2.0","method":"eth_estimateGas","params": [{"from": "0xdfb0c20131328c90cc58650a3231695042a28875","to": "0x47e4783eAF21343166651f2Fd98D1D9dfb0929De","value": "0x0","data": "0xd98471b2"}],"id":234}'	
```

finessevanes:
Hey @yann300, was https://github.com/WalletConnect/walletconnect-monorepo/pull/464 able to fix your issue? If not, are you still experiencing it?

PR #464, opened 2021-03-23
Fix JSON-RPC error handling in handleRequest

gabmontes:
When using [web3-react](https://github.com/NoahZinsmeister/web3-react) and the WalletConnect provider, if a call reverts and the node returns a JSON-RPC error (having props `id`, `error` and `jsonrpc` but no `result`), that error is not properly parsed in the `handleRequest` function.

The current code behaves correctly if the call succeeds with a result or if completely fails but if the call succeeds with an error message, it tries to get the `result` property and ignores the error, returning an `undefined` to the caller.

This function was previously found buggy in #96, patched in #100 and is the most probable cause of #334.


pedrouid:
Thanks for submitting this PR @gabmontes 💪 and sorry for taking so long to review it!

PR #488, opened 2021-04-26
feat: auto-navigate iframe to qrcode URI

michaelfig:
This little hack auto-navigates a hidden iframe to the `wc:...` URI when it is displayed in a browser.

That allows wallets to register as handlers for the `wc` URI scheme, so that they can be automatically popped up on several platforms (my use case was a desktop native app).

I'd really like to see something like this merged, but I'm not set on the details for how to do it.  Please let me know what it would take to do so.

### DIFF
--- a/packages/helpers/qrcode-modal/src/browser/components/Modal.tsx
+++ b/packages/helpers/qrcode-modal/src/browser/components/Modal.tsx
@@ -25,6 +25,7 @@ function Modal(props: ModalProps) {
   const displayProps = {
     text: props.text,
     uri: props.uri,
+    navigate: props.qrcodeModalOptions ? props.qrcodeModalOptions.navigate : true,
     qrcodeModalOptions: props.qrcodeModalOptions,
   };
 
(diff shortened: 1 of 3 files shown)


packages/helpers/qrcode-modal/src/browser/components/Modal.tsx
```diff
@@ -25,6 +25,7 @@ function Modal(props: ModalProps) {
   const displayProps = {
     text: props.text,
     uri: props.uri,
+    navigate: props.qrcodeModalOptions ? props.qrcodeModalOptions.navigate : true,
```
michaelfig, line 28: The default here if qrcodeModalOptions is unspecified is `true`.  Arguably, for maximum compatibility, that should be `false`, but it seems rather harmless to me to always do the iframe trick.

pedrouid:
Hey @michaelfig, sorry it took me so long to review this. Is it really required to have an iframe for this? Wouldn't a new tab achieve the same goal?

michaelfig:
> Is it really required to have an iframe for this? Wouldn't a new tab achieve the same goal?

Yes, you're probably right.  Don't worry about this PR for now.  I am still working out how Wallet Connect would be of benefit to our platform, and until that's solidified, my changes are probably premature.

I'll close this PR until we have a mutually-agreeable approach if Wallet Connect doesn't already work for my use-case.

Thanks.

PR #510, opened 2021-05-14
fix to re-subscript after websocket error

skqksh:
After `WebSocket network error`, client couldn't get result of sendCustomRequest function.

스크린샷 2021-05-14 오전 11 31 57


pedrouid:
I will be publishing this as a release candidate first

PR #526, opened 2021-06-09
v1.5.0

pedrouid:
- fix clip-to-clipboard issues
- distribute ESM modules
- add storageId config


packages/helpers/signer-connection/src/index.ts
```diff
@@ -120,6 +120,14 @@ export class SignerConnection extends IJsonRpcConnection {
...
       throw new Error("Failed to register WalletConnect connector");
     }
+    if (this.wc.accounts.length) {
+      this.accounts = this.wc.accounts;
+    }
+    if (this.wc.chainId) {
+      this.chainId = this.wc.chainId;
+    }
+    // this.accounts = this.wc.accounts;
+    // this.chainId = this.wc.chainId;
```
mightymatth, line 130: should this be removed?

PR #544, opened 2021-07-06
fix #543

### LINKED ISSUE #543: iOS sign error
When I connect to D'CENT Wallet on iOS and then sign, it is connected to the onto wallet.

SeungjunWe:
If you use useCallback , the handleClickIos function will not change even if the pageLinks change and a re-render occurs.

This causes the handleClickIOS function of the ninth button on page 1 to be executed even if I click the D'CENT button on the ninth on page 2.

I fixed this issue by passing pageLinks as deps to useCallback .


pedrouid:
Interesting! I wasn't aware of this! 🙌  Thanks for the PR!

Could you take a quick look at the merge conflicts? Then I will publish it soon after 💪

pedrouid:
Thank you for debugging this!

Issue #566, opened 2021-08-03
"color" is undefined when selecting Desktop option in QR Modal


sammdec:
Affected version:  1.5.2
Steps to repro:
* Specify mobileLinks array in WalletConnect constructor use value of just 'rainbow'
* Boot app open QR modal & click the Desktop toggle the modal will throw the exception

ThaiBinh-13:
Is there any solutions for this issue?

8bu:
Is there any config to disable Desktop option?
Don't know what is it for..

nuCodeIO:
This issue is effecting our project as well, when will this fix be merged and avaiable?

finessevanes:
@sammdec @nucodeio is this still an issue?

PR #574, opened 2021-08-09
Fix: #566

8bu:
Fix of #566

This bug is kinda annoy since I can't really config whitelist registry without causing error.
It was caused by this function & return an array of undefined. I'm not sure why the lib was coded this way but to prevent any regression I just add a filter here. Hope this help.

image

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.6.0-alpha.0",
+  "version": "1.6.0",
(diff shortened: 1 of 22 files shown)


ThaiBinh-13 (approved):

pedrouid:
I will be publishing this is as release candidate first

pedrouid:
Actually this one will go directly to production

PR #575, opened 2021-08-09
Inherit prototype when wrapping wallet connector in react-native-dapp

### LINKED ISSUE #571: TypeError: connector.signPersonalMessage is not a function
Hi, I'm trying to sign a message, and after executing connect, the following is executed:
```js
const message = "test message";

const hexMsg = convertUtf8ToHex(message);

const msgParams = [hexMsg, accounts[0]];

connector.signPersonalMessage(msgParams).then(
  function (result) {
    console.log("sign result:", result)
  }
)
```

However, the following error occurs:
```
TypeError: connector.signPersonalMessage is not a function. (In 'connector.signPersonalMessage(msgParams)', 'connector.signPersonalMessage' is undefined)
```

This is also the case with some other functions in connector, but connect function appears to be functional (I get accounts, chainId, etc. back).

Any help is much appreciated. Thanks.

carlosdp:
Fixes #571 

Currently, when the `WalletConnect` connector is wrapped to override the `connect` function for react-native, we use an object splat to carry over the properties from the original `connector`.

However, this **does not** carry over the *prototype* for `connector`, and therefore the new `nextConnector` is missing all the functions from the original `connector`, except the newly wrapped `connect` function. I assume this worked at some point because these functions were probably in the properties (instead of the prototype) in `WalletConnect` when this library was originally written, and at some point recently that must have changed.

This should be merged ASAP, as the library is fully broken at the moment!

### DIFF
--- a/packages/helpers/react-native-dapp/src/providers/WalletConnectProvider.tsx
+++ b/packages/helpers/react-native-dapp/src/providers/WalletConnectProvider.tsx
@@ -269,7 +269,7 @@ export default function WalletConnectProvider({
         ...intermediateValue,
         walletServices,
         connectToWalletService,
-        connector: {
+        connector: Object.assign(Object.create(connector), {
           ...connector,
           connect: async (opts?: ICreateSessionOptions) => {
             if (!walletServices.length) {
@@ -283,7 +283,7 @@ export default function WalletConnectProvider({
             setConnector(nextConnector);
             return nextConnector.connect(opts);
           },
-        } as WalletConnect,
+        } as WalletConnect),
       }
     }
     return {


carlosdp:
I published this branch to `@carlosdp/react-native-dapp` in the mean time, for anyone who needs it.

pedrouid:
Could you review this? cc @cawfree

cawfree:
Yeah this looks good to me. These functions are part of the original example flow on [create-react-native-dapp](https://github.com/cawfree/create-react-native-dapp), so it makes sense that if the invocation signature had changed significantly, it would prevent them from being pulled over.

Thanks for spotting this @carlosdp!

pedrouid:
Thanks @cawfree for the quick review 💪 

Thanks @carlosdp for the PR

I will get this included on the v1.6.1 release

Issue #588, opened 2021-08-24
React Native: Linking fails on Android


tuomohopia:
# Linking Fail
> On React Native

Wallet Connect fails to link on React Native, observed on Android.

## Problem
What happens when I'm attempting `connector.connect()` in the below code is that Android wants to open a website link to https://walletconnect.org/wallets instead of a deep link to an wallet app on the same device.

I have the following supported wallet apps installed on my Android phone:
* MetaMask
* Trust Wallet

The code looks something like this
```javascript
import { useWalletConnect, withWalletConnect } from '@walletconnect/react-native-dapp'

const Registration = () => {
	const connector = useWalletConnect()

	return (
		<Button
			onPress={async () => {
				try {
					connector.connect({ chainId: 56 })
				} catch (err) {
					console.error("Error adding a Wallet with WalletConnect", err)
				}
			}}
		/>
	)
}

export default withWalletConnect(Registration, {
	redirectUrl: "myapp://home",
	storageOptions: {
		// @ts-expect-error: inconsistent dependency types
		asyncStorage: AsyncStorage,
	},
})
```

## Cause

Now, I checked the lib code generated to `node_modules` for `@walletconnect/react-native-dapp`, and it goes like this:
```javascript
        if (Platform.OS === 'android') {
            const canOpenURL = await Linking.canOpenURL(uri);
            if (!canOpenURL) {
                Linking.openURL('https://walletconnect.org/wallets');
                throw new Error('No wallets found.');
            }
            await Linking.openURL(uri);
        }
```

So evidently, `Linking.canOpenURL(uri)` fails. This results in the code defaulting to opening a website URL as seen above instead of trying to connect to a wallt app.

Logging the `uri`, it reads like `wc:0891245e-f6e6-4482-9b02-a1e536d42406@1?bridge=https%3A%2F%2Fj.bridge.walletconnect.org&key=d95c899bd3d92499da8c39f797111efa95cebb46081f0f79b88dd752948d4750`

Is this a valid URI that the app should be checking if there are other apps using this scheme or not?

tuomohopia:
Experimenting with the code, I just found that by **erasing this if-clause** from the lib code, I'm able to correctly open the link as a walletconnect link again with Trust Wallet and MetaMask installed on my device:

```javascript
            if (!canOpenURL) {
                Linking.openURL('https://walletconnect.org/wallets');
                throw new Error('No wallets found.');
            }
```

[...]

Issue #563, opened 2021-07-31
Connecting to WalletConnect on mobile redirects to app store listing


vernonjohnson:
When connecting to WalletConnect from mobile (iOS) I'm redirected to the app store listing for the mobile wallets (Metamask, Rainbow).

<img src="https://user-images.githubusercontent.com/25010543/127751094-88b7a3de-140b-4466-a72a-90f3ec953723.PNG" width="200" height="400">
<img src="https://user-images.githubusercontent.com/25010543/127751098-cc770ff3-5434-41ac-befd-d7e394a5f078.PNG" width="200" height="400">

hassantsyed:
I am experiencing similar issues on android. However, I get redirected to the website wallet registry.
I cannot connect my wallet. This is with version 1.5.2.

kkaczynski:
The same on my side - Android is opening website wallet registry.

tuomohopia:
I'm experiencing this too, have been with all the versions above `1.5.0`.

Having missed this, I submitted my own issue about it: https://github.com/WalletConnect/walletconnect-monorepo/issues/588

tuomohopia:
@hassantsyed @kkaczynski @verminator23  I just found out how I'm able to solve this locally.

Please see https://github.com/WalletConnect/walletconnect-monorepo/issues/588#issuecomment-905034477 of how it resolves at least on my setup locally.

I haven't been able to identify yet what actually produces this code and how to solve this properly. Could use your help figuring this out if you're still experiencing this.

ai-slave:
Same @ iOS, clicking 1inch redirects to another app strore listing (some other waller)

finessevanes:
@ai-slave @vernonjohnson @tuomohopia @hassantsyed @cryptoexpert2000 is this still an issue?

ai-slave:
Still opens App Store instead of wallet for me

PR #589, opened 2021-08-24
Remove check that prevents linking on Android

tuomohopia:
This PR removes a check that causes the app to open a web link instead of linking to a wallet app on Android, with React Native.

The problem is described in detail here
https://github.com/WalletConnect/walletconnect-monorepo/issues/588
and is experienced by other people as well: https://github.com/WalletConnect/walletconnect-monorepo/issues/563

Removing this check causes the app to simply open the link on a supported wallet app on the same device.


cawfree (changes requested):
Can you `try` to open the URL immediately, and if this fails, resort to opening `https://walletconnect.org/wallets`?

PR #594, opened 2021-08-27
Ensure QR Code not shown in RN Web apps when on a mobile browser

META-DREAMER:
Previously the modal would render a QR code anytime react-native was running on web even if it was running on mobile Safari / Chrome etc. This fixes it by checking if the user agent device type corresponds to a mobile or tablet device (iOS or Android).

Test Plan: Open a react-native-web app with WalletConnect on mobile Safari and ensure a grid of wallet options is shown instead of a QR Code.

### Before and After
<p float="left">
  
  
</p>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,7 @@
 	"requires": true,
 	"packages": {
 		"": {
+			"name": "walletconnect-monorepo",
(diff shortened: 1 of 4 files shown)


ganchoradkov:
closing as stale

PR #607, opened 2021-09-18
Optional Callback when Connect Fails

tuomohopia:
# Connect Fail Fallback
> primarily for Android

This PR adds an optional callback to the Provider that gets invoked when opening the link fails.

On Android, by now many people have observed that for some reason React Native does not detect that `wc:..` links can be open by an installed app as `Linking.canOpenUrl(url)` produces `false` instead of `true` even when a registered app to open the link with is installed. With the current behavior, this leads to the library defaulting to open the Wallet Connect homepage, with the supported wallet page.

This PR gives the library user a choice to supply a custom callback that gets invoked instead.

The issue was originally raised here:
* https://github.com/WalletConnect/walletconnect-monorepo/issues/588
* https://github.com/WalletConnect/walletconnect-monorepo/pull/589

The idea for the custom callback originated here: https://github.com/WalletConnect/walletconnect-monorepo/issues/588#issuecomment-905824044

### DIFF
--- a/packages/helpers/react-native-dapp/README.md
+++ b/packages/helpers/react-native-dapp/README.md
@@ -38,7 +38,7 @@ export default function App(): JSX.Element {
     <WalletConnectProvider
       redirectUrl={Platform.OS === 'web' ? window.location.origin : 'yourappscheme://'}
       storageOptions= {{
-        asyncStorage AsyncStorage,
+        asyncStorage: AsyncStorage,
       }}>
       <>{/* awesome app here */}</>
     </WalletConnectProvider>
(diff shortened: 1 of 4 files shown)


tuomohopia:
@cawfree would you mind checking if I'm doing the right things here?

I have no idea how to actually test the library code against my app.

Moreover, I'm not quite sure if you'd like the optional callback typed here instead of where I put it: https://github.com/WalletConnect/walletconnect-monorepo/blob/0871582be273f8c21bb1351315d649ea47ee70b7/packages/helpers/types/index.d.ts#L240-L250

ganchoradkov:
closing as v1 is being sunset

PR #613, opened 2021-09-27
qrcode-modal works in electron

pedrorv0106:
QRCode modal does not work in electron. Because it checks only if it's only node or not when open or close dialog.
Added isElectron function to check if it's electron.


ganchoradkov:
closing as v1 is being sunset

PR #634, opened 2021-11-04
Automatic git hash

ilanolkies:
Hi! I added this piece of code to get the `GITHASH` from `git` if not set in env. I put the config value with more priority than the actual `git`

Ref: https://stackoverflow.com/questions/34518389/get-hash-of-most-recent-git-commit-in-node

```
GITHASH=fromenv npm start
```

Screen Shot 2021-11-04 at 01 33 26

```
npm start
```

Screen Shot 2021-11-04 at 01 33 39

### DIFF
--- a/servers/relay/src/config/index.ts
+++ b/servers/relay/src/config/index.ts
@@ -1,7 +1,18 @@
 import { THIRTY_DAYS } from "../constants";
 import { RelayModes } from "../types";
 
-const GITHASH = process.env.GITHASH || "0000000";
+let GITHASH = process.env.GITHASH
+
+if (!GITHASH) {
+  try {
+    GITHASH = require('child_process')
+      .execSync('git rev-parse HEAD')
+      .toString().trim()
+  } catch {
+    GITHASH = "0000000"
+  }
+}
+
 const VERSION = require("../../package.json").version || "0.0.0";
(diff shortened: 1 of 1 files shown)


sbc64:
Thanks for your contribution @ilanolkies,

This PR is helpful but only when the relay server is ran from within the git repository. In our current CI/CD pipeline the git repository is not included in the docker image of the relay so there would be no result from running `git rev-parse HEAD` because there is no git repository in the docker image. Also inside the docker image there is no `git` binary/command in the docker image, so when `child-process` calls the `git` command it would likely return an error on the command call.

If you really need the GITHASH environment variable right now you can do the following:

`GITHASH=$(git rev-parse HEAD) npm run start --prefix servers/relay`

image

ilanolkies:
Nice catch! I don’t need it 😊 I only want to make this tool more transparent. We were trying to find the hash you were deploying on your public bridge and was impossible since it answers 0000

08EA58B1-F0FE-4682-858E-A7B5E9AF1467
#

sbc64:
Solved in this PR by solving the scope of of `let` `in` in `default.nix`: https://github.com/WalletConnect/walletconnect-monorepo/pull/791/files#diff-245563af9e34754739b12b294f3324c1d42c116ace7ce8735e21dfbc82d8e064R68

PR #668, opened 2021-12-12
Rm query-string and use URLSearchParams

### LINKED ISSUE #667: query-string dependency can be removed in favor of native URLSearchParams
- https://caniuse.com/?search=URLSearchParams
- query-string makes 4.3% bundle size of 1.6.6 https://bundlephobia.com/package/@walletconnect/client@1.6.6

e.g. https://github.com/WalletConnect/walletconnect-monorepo/blob/7aeb3fe5dd0ec6c8f52a5d19d1dc69246e0076be/packages/helpers/utils/src/url.ts#L1

TimDaub:
In the @walletconnect/client@1.6.6, query-string accounts for 4.3% of
the overall bundle size. However, client-side support for
URLSearchParams has reached mainstream and so we can just use it instead
of including more bundle-size-increasing dependencies.

- Fixes #667

### DIFF
--- a/packages/helpers/utils/package.json
+++ b/packages/helpers/utils/package.json
@@ -67,8 +67,7 @@
     "@walletconnect/jsonrpc-utils": "^1.0.0",
     "@walletconnect/types": "^1.6.6",
     "bn.js": "4.11.8",
-    "js-sha3": "0.8.0",
-    "query-string": "6.13.5"
+    "js-sha3": "0.8.0"
   },
   "gitHead": "165f7993c2acc907c653c02847fb02721052c6e7"
 }
--- a/packages/helpers/utils/src/url.ts
+++ b/packages/helpers/utils/src/url.ts
@@ -1,5 +1,3 @@
-import * as queryStringUtils from "query-string";
-
 export function getQueryString(url: string): string {
   const pathEnd: number | undefined = url.indexOf("?") !== -1 ? url.indexOf("?") : undefined;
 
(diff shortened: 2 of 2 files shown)


pedrouid:
What's the support for NodeJS and React-Native like for this?

TimDaub:
MDN compat table from today: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

Screen Shot 2022-01-06 at 18 35 04

For react native, I'm using the following source: https://reactnative.dev/docs/javascript-environment#polyfills

Screen Shot 2022-01-06 at 18 38 30

- On Chrome, everything looks green and so I'd think that it's fully supported at a specific minimal version
- On RN iOS that uses JavaScriptCore, we can see in the compat-table supports everything for "Safari for iOS" and that `delete` isn't supported, though I'm not sure if "Safari for iOS" and "JavaScriptCore" are exactly the same.

I ended up opening an issue at https://github.com/kangax/compat-table/issues/1782 to try to understand where compatibility for the https://url.spec.whatwg.org/ URL What-WG standard is tracked.


Edit: [I asked over at kangax/compat-table](https://github.com/kangax/compat-table/issues/1782#issuecomment-1006804256) and it seems that to tell, we'd have to either run the jsc test ourselves or hope for an answer from apple. I'm waiting for some mails to come back...

pedrouid:
Thanks for looking into this 🙌  I'm not in a rush to ship 1.7.1 so I will wait for your confirmation and then include it in the next release (PR #688)

pedrouid:
I've published 1.7.1 with other fixes and merged this one into next tag to be published under 1.7.2-rc.0 for testing across platforms

PR #670, opened 2021-12-12
Replace bn.js with BigInt

### LINKED ISSUE #669: Replace BN.js with BigInt and parseInt
- https://bundlephobia.com/package/@walletconnect/core@1.6.6 bn.js takes a lot of space in the bundle
- It's not entirely clear if BN.js is totally necessary. It's main use case as I could see it is in converting from and to hexadecimal
- BigInt has good browser support: https://caniuse.com/?search=bigint

TimDaub:
- Fixes #669


TimDaub:
thanks 🤩

PR #688, opened 2022-01-05
v1.7.2

pedrouid:
- replace bn.js with BigInt
- replace query-string with URLSearchParams

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.7.0",
+  "version": "1.7.1-alpha.0",
   "packages": [
     "packages/**/*"
   ]
--- a/lerna.json
+++ b/lerna.json
@@ -1,4 +1,6 @@
 {
-  "version": "1.7.1",
-  "packages": ["packages/**/*"]
+  "version": "1.7.2-rc.0",
+  "packages": [
+    "packages/**/*"
+  ]
 }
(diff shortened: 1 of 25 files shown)


pedrouid:
@TimDaub you can see below a comparison between v1.7.1 and v1.7.2-rc.0 which includes both PRs you have submitted and we can see substantial bundle reduction 💪  Thanks a lot for your contributions!

Only thing missing now is testing agaisnt react-native applications to support as low as v0.60 potentially

Screenshot 2022-01-11 at 12 10 06

TimDaub:
wow this is super cool! thanks for sharing @pedrouid

xzilja:
@pedrouid Tested these cases with react native running on hermes engine

```
  console.log(parseInt('1010', 10))
  console.log(parseInt('0x1F7', 16))
  console.log(BigInt(9007199254740991).toString())
  console.log(BigInt(9007199254740991).toString(16))
  console.log(new URLSearchParams('q=URLUtils.searchParams&topic=api').toString())
```

`BigInt` - is not supported at all, related issue https://github.com/facebook/hermes/issues/510
`parseInt` - Is fine as expected
`URLSearchParams` - Doesn't error, but it returns empty result to me, where as I expected `q=URLUtils.searchParams&topic=api` (which happens in browser)

So overall this whole change would be breaking rn

pedrouid:
That's a shame! We will need to develop separate target libraries to support these changes for NodeJS and Browser while still using the current dependencies for React-Native

PR #816, opened 2022-02-24
fix(ethereum-provider): adds public sendAsync method to satisfy web3.js provider interface

bkrem:
## Issue

When passing the v2 `EthereumProvider` to the `Web3` constructor, Typescript will currently infer the wrong internal type for the provider (`WebsocketProvider` instead of `AbstractProvider`) and error with the following:

```ts
// Argument of type 'EthereumProvider' is not assignable to parameter of type 'provider'.
// Type 'EthereumProvider' is missing the following properties from type 'WebsocketProvider': 
// isConnecting, requestQueue, responseQueue, connection, and 5 more.
const provider = new Web3(ethereumProvider);
```

This is due to the `web3` library expecting the `sendAsync` method to be defined, while `request` is seen as optional:

```ts
// web3.js 1.7.0
export interface AbstractProvider {
    sendAsync(payload: JsonRpcPayload, callback: (error: Error | null, result?: JsonRpcResponse) => void): void;
    send?(payload: JsonRpcPayload, callback: (error: Error | null, result?: JsonRpcResponse) => void): void;
    request?(args: RequestArguments): Promise<any>;
    connected?: boolean;
  }
```

This issue does not arise when passing `EthereumProvider` to `ethers`, since the interface exposed by `ethers` is fully agnostic to the implemented RPC methods:

```ts
// ethers.js
export type ExternalProvider = {
    isMetaMask?: boolean;
    isStatus?: boolean;
    host?: string;
    path?: string;
    sendAsync?: (request: { method: string, params?: Array<any> }, callback: (error: any, response: any) => void) => void
    send?: (request: { method: string, params?: Array<any> }, callback: (error: any, response: any) => void) => void
    request?: (request: { method: string, params?: Array<any> }) => Promise<any>
}
```

## Solution

By providing a thin, callback-based `sendAsync` wrapper around `EthereumProvider.request`, we can satisfy the interface expectations for the `web3` provider with minimal changes and resolve the misleading type error.

### DIFF
--- a/packages/ethereum-provider/src/index.ts
+++ b/packages/ethereum-provider/src/index.ts
@@ -113,6 +113,15 @@ class EthereumProvider implements IEthereumProvider {
     return this.http.request(args);
   }
 
+  public sendAsync(
+    args: RequestArguments,
+    callback: (error: Error | null, response: any) => void,
+  ): void {
+    this.request(args)
+      .then(response => callback(null, response))
+      .catch(error => callback(error, undefined));
+  }
+
   get connected(): boolean {
     return (this.signer.connection as SignerConnection).connected;
   }
--- a/packages/ethereum-provider/test/index.spec.ts
+++ b/packages/ethereum-provider/test/index.spec.ts
@@ -12,8 +12,7 @@ import {
 
 import { WalletClient } from "./shared";
 
-import WalletConnectProvider from "../src";
-import { SIGNER_EVENTS } from "../../signer-connection/dist/cjs";
+import EthereumProvider from "../src";
 
 const CHAIN_ID = 123;
 const PORT = 8545;
@@ -82,10 +81,10 @@ const TEST_ETH_TRANSFER = {
   data: "0x",
 };
 
-describe("WalletConnectProvider", function() {
+describe("EthereumProvider", function() {
   this.timeout(30_000);
   let testNetwork: TestNetwork;
-  let provider: WalletConnectProvider;
+  let provider: EthereumProvider;
   let walletClient: WalletClient;
   let walletAddress: string;
   let receiverAddress: string;
@@ -95,7 +94,7 @@ describe("WalletConnectProvider", function() {
       port: PORT,
       genesisAccounts: [ACCOUNTS.a, ACCOUNTS.b],
     });
-    provider = new WalletConnectProvider(TEST_PROVIDER_OPTS);
+    provider = new EthereumProvider(TEST_PROVIDER_OPTS);
     walletClient = await WalletClient.init(provider, TEST_WALLET_CLIENT_OPTS);
     walletAddress = walletClient.signer.address;
     receiverAddress = ACCOUNTS.b.address;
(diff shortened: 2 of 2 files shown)


bkrem:
I wasn't able to run the tests for `EthereumProvider` locally since the `make dev` setup for the relay and redis servers errored out for me:

```
docker run --name builder --rm -v nix-store:/nix -v /Users/bk/repos/walletconnect/walletconnect-monorepo:/src -w /src nixos/nix nix-shell -p bash --run "nix-build ./ops/waku-docker.nix --argstr wakuVersionTag v0.5.1 --argstr nixNimRepoSha256 0k55hw1wqcyrpf9cxchhxdb92p75mmskkpvfn1paivl1r38pyb4a && cp -r -f -L result build/build-img-waku && rm -rf result"
warning: error: unable to download 'https://github.com/WalletConnect-Labs/nix-nim-waku/archive/v0.5.1.tar.gz': HTTP error 300 ('')

       response body:

       the given path has multiple possibilities: #<Git::Ref:0x00007fa53f04f1a0>, #<Git::Ref:0x00007fa53f04efc0>; retrying in 316 ms
warning: error: unable to download 'https://github.com/WalletConnect-Labs/nix-nim-waku/archive/v0.5.1.tar.gz': HTTP error 300 ('')

       response body:

       the given path has multiple possibilities: #<Git::Ref:0x00007fa53f04f1a0>, #<Git::Ref:0x00007fa53f04efc0>; retrying in 705 ms
warning: error: unable to download 'https://github.com/WalletConnect-Labs/nix-nim-waku/archive/v0.5.1.tar.gz': HTTP error 300 ('')
```

Maybe @sbc64 knows how to work around this?

xzilja (approved):
Looks good 👍

pedrouid:
Hey @bkrem, all looks good to me. But could you check what happened to the package-lock.json? Are you using a different version of NPM? 

This PR should be changing around 30-40 lines but it shows a diff of thousands because of the package-lock.

Other than that it looks good to mere

pedrouid:
I've also pushed a commit to v1.0 that includes this same fix (https://github.com/WalletConnect/walletconnect-monorepo/commit/a99259ef26f34a0874791cb4ca63ad89b9d72d44)

bkrem:
> But could you check what happened to the package-lock.json? Are you using a different version of NPM?

I'm also not sure what that massive diff was about, I'm using `node@16.3.0/npm@7.22.0` which should be `lockfileVersion: 2` as specified in the lockfile. I've rebased and removed that lockfile commit since it doesn't really belong in this PR anyways 👍 

Which version of NPM are you using by default @pedrouid? Maybe this makes sense for us to pin via the [`engines` field](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines) to align envs between everyone working on the packages.

> I've also pushed a commit to v1.0 that includes this same fix 

Legend, thank you for already handling that 💯

Merging is still blocked for me here due to the broken CI step that's marked as required.

PR #833, opened 2022-03-04
fix(client): handles missing `projectId` gracefully with related error

bkrem:
## Issue

The need for a `projectId` to be passed for the v2 client seems to trip up quite a few people and it's not obvious what the issue is when omitting it, since the resulting error message comes from deeper in the stack (`ws.ts`) and does not indicate what the issue is:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'message')
```

## Solution

By adding an explicit guard statement to check for the required `opts.projectId` value, we can "throw gracefully" and provide an error message indicating that the `projectId` is missing via the existing `ERROR.MISSING_OR_INVALID` enum:

```
Error: Missing or invalid projectId
```

I've tried to place the guard statement at the earliest point possible (after `this.logger` has been bound, but before projectId is passed into further methods that assume it is defined), but if this should live somewhere else please point it out.


bkrem:
Closing this for now since it's out-of-sync with upcoming changes.

Will keep the branch and rework this for v2.0-rc and beyond if needed.

PR #855, opened 2022-03-14
Remove legacy examples

xzilja:


servers/relay/test/http.spec.ts
```diff
@@ -1,11 +1,11 @@
-import "mocha";
 import axios, { AxiosInstance } from "axios";
-import { use, expect } from "chai";
+import { expect, use } from "chai";
 import chaiHttp from "chai-http";
 import { Agent } from "https";
-
+import "mocha";
```
bkrem: This seems like an unrelated change, was this intentional?

bkrem (approved):
🧹 ✨

servers/relay/test/http.spec.ts
xzilja: prettier slipped through when I visited this file

pedrouid (approved):
👍

PR #895, opened 2022-03-31
Setup docker and docker compose to build docker images 

### LINKED ISSUE #727: Use more traditional docker build methods for building the relay-server

### LINKED ISSUE #886: Replace `nix-nim-waku` with `statusteam/nim-waku:deploy-v2-prod`
Use the status team docker container for the waku node.

https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/ops/waku-docker.nix#L12-L85

It should have an entry script that is similar to the above link minus the scoped path invocations that nix has that look this: `"${pkg.coreutils}/bin/od"`

ffrizzo:


build/package/docker-compose.override.yml
```diff
@@ -0,0 +1,98 @@
...
+    ports:
+      - 6379:6379
+    volumes:
+      - redis:/data
+    healthcheck:
+      test: ["CMD", "redis-cli","ping"]
+      interval: 5s
+      timeout: 5s
+      retries: 5 
+  waku-store:
```
ffrizzo: @sbc64 I know we have chat about this one and want to check if make sense to remove waku from this setup

PR #924, opened 2022-04-07
refactor: removes dependency on `storage` controller

### LINKED ISSUE #912: Remove dependency on storage.ts in client/controllers
## Steps

- [x] Move methods from storage.ts into their respective controllers
- [x] Rewire existing `this.storage.xyz` references to use the methods in the respective controllers instead
- [x] Remove `storage.ts/this.storage` and associated types completely

## Affected controllers

- [x] - crypto.ts
- [x] - store.ts
- [x] - expirer.ts
- [x] - history.ts
- [x] - messages.ts
- [x] - subscriptions.ts

### Shared requirements

- [x] Storage prefix should be inherited by the parent module.
- [x] Storage path should still derive from context, i.e. `wc@2:client:...`, `wc@2:core:...`

bkrem:
Resolves #912 

### Key changes

- [x] Moved methods from `storage.ts` into their respective controllers
- [x] Rewired existing `this.[...].storage.xyz` references to use the methods in the respective controllers instead
- [x] Removed `storage.ts/this.storage` and associated types completely

### Checks

- Ran `packages/client` unit tests locally ✅ 
- Consumed the symlinked client in `react-dapp-v2` example, storage behaves as before ✅


packages/client/src/controllers/crypto.ts
```diff
@@ -30,6 +35,10 @@ export class KeyChain implements IKeyChain {
     return getLoggerContext(this.logger);
   }
 
+  get storageKey(): string {
+    return this.client.storagePrefix + this.version + "//" + formatStorageKeyName(this.context);
```
bkrem, line 39: This can probably be abstracted into a reusable util function, I wanted to keep things simple during refactoring and not also have to symlink in `@walletconnect/utils`

IljaDaderko:
Will deep dive tomorrow, at first glance one thing that stands out are files like relayer / subscriber, I noticed you added `client` to constructors i.e.

https://github.com/WalletConnect/walletconnect-monorepo/pull/924/files#diff-6d4340535a8b22f174b74546eb7837daea835280a29c2b1ba3654cb469811339R61

however client is never used anywhere within those files. So I assume, somewhere else there are controllers accessing it like `this.relayer.client` or `this.subscriber.client`? 

I feel like this is super confusing and we should avoid it if possible. For now a comment explaining why client is needed in those constructors could be sufficient enough for us to re-do this latter on during cleanup I guess?

pedrouid (approved):
👍

Issue #937, opened 2022-04-11
Refactor storage dependency model, target by platform and suggest additional dependencies


chadyj:
Depends on: 
- https://github.com/WalletConnect/walletconnect-utils/issues/12

---

- [x] Make dependencies like `react-native-async-storage` and `better-sqlite3` externals/optionals (so that webpack doesn't bundle them)
- [x] Import and use expected external dependency (if any) for a given platform target, rather than using an injected dependency.
- [x] Update docs to reflect approach of adding additional external dependencies as needed 
- [x] Update client instantiation of `keyvaluestorage`, remove options from constructor (as we will define these on our own now)

xzilja:
We want to take full control of storage mechanism used on all our platforms, which means that we'll no longer ask / expect users to pass their own storage library during initialisation. Similarly we will be using default database name. These require few changes:

- [ ] Remove options from constructor (as we will define these on our own now)
- [ ] Make dependencies like `react-native-async-storage` and `better-sqlite` externals (so that webpack doesn't bundle them)
- [ ] Detect which platform keyvaluestorage package is used on
- [ ] Based on platform import correct storage package
- [ ] Documentation needs to be updated to reflect these changes as users will now have to install those external packages

bkrem:
@IljaDaderko I've merged your bullet points into the initial ones I had in the description. Only thing I've changed is instead of:


> - Detect which platform keyvaluestorage package is used on
> - Based on platform import correct storage package


we have 

> - Import and use expected external dependency for a given platform target, rather than using an injected dependency.

since `keyvaluestorage` already defines platform-specific entry targets as Pedro mentioned during sync: https://github.com/WalletConnect/walletconnect-utils/blob/master/misc/keyvaluestorage/package.json#L24

bkrem:
Closing this since all tasks have been addressed and relevant PRs merged 👍

PR #939, opened 2022-04-12
refactor(client): removes `storage` from top-level `ClientOptions`, use `@walletconnect/keyvaluestorage`

bkrem:
Part of:
- #937 

Depends on:

- [x] https://github.com/WalletConnect/walletconnect-utils/pull/13
- [x] Publishing a version of `@walletconnect/keyvaluestorage` to NPM

### Key changes

- Replace external `keyvaluestorage` pkg with new `@walletconnect/keyvaluestorage@1.0.0`
- Removes `storage` from top-level `ClientOptions` interface
- Updates tests
- Fixes CI by using `npm run bootstrap -- --ci` in the dockerfile, to ensure we install optional peerDeps on CI runs

### DIFF
--- a/ops/package/Dockerfile-dev
+++ b/ops/package/Dockerfile-dev
@@ -10,8 +10,9 @@ FROM base as build
 WORKDIR /monorepo
 
 COPY ./ ./
-RUN npm install 
-RUN npm run bootstrap
+RUN npm ci
+RUN npm run bootstrap -- --ci
 RUN npm run build
 
 CMD ["node", "-v"]
+
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,6 +39,7 @@
 				"@walletconnect/jsonrpc-types": "^1.0.0",
 				"@walletconnect/jsonrpc-utils": "^1.0.0",
 				"@walletconnect/jsonrpc-ws-connection": "^1.0.0",
+				"@walletconnect/keyvaluestorage": "^1.0.0",
 				"@walletconnect/logger": "^1.0.0",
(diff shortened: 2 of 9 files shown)


bkrem:
CI is resolving the optional peer deps specified by `@walletconnect/keyvaluestorage` as expected now, but still seeing a single test timeout in the `client` package, all other tests on client are passing, which is super weird.

Will try bumping the timeout, but this seems like something else, since running the docker scripts the same way locally results in everything passing as expected.

bkrem:
Ok tried upping the standard timeout for `client` tests to 20 seconds, but still no clue what is causing the CI runs to timeout one way on one run, another way on another run.

E.g. on some runs the only test that fails is:

```
  42 passing (37s)
  1 failing

  1) Relayer
       A pings B after B socket reconnects:
     Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/monorepo/packages/client/test/relayer.spec.ts)
```

On other ones it was:

```
  42 passing (19s)
  1 failing

  1) Pairing
       A pings B with existing pairing:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/monorepo/packages/client/test/pairing.spec.ts)
```

If I run the CI steps locally (`make dev && make test-relay`) everything passes as expected, see attached log:
[make-test-relay-local.log](https://github.com/WalletConnect/walletconnect-monorepo/files/8512335/make-test-relay-local.log)

xzilja (commented):
Looks good 👍

PR #968, opened 2022-04-23
WalletConnectCore

pedrouid:


packages/client/src/controllers/engine.ts
```diff
@@ -57,7 +57,7 @@ export default class Engine extends IEngine {
       uri = newUri;
     }
 
-    const publicKey = await this.client.crypto.generateKeyPair();
+    const publicKey = await this.client.core.crypto.generateKeyPair();
```
xzilja, line 60: Mixed feelings about this. Now that we have a lot of stuff abstracted away into core can we pass it to engine directly so this becomes `this.core.crypto`? In fact can core be initialised within engine to begin with?

packages/client/src/controllers/expirer.ts
```diff
@@ -18,9 +18,9 @@ export class Expirer extends IExpirer {
 
   private cached: ExpirerTypes.Expiration[] = [];
 
-  constructor(public client: IClient, public logger: Logger) {
-    super(client, logger);
-    this.client;
+  constructor(public core: ICore, public logger: Logger) {
+    super(core, logger);
+    this.core;
```
xzilja, line 23: This seems to not be doing anything?

packages/core/.watchmanconfig
```diff
@@ -0,0 +1 @@
+{}
```
xzilja, line 1: This is more for my info. What uses watchman in monorepo?

packages/core/src/constants/crypto.ts
```diff
@@ -0,0 +1 @@
+export const CRYPTO_CONTEXT = "crypto";
```
xzilja, line 1: I noticed that we have few files like this, that just export on maybe 2 contexts for storage. Perhaps all of these can be combined into one file?

packages/core/src/controllers/subscriber.ts
```diff
@@ -280,11 +237,14 @@ export class Subscriber extends ISubscriber {
   }
 
   private async setRelayerSubscriptions(subscriptions: SubscriberTypes.Active[]): Promise<void> {
-    await this.client.storage.setItem<SubscriberTypes.Active[]>(this.storageKey, subscriptions);
+    await this.relayer.core.storage.setItem<SubscriberTypes.Active[]>(
```
xzilja, line 240: `this.relayer.core.storage` This looks a bit unexpected, I'd expect it to be like `this.core.storage` ? i.e. relayer should come from core?

packages/types/src/core.ts
```diff
@@ -0,0 +1,42 @@
...
+    projectId?: string;
+    name?: string;
+    relayUrl?: string;
+    logger?: string | Logger;
+    keychain?: IKeyChain;
+    storageOptions?: KeyValueStorageOptions;
+  }
+}
+
+export abstract class ICore extends IEvents {
```
xzilja, line 21: Why does core need to extend events?

xzilja (approved):
Looks much nicer now 🚀 I left few questions / suggestions

PR #977, opened 2022-04-25
Changed check for RPC JSON response success [v1.0]

### LINKED ISSUE #930: (REACT-NATIVE) JSON RPC response format is invalid
I tried to use the `sendCustomRequest `and the method is `wallet_switchEthereumChain` to request to switch the network on the metamask mobile, when I press switch network button on the metamask app it return me this error on my dApp `JSON RPC response format is invalid`. Here is my sample code

```
 const requestSwitchNetwork = async () => {
    try {
      await connector.sendCustomRequest({
        id: 0,
        jsonrpc: '2.0',
        method: 'wallet_switchEthereumChain',
        params: [ { chainId: ethers.utils.hexlify(CHAIN_ID) } ],
      });
    } catch (error) {
      // Always getting error
      console.log(error);
    }
  };
```
I tried this on the metamask API playground and it works fine.
https://metamask.github.io/api-playground/api-documentation/#wallet_switchEthereumChain

poriyaalarsamvelraja:
Fixes  #930

According to[ RPC JSON format 2.0 ](https://www.jsonrpc.org/specification#response_object)

```
This member MUST NOT exist if there was an error invoking the method.
```

The condition `if(payload.result)` would fail if the result is `null` which is true for many success cases.

for example for [wallet_switchEthereumChain](https://docs.metamask.io/guide/rpc-api.html#wallet-switchethereumchain)

### DIFF
--- a/packages/clients/core/src/index.ts
+++ b/packages/clients/core/src/index.ts
@@ -1003,7 +1003,7 @@ class Connector implements IConnector {
         this._handleSessionResponse(error.message);
         return;
       }
-      if (payload.result) {
+      if (isJsonRpcResponseSuccess(payload)) {
         this._handleSessionResponse(errorMsg, payload.result);
       } else if (payload.error && payload.error.message) {
         this._handleSessionResponse(payload.error.message);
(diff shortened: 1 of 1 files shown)


poriyaalarsamvelraja:
Thanks @pedrouid

pedrouid:
@poriyaalarsamvelraja I have published v1.7.8 with these changes

poriyaalarsamvelraja:
Thanks Mate

PR #980, opened 2022-04-26
fix(engine): handles unknown topic scenario for `Engine.ping`, reenables tests

bkrem:
- Added standard error handling for the unknown topic case in `Engine.ping`
- Re-enabled relevant tests since the issue of false positives/passes is now removed.
- Expanded `initTwoClients` test util fn to accept `ClientOptions` overrides
- Found an issue around restoring to `this.client.pairing` from persistence, updated the relevant comment in the tests

### DIFF
--- a/packages/client/test/client.spec.ts
+++ b/packages/client/test/client.spec.ts
@@ -96,9 +96,7 @@ describe("Client", () => {
         } = await testConnectMethod(clients);
         await clients.B.ping({ topic });
       });
-      // FIXME: Bug: This is failing because session topics are being restored to `client.pairing`.
-      // Restoring session topics works (see equivalent test below).
-      it.skip("clients can ping each other after restart", async () => {
+      it("clients can ping each other after restart", async () => {
         const beforeClients = await initTwoClients({
           storageOptions: { database: TEST_CLIENT_DATABASE },
         });
--- a/packages/client/test/client.spec.ts
+++ b/packages/client/test/client.spec.ts
@@ -42,11 +42,10 @@ describe("Client", () => {
         await expect(clients.A.pairing.get(topic)).to.eventually.be.rejectedWith(
           `No matching pairing with topic: ${topic}`,
         );
-        // FIXME: engine.ping is not handling this base case currently, this doesn't throw/reject.
-        // const promise = clients.A.ping({ topic });
-        // await expect(promise).to.eventually.be.rejectedWith(
-        //   `No matching pairing with topic: ${topic}`,
-        // );
+        const promise = clients.A.ping({ topic });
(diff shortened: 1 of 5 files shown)


xzilja (approved):
👌

packages/client/test/client.spec.ts
```diff
@@ -93,9 +96,12 @@ describe("Client", () => {
         } = await testConnectMethod(clients);
         await clients.B.ping({ topic });
       });
-      // TODO: this test requires `engine.ping` to handle unknown topics to avoid false positives.
-      it.skip("clients ping each other after restart", async () => {
-        const beforeClients = await initTwoClients();
+      // FIXME: Bug: This is failing because session topics are being restored to `client.pairing`.
+      // Restoring session topics works (see equivalent test below).
+      it("clients can ping each other after restart", async () => {
```
bkrem, line 99: @IljaDaderko @pedrouid 

I tried to debug this yesterday but couldn't get to the bottom of it, even with lots of logging in `Store`.

1. session/pairing topics are initially correct during `beforeClients`
2. when we re-initialize/restore here as `afterClients` `client.pairing.keys` is restored with the same topic hashes as `client.session.keys` 
3. -> test fails.

Logs of `this.client.session.keys` and `this.client.pairing.keys` before and after restoring in this test:

```
PING FOR TOPIC:  ec32c1ee93827a31f4fe0780b9db82d1ee5ccaae6753f61795745f57189a8dbf
session topics:  [ '17d720294eb8ab4e6eb62363520d6cd5d6304986ff2e19def26f65d5ecd618cd' ]
pairing topics:  [ 'ec32c1ee93827a31f4fe0780b9db82d1ee5ccaae6753f61795745f57189a8dbf' ]
>>>>>> RESTART <<<<<<<
PING FOR TOPIC:  ec32c1ee93827a31f4fe0780b9db82d1ee5ccaae6753f61795745f57189a8dbf
session topics:  [ '17d720294eb8ab4e6eb62363520d6cd5d6304986ff2e19def26f65d5ecd618cd' ]
pairing topics:  [ '17d720294eb8ab4e6eb62363520d6cd5d6304986ff2e19def26f65d5ecd618cd' ]
```

bkrem: Somewhat ironically, I bumped into the root cause of this while working on https://github.com/WalletConnect/walletconnect-monorepo/pull/983:

Since `Store` now lives in core, using `formatStorageKey(this.context)` quietly broke storage keys, since `this.context` inside core is simply e.g. `client_a`, rather than `session`, `pairing` etc. 
Using `this.name` instead to form the suffix fixes this bug.

In short: this issue is fixed by https://github.com/WalletConnect/walletconnect-monorepo/pull/983

PR #982, opened 2022-04-26
[v2 rc] Namespaces

xzilja:


packages/client/src/controllers/engine.ts
```diff
@@ -731,13 +686,16 @@ export class Engine extends IEngine {
...
   private onSessionRequest: EnginePrivate["onSessionRequest"] = async (topic, payload) => {
     // TODO(ilja) validation
-    const { methods } = await this.client.session.get(topic);
+    const { namespaces } = await this.client.session.get(topic);
     const { id, params } = payload;
     const { chainId, request } = params;
-    if (!methods.includes(params.request.method)) {
-      await this.sendError(id, topic, ERROR.UNAUTHORIZED_JSON_RPC_METHOD.format());
-    } else if (!params.chainId) {
+    const isChain = Boolean(chainId && namespaces.filter(n => n.chains.includes(chainId)).length);
```
bkrem, line 698: Could use `Array.some` here to get a boolean rather than going via implicit casting from length to boolean:

```suggestion
    const isChain = Boolean(chainId && namespaces.some(n => n.chains.includes(chainId)));
```

bkrem, line 698: Could use `Array.some` here to get a boolean rather than going via implicit casting from length to boolean:

```suggestion
    const isChain = Boolean(chainId && namespaces.some(n => n.chains.includes(chainId)));
```

packages/client/src/controllers/engine.ts
```diff
@@ -731,13 +686,16 @@ export class Engine extends IEngine {
...
+    const isMethod = Boolean(namespaces.filter(n => n.chains.includes(request.method)).length);
```
bkrem, line 699: Similarly:

```suggestion
    const isMethod = namespaces.some(n => n.chains.includes(request.method));
```

packages/utils/src/misc.ts
```diff
@@ -123,6 +128,11 @@ export function formatStorageKeyName(context: string): string {
   return parseContextNames(context).join(COLON);
 }
 
+// -- object ------------------------------------------------ //
+export function isNamespaceEqual(a: SessionTypes.Namespace, b: SessionTypes.Namespace) {
+  return JSON.stringify(a) === JSON.stringify(b);
```
bkrem, line 129: This doesn't account for key ordering in the objects, returning false when the namespaces are effectively equal e.g.

```
a = {chains: ["chain"], events: ["event"], methods: ["method"] }
b = {chains: ["chain"],  methods: ["method"], events: ["event"] }

-> returns `false`
```

bkrem: I'd say this is a case for simply adding `lodash.isEqual` to take deep object comparisons off our hands (https://lodash.com/docs/4.17.15#isEqual).

Otherwise it's a matter of manually iterating each key and in turn iterating the string arrays between e.g. `a.chains` and `b.chains` etc

xzilja:
@bkrem good suggestions 👌 I am addressing these on next PR (expiry)
https://github.com/IljaDaderko/walletconnect-monorepo/commit/80040bdeabb132a1423ea51c4c9e20fa61bfb06d

I noticed that we are no longer using `lodash.union` so removed it and to work around not adding `isEqual`, am sorting objects like this 

```ts
export function isNamespaceEqual(a: SessionTypes.Namespace, b: SessionTypes.Namespace) {
  const sortedA = Object.fromEntries(Object.entries(a).sort());
  const sortedB = Object.fromEntries(Object.entries(b).sort());

  return JSON.stringify(sortedA) === JSON.stringify(sortedB);
}
```

bkrem:
@IljaDaderko 

Sorry but this still has edge cases since it only accounts for the keys and not different ordering inside of the arrays:
https://codesandbox.io/s/vibrant-waterfall-zoofn0?file=/src/index.js

Something like `lodash.isEqual` is tiny and takes care of all these edge cases for us, rolling your own function effectively means writing all the tests to verify it too since JS object comparisons are a PITA.

xzilja:
@bkrem ah, you are totally right. Will add lodash's utility in then, otherwise helper function will be massive

bkrem:
> Will add lodash's utility in then, otherwise helper function will be massive

Cheers 👍 yeah been there done that. Attempting to write your own implementations of standard JS utilities is a bad idea 9 times out of 10 unless you're willing to reimplement the lodash test suite yourself to avoid the bugs 😰

bkrem (approved):

PR #983, opened 2022-04-26
refactor(controllers): use `this.name` to form `storageKey` suffix

bkrem:
Also:
- chore(utils): removes now unused `formatStorageKeyName`

Since the keys are no longer based on parent context, this changes the storage key suffix in the following ways:

```
crypto:keychain -> keychain
relayer:messages -> messages
relayer:subscription -> subscription
...
```


xzilja (approved):
Nice 👍

PR #985, opened 2022-04-26
[v2 rc] Expirer

xzilja:
Review after namespace change is merged https://github.com/WalletConnect/walletconnect-monorepo/pull/982

### DIFF
--- a/packages/client/src/controllers/engine.ts
+++ b/packages/client/src/controllers/engine.ts
@@ -76,6 +76,7 @@ export class Engine extends IEngine {
       else if (data) {
         data.self.publicKey = publicKey;
         await this.client.session.set(data.topic, data);
+        await this.setExpiry(data.topic, data.expiry);
         resolve(data);
       }
     });
@@ -96,7 +97,7 @@ export class Engine extends IEngine {
     await this.client.pairing.set(topic, pairing);
     await this.client.core.crypto.setSymKey(symKey, topic);
     await this.client.core.relayer.subscribe(topic, { relay });
-    // TODO(ilja) this.expirer / timeout pairing ?
+    await this.setExpiry(topic, expiry);
 
     return pairing;
   };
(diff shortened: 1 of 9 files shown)


packages/client/src/controllers/engine.ts
```diff
@@ -638,10 +672,8 @@ export class Engine extends IEngine {
   ) => {
     // TODO(ilja) validation
     const { id } = payload;
-    await this.client.core.relayer.unsubscribe(topic);
     await this.sendResult<"wc_sessionDelete">(id, topic, true);
-    await this.client.session.delete(topic, ERROR.DELETED.format());
-    await this.client.core.crypto.deleteSymKey(topic);
```
bkrem, line 644: `this.deleteSession` now calls `this.client.core.crypto.deleteKeyPair(topic)` internally instead of `this.client.core.crypto.deleteSymKey(topic)` as shown here previously. 

Is that intentional?

packages/client/src/controllers/engine.ts
```diff
@@ -299,16 +308,41 @@ export class Engine extends IEngine {
...
 
     return { topic, uri };
   }
 
   private activatePairing: EnginePrivate["activatePairing"] = async topic => {
-    await this.client.pairing.update(topic, {
-      active: true,
-      expiry: calcExpiry(THIRTY_DAYS),
-    });
+    await this.client.pairing.update(topic, { active: true });
```
pedrouid, line 317: expiry was removed from here… this looks like a typo

packages/client/src/controllers/engine.ts
xzilja: Yes, however this is to be confirmed with @pedrouid Before deleting session and pairing called `deleteSymKey` for both, but I believe it needs to be `deleteKeyPair` for session and `deleteSymKey` for pairing

packages/client/src/controllers/engine.ts
xzilja: Its not, it is handled within `this.setExpiry`. that's called right after. Purpose of this is to have call just 1 method to simultaneously update pairing / session expiry as well as set them on expirer

bkrem (approved):

packages/client/src/controllers/engine.ts
pedrouid: you always need to `deleteSymKey(topic)` for both pairing and session

however for session you also have a keypair mapped by publicKey under `self` which needs to be deleted with `deleteKeyPair(publicKey)`

packages/client/src/controllers/engine.ts
pedrouid: gotcha. marked as resolved

PR #989, opened 2022-04-27
[v2 rc] storagePrefix

pedrouid:


packages/client/test/client.spec.ts
```diff
@@ -1,3 +1,4 @@
+import { delay } from "@walletconnect/time";
```
bkrem: Accidental import?

packages/types/src/store.ts
```diff
@@ -13,7 +13,13 @@ export abstract class IStore<Key, Value> {
 
   public abstract readonly values: Value[];
 
-  constructor(public core: ICore, public logger: Logger, public name: string) {}
+  constructor(
+    public core: ICore,
+    public logger: Logger,
+    public name: string,
+    // @ts-ignore
```
bkrem, line 20: Is the `@ts-ignore` still needed?

bkrem (approved):

xzilja (approved):
Looks good, but can we not just

```ts
  get storageKey(): string {
    return CLIENT_STORAGE_PREFIX + this.version + "//" + this.name;
  }
```

And thus get rid of `private storagePrefix` in all classes? As well as code that assigns it

packages/client/test/client.spec.ts
pedrouid: Yes, I was trying to fix ping test

packages/types/src/store.ts
pedrouid: It wouldn’t compile without it

pedrouid:
> Looks good, but can we not just
> 
> ```ts
>   get storageKey(): string {
>     return CLIENT_STORAGE_PREFIX + this.version + "//" + this.name;
>   }
> ```
> 
> And thus get rid of `private storagePrefix` in all classes? As well as code that assigns it

True, but made it consistent with store.ts which needs it

packages/client/test/client.spec.ts
bkrem:
```suggestion

```

Issue #915, opened 2022-04-06
Improve unit tests, add test coverage tooling


chadyj:
Improve/create unit tests for:

- existing controllers that will remain in v2.0-rc
- @walletconnect/utils
- add test coverage tooling

PR #992, opened 2022-04-28
test(core): sets up test suites for core controllers

bkrem:
Part of:

- #915 

## Key changes
- Tests public interfaces of core controllers
- Avoids dependence on interactions with the relay-server by mocking specific functions where needed.

## Test suites implemented
- [x] core.spec.ts
- [x] keychain.spec.ts (only a `storageKey` check for now since it's a dead simple controller)
- [x] messages.spec.ts
- [x] publisher.spec.ts
- [x] relayer.spec.ts
- [x] store.spec.ts
- [x] subscriber.spec.ts

## Remaining

- [x] crypto.spec.ts (see https://github.com/WalletConnect/walletconnect-monorepo/pull/1033)

### DIFF
--- a/packages/core/test/keychain.spec.ts
+++ b/packages/core/test/keychain.spec.ts
@@ -1,0 +1,31 @@
+import "mocha";
+import { getDefaultLoggerOptions } from "@walletconnect/logger";
+import pino from "pino";
+
+import {
+  Core,
+  CORE_DEFAULT,
+  CORE_STORAGE_PREFIX,
+  KeyChain,
+  KEYCHAIN_CONTEXT,
+  KEYCHAIN_STORAGE_VERSION,
+} from "../src";
+import { expect, TEST_CORE_OPTIONS } from "./shared";
+import { ICore } from "@walletconnect/types";
+
+describe("Keychain", () => {
+  const logger = pino(getDefaultLoggerOptions({ level: CORE_DEFAULT.logger }));
+
+  let core: ICore;
+
+  beforeEach(() => {
+    core = new Core(TEST_CORE_OPTIONS);
+  });
+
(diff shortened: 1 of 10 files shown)


packages/core/test/keychain.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+} from "../src";
+import { expect, TEST_CORE_OPTIONS } from "./shared";
+import { ICore } from "@walletconnect/types";
+
+describe("Keychain", () => {
+  const logger = pino(getDefaultLoggerOptions({ level: CORE_DEFAULT.logger }));
+
+  let core: ICore;
+
+  beforeEach(() => {
```
xzilja: This looks like core can be instantiated on the same level as logger?

xzilja (approved):
Good stuff, love having many tests for core now 👍 Only comment

I think in most cases `beforeEach` can be replaced with before as you can re-use same instances of controllers

packages/core/test/keychain.spec.ts
bkrem: Yeah definitely overkill here with the `beforeEach`, will simplify 👍 was copied across from another test

bkrem:
> Good stuff, love having many tests for core now

We're up 43 tests on core now from nothing 🙌 

> I think in most cases beforeEach can be replaced with before as you can re-use same instances of controllers

While this probably works fine with a lot of the tests, I'd prefer to get a fresh instance with a clean state between each test since it's easier to reason about and makes sure each test is truly atomic. 
Otherwise you end up with odd situations where you have to account for another item in one of the controller's internal `Map` objects that's there from a previous `.set`, which then breaks a test if another one is removed etc

PR #997, opened 2022-04-29
[v2 rc] Async and Init cleanup

xzilja:
- Remove unnecessary async / awaits
- Add missing awaits
- Re-do initialisation (init) methods to not rely on events, add additional checks and throw errors where necessary

### DIFF
--- a/packages/core/src/controllers/subscriber.ts
+++ b/packages/core/src/controllers/subscriber.ts
@@ -36,6 +36,8 @@ export class Subscriber extends ISubscriber {
 
   private cached: SubscriberTypes.Active[] = [];
 
+  private enabled = false;
+
   constructor(public relayer: IRelayer, public logger: Logger) {
     super(relayer, logger);
     this.relayer = relayer;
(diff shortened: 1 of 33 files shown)


packages/utils/src/error.ts
```diff
@@ -57,6 +57,8 @@ export const ERROR_TYPE = enumify({
   UNSUPPORTED_NOTIFICATION: "UNSUPPORTED_NOTIFICATION",
   UNSUPPORTED_SIGNAL: "UNSUPPORTED_SIGNAL",
   USER_DISCONNECTED: "USER_DISCONNECTED",
+  // 8000 (Not Initialized)
+  NOT_INITIALIZED: "NOT_INITIALIZED",
```
bkrem: Doesn't this belong in the internal section i.e. `1xxx`? Why are we creating a new category for this?

packages/utils/src/error.ts
```diff
@@ -549,6 +551,16 @@ export const ERROR: Record<ErrorType, Error> = {
       message: ERROR[ERROR_TYPE.USER_DISCONNECTED].stringify(params),
     }),
   },
+  // 8000 (Not Initialized)
+  [ERROR_TYPE.NOT_INITIALIZED]: {
+    type: ERROR_TYPE.NOT_INITIALIZED,
+    code: 8000,
+    stringify: (params?: any) => `Not Initialized ${params ? `: ${params?.toString()}` : ""}`,
```
bkrem:
```suggestion
    stringify: (params?: any) => `${params ? `: ${params?.toString()}` : ""} was not initialized`,
```

packages/core/src/controllers/subscriber.ts
```diff
@@ -144,29 +135,22 @@ export class Subscriber extends ISubscriber {
 
   private onEnable() {
     this.cached = [];
-    this.events.emit(SUBSCRIBER_EVENTS.enabled);
```
bkrem, line 147: Are we sure this isn't needed? Are other controllers relying on this being emitted?

cc @pedrouid

xzilja: Not needed, this was internal event signifying that initialization completed. All of these were replaced with a boolean flag instead and made consistent to not rely on events like this in all controllers. I'll also remove this one from the list of `SUBSCRIBER_EVENTS`

packages/utils/src/error.ts
xzilja: Moved to internal and changed code 👍

packages/utils/src/error.ts
xzilja: Updated 👍

PR #1017, opened 2022-05-04
Check if node app is not an electron app

### LINKED ISSUE #1007: WalletConnect modal doesn't open fine for Electron apps
**Is your feature request related to a problem? Please describe.**

At the moment, the WalletConnect modal is checking first if we're running a NodeJS app. The problem is electron apps have both (Node & Browser), but it's showing the QR code as text.

**Describe the solution you'd like**
Change this [code](https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/legacy/modal/src/index.ts#L6-L9) to check first if we have a browser available.

**Aditional Context**
I'm working in a fix for this

marco-souza:
WalletConnect modal is checking first if we're running a NodeJS app. The problem is electron apps have Node & Browser available. The current implementation it's showing the QR code as text, even though we could just use the browser version.

This PR fixes this issue.

It closes #1007

### DIFF
--- a/legacy/modal/src/index.ts
+++ b/legacy/modal/src/index.ts
@@ -6,7 +6,8 @@ import * as browserLib from "./browser";
 const isNode = () =>
   typeof process !== "undefined" &&
   typeof process.versions !== "undefined" &&
-  typeof process.versions.node !== "undefined";
+  typeof process.versions.node !== "undefined" &&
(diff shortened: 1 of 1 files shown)


marco-souza:
Hey @pedrouid sorry to bother you, but could you take a look when you have some time?

darko7 (approved):

PR #1026, opened 2022-05-09
Removed socket response for empty message

gigajuwels:
This PR removes the socket response when a blank message is received. This is more performant and since there is no outbound data for bad messages and ping packets can also be handled somewhat gracefully.

### DIFF
--- a/servers/relay/src/ws.ts
+++ b/servers/relay/src/ws.ts
@@ -96,7 +96,8 @@ export class WebSocketService {
       this.logger.trace({ type: "message", direction: "incoming", message });
 
       if (!message || !message.trim()) {
-        this.send(socketId, "Missing or invalid socket data");
+        //this.send(socketId, "Missing or invalid socket data");
+        // this gets triggered by web sockets that send ping packets
         return;
       }
       const payload = safeJsonParse(message);


sbc64:
Hello Julia,

I was not capable of reproducing the comment that says:
`        // this gets triggered by web sockets that send ping packets`
As you can see with the following:

```
➜ npx wscat -P --slash --connect ws://localhost:8080
Connected (press CTRL+C to quit)
> /ping
< Received pong
< Received ping
> /ping
< Received pong
< Received ping
>
< Missing or invalid socket data
> a
< Socket message is invalid
< Received ping
< Received ping
< Received ping
>
```

I like the idea of removing a message response when the message is empty

gigajuwels:
This specific issue was found in C#'s keepalive implementation, it sends a blank message instead of a ping which seems odd. 

I'll send an example sometime today

gigajuwels:
image
I was actually able to reproduce on the production bridge server. 
Can you confirm this is the right place for this change?

sbc64:
I managed to reproduce this against the deployed infrastructure. Again, thanks for looking into this. The `Missing or invalid socket data` needs to be fixed in another location related with the new infrastructure: https://medium.com/walletconnect/walletconnect-v1-0-infrastructure-improvements-620c3b805026

We will fix this soon and after it is fixed I will close this PR. Although next time this should be an issue.

arein:
@gigajuwels is this still an issue?

bkrem:
Closing as stale. `servers/relay` has been removed in the meantime in favour of our newer Rust relay (OSS TBD)

PR #1044, opened 2022-05-16
Update signer to account for new namespace changes

xzilja:


packages/utils/src/validators.ts
```diff
@@ -1,26 +1,50 @@
...
+  let compatible = true;
+
+  if (!hasOverlap(sessionKeys, paramsKeys)) return false;
+
+  sessionKeys.forEach(key => {
+    const { accounts, methods, events, extension } = session.namespaces[key];
+    const chains = getAccountsChains(accounts);
+    const requiredNamespace = requiredNamespaces[key];
+
+    if (
```
bkrem, line 24: Just to verify expected boolean logic here because you two are deeper in the spec for these changes than me atm:

This will invalidate `compatible` only if the chains, methods and events _all_ don't have overlap (due to AND `&&`). Is that wanted? 

If a single entity not having overlap is enough to invalidate `compatible` we need to use boolean OR `||` to chain conditions here.

xzilja: All of them should have overlap so `&&` here used here needs to be like this. Basically requiredNamespaces should have everything to satisfy session namespaces or more.

bkrem: Right but this is a negation of `hasOverlap`, so you need to use `||` if any missing overlap should mean compatible is false. Otherwise it means compatible is only false if they all don't have overlap.

Compare with this check further down which seems to contradict this one:

```ts
        const overlap = requiredNamespace.extension?.find(
          ext =>
            hasOverlap(chains, ext.chains) &&
            hasOverlap(methods, ext.methods) &&
            hasOverlap(events, ext.events),
        );
        if (!overlap) compatible = false;
```

xzilja: 🤦 you are right, changing

bkrem: Yeah negation on multiple conditions is a bi***, I had to parse this through in my head at least 5 times 😂  

This is the equivalent using `&&` but I think it's bad syntax-wise (easy to miss the `!` negation at the beginning):

```ts
!(
  hasOverlap(chains, requiredNamespace.chains) && 
  hasOverlap(methods, requiredNamespace.methods) && 
  hasOverlap(events, requiredNamespace.events)
)
```

bkrem (approved):

pedrouid (approved):
👍

PR #1052, opened 2022-05-19
[v2] Pass full jsonrpc payloads on events, add separate session and pairing expiry events

xzilja:
* Passing full jsonrpc payloads in events (i.e. including id / jsonrpc etc...) this was how it worked before, ids are needed for responses as well
* Added 2 new events for session expiry to differentiate them from delete requests

### DIFF
--- a/packages/client/src/constants/client.ts
+++ b/packages/client/src/constants/client.ts
@@ -22,7 +22,9 @@ export const CLIENT_EVENTS: Record<ClientTypes.Event, ClientTypes.Event> = {
   session_ping: "session_ping",
   pairing_ping: "pairing_ping",
   session_delete: "session_delete",
+  session_expired: "session_expired",
   pairing_delete: "pairing_delete",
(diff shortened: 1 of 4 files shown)


packages/client/src/constants/client.ts
```diff
@@ -22,7 +22,9 @@ export const CLIENT_EVENTS: Record<ClientTypes.Event, ClientTypes.Event> = {
   session_ping: "session_ping",
   pairing_ping: "pairing_ping",
   session_delete: "session_delete",
+  session_expired: "session_expired",
   pairing_delete: "pairing_delete",
+  pairing_expired: "pairing_expired",
```
pedrouid: For consistency here it should be renamed to “expire”

xzilja: Updated 👍

bkrem (approved):

pedrouid (approved):

PR #1057, opened 2022-05-20
[v2] Event argument type changes

xzilja:
* Introduces more consistent way of exposing event arguments

### DIFF
--- a/packages/client/src/controllers/engine.ts
+++ b/packages/client/src/controllers/engine.ts
@@ -459,7 +459,7 @@ export class Engine extends IEngine {
       const { params, id } = payload;
       const proposal = { id, pairingTopic: topic, ...params };
       await this.client.proposal.set(id, proposal);
-      this.client.events.emit("session_proposal", { ...payload, params: proposal });
+      this.client.events.emit("session_proposal", { id, params: proposal });
     } catch (err) {
       this.client.logger.error(err);
     }
@@ -562,10 +562,7 @@ export class Engine extends IEngine {
       const { params, id } = payload;
       await this.client.session.update(topic, { namespaces: params.namespaces });
       await this.sendResult<"wc_sessionUpdate">(id, topic, true);
-      this.client.events.emit("session_update", {
-        ...payload,
-        params: { ...payload.params, topic },
-      });
+      this.client.events.emit("session_update", { id, topic, params });
     } catch (err) {
       this.client.logger.error(err);
     }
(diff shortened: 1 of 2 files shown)


packages/client/src/controllers/engine.ts
```diff
@@ -702,8 +699,9 @@ export class Engine extends IEngine {
 
   private onSessionRequest: EnginePrivate["onSessionRequest"] = (topic, payload) => {
     try {
-      this.isValidRequest({ topic, ...payload.params });
-      this.client.events.emit("request", { ...payload, params: { ...payload.params, topic } });
+      const { id, params } = payload;
+      this.isValidRequest({ topic, ...params });
+      this.client.events.emit("request", { id, topic, params });
```
bkrem, line 704: This is also so much nicer in terms of syntax with the explicit object keys 👌 

The `...payload` spreads made it hard to know exactly what the final object looks like without digging into type defs

packages/types/src/client.ts
```diff
@@ -23,30 +22,28 @@ export declare namespace ClientTypes {
...
-    session_proposal: JsonRpcRequest<ProposalTypes.Struct>;
-    session_update: JsonRpcRequest<{ topic: string; namespaces: SessionTypes.Namespaces }>;
-    session_extend: JsonRpcRequest<{ topic: string }>;
-    session_ping: JsonRpcRequest<{ topic: string }>;
-    pairing_ping: JsonRpcRequest<{ topic: string }>;
-    session_delete: JsonRpcRequest<{ topic: string }>;
-    pairing_delete: JsonRpcRequest<{ topic: string }>;
+    session_proposal: Omit<BaseEventArgs<ProposalTypes.Struct>, "topic">;
+    session_update: BaseEventArgs<{ namespaces: SessionTypes.Namespaces }>;
+    session_extend: Omit<BaseEventArgs<unknown>, "params">;
```
bkrem, line 34: Shouldn't we be using `<any>` rather than `<unknown>` here if we want `T` to be truly generic?

bkrem (approved):

packages/types/src/client.ts
xzilja: I personally try to avoid any whenever possible. In this case is doesn't matter as params are omitted. I think this is better than doing `params?: T` as with that one, whenever props are actually present users will have to add `if (params)`.

bkrem: Ah yeah in this case it's the omitted key anyways so doesn't matter 👍 

I agree with avoiding `any` generally, except with generics where you just want a truly transparent type/let the end user assign (literally) anything they want. I suspect with `unknown` the TS compiler would ask you to cast from `unknown`, but I'd have to test it to actually be sure.

PR #1064, opened 2022-05-23
Add padding to base64 encoding

pedrouid:

PR #1065, opened 2022-05-23
Ilja/beta.54

### LINKED ISSUE #1063: SDK beta.54
List of items to resolve for beta.54

- [x] Prefix `session_request` and `session_events` methods
- [x] Add "cleanup" method to engine during init to delete expired sessions and pairings before setting up event listeners
- [x] Ensure awaited `request` returns full jsonrpc response
- [x] Send error responses upon invalid requests (inside on...Request methods)
- [x] Add `acknowledged` promise to delete, update and extend methods
- [x] Add extra validation to engine methods that checks if topic is expired or not (throw and delete if it is)

xzilja:
closes #1063 (also see this for list of changes)
Also implements https://github.com/WalletConnect/walletconnect-monorepo/pull/1064

### DIFF
--- a/packages/client/src/controllers/engine.ts
+++ b/packages/client/src/controllers/engine.ts
@@ -648,8 +648,8 @@ export class Engine extends IEngine {
     try {
       this.isValidDisconnect({ topic, reason: payload.params });
       const { id } = payload;
-      await this.sendResult<"wc_sessionDelete">(id, topic, true);
       await this.deleteSession(topic);
+      await this.sendResult<"wc_sessionDelete">(id, topic, true);
       this.client.events.emit("session_delete", { id, topic });
     } catch (err) {
       this.client.logger.error(err);
@@ -676,8 +676,8 @@ export class Engine extends IEngine {
     try {
       this.isValidDisconnect({ topic, reason: payload.params });
       const { id } = payload;
-      await this.sendResult<"wc_pairingDelete">(id, topic, true);
       await this.deletePairing(topic);
+      await this.sendResult<"wc_pairingDelete">(id, topic, true);
       this.client.events.emit("pairing_delete", { id, topic });
     } catch (err) {
       this.client.logger.error(err);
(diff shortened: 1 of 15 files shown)


packages/client/src/controllers/engine.ts
```diff
@@ -187,45 +202,51 @@ export class Engine extends IEngine {
...
 
   public request: IEngine["request"] = async params => {
-    this.isValidRequest(params);
+    this.isInitialized();
+    await this.isValidRequest(params);
     const { chainId, request, topic } = params;
     const id = await this.sendRequest(topic, "wc_sessionRequest", { request, chainId });
     const { done, resolve, reject } = createDelayedPromise<JsonRpcResponse>();
-    this.events.once<"request">(engineEvent("request", id), ({ error, data }) => {
+    this.events.once<"session_request">(engineEvent("session_request", id), ({ error, data }) => {
```
bkrem, line 240: I saw this item checked in the issue:
`Ensure awaited request returns full jsonrpc response`

Can you point me to where the actual response changes (rather than simply the event enum change)? Having a hard time finding the relevant diff

packages/client/test/client.spec.ts
```diff
@@ -80,7 +78,7 @@ describe("Client Integration", () => {
       const clients = await initTwoClients();
       const fakeTopic = "nonsense";
       await expect(clients.A.ping({ topic: fakeTopic })).to.eventually.be.rejectedWith(
-        `No matching pairing or session with topic: ${fakeTopic}`,
+        "Missing or invalid topi",
```
pedrouid: Typo?

packages/client/src/controllers/engine.ts
xzilja: Yep, its here
https://github.com/WalletConnect/walletconnect-monorepo/pull/1065/files#diff-dd17e144eb99c656245e2a405b314188f3d942c7bb843f06bea9b67d66c6358cL716-R775

Will return full payload as opposed to just result

packages/client/test/client.spec.ts
xzilja: Yep, fixed 🆗 at some point will prob use `ERROR.SOMETHING.stringify()` for tests as well so its all consistent

packages/client/src/controllers/engine.ts
bkrem: Gotcha thanks, missed that specific change to `payload` on my first pass 👍

PR #1069, opened 2022-05-23
Add type to result / simplify it

xzilja:
- Add `request<T>` to type response data
- Only return result not error

### DIFF
--- a/packages/client/src/controllers/engine.ts
+++ b/packages/client/src/controllers/engine.ts
@@ -13,6 +13,7 @@ import {
 import { FIVE_MINUTES } from "@walletconnect/time";
 import {
   IEngine,
+  EngineTypes,
   IEngineEvents,
   RelayerTypes,
   EnginePrivate,
@@ -46,7 +47,6 @@ import {
   isExpired,
   isUndefined,
 } from "@walletconnect/utils";
-import { JsonRpcResult } from "@walletconnect/jsonrpc-types";
 
 export class Engine extends IEngine {
   private events: IEngineEvents = new EventEmmiter();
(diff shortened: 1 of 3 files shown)


packages/types/src/engine.ts
```diff
@@ -34,7 +34,7 @@ export declare namespace EngineTypes {
     pairing_ping: { error?: ErrorResponse };
     session_delete: { error?: ErrorResponse };
     pairing_delete: { error?: ErrorResponse };
-    session_request: { error?: ErrorResponse; response?: JsonRpcResponse };
+    session_request: { error?: ErrorResponse; result?: JsonRpcResult };
```
pedrouid, line 37: This should be any

pedrouid:
```suggestion
    session_request: { error?: ErrorResponse; result?: any };
```

packages/client/src/controllers/engine.ts
```diff
@@ -778,7 +775,7 @@ export class Engine extends IEngine {
   ) => {
     const { id } = payload;
     if (isJsonRpcResult(payload)) {
-      this.events.emit(engineEvent("session_request", id), { response: payload });
+      this.events.emit(engineEvent("session_request", id), { result: payload });
```
pedrouid:
```suggestion
      this.events.emit(engineEvent("session_request", id), { result: payload.result });
```

packages/types/src/engine.ts
```diff
@@ -308,7 +308,7 @@ export abstract class IEngine {
 
   public abstract extend(params: EngineTypes.ExtendParams): EngineTypes.AcknowledgedPromise;
 
-  public abstract request(params: EngineTypes.RequestParams): Promise<JsonRpcResponse>;
+  public abstract request<T>(params: EngineTypes.RequestParams): Promise<JsonRpcResult<T>>;
```
pedrouid:
```suggestion
  public abstract request<T>(params: EngineTypes.RequestParams): Promise<T>;
```

pedrouid (changes requested):

PR #1074, opened 2022-05-24
Rename @walletconnect/client to @walletconnect/sign-client

pedrouid:
Rename all references to Client to SignClient as either a Class, variable prefix or constant


xzilja (approved):
Massive work 👏 One question, will `@walletconnect/types` (and similar) change to `@walletconnect/sign-types`? Or not at this stage?

bkrem (approved):

pedrouid:
> Massive work 👏 One question, will `@walletconnect/types` (and similar) change to `@walletconnect/sign-types`? Or not at this stage?

Not at this stage but definitely could use better type management

PR #1080, opened 2022-05-25
proposals expiry

pedrouid:


xzilja (approved):
looks goof 👍 the only thing I'd change (if possible) is allow to pass both id or number to expirer (as key or as part of params) and handle all `formatTopicTarget` / `formatIdTarget` logic in there. Mostly to reduce complexity in the engine.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -760,13 +783,19 @@ export class Engine extends IEngine {
...
     this.client.expirer.on(EXPIRER_EVENTS.expired, async (event: ExpirerTypes.Expiration) => {
-      const { topic } = event;
-      if (this.client.session.keys.includes(topic)) {
-        await this.deleteSession(topic);
-        this.client.events.emit("session_expire", { topic });
-      } else if (this.client.pairing.keys.includes(topic)) {
-        await this.deletePairing(topic);
-        this.client.events.emit("pairing_expire", { topic });
+      const [type, value] = event.target.split(":");
+      if (type === "topic" && typeof value === "string") {
```
bkrem, line 787: Not a huge fan of having two levels of separate `if/else` branching here, but we can address this at a later point to expedite this for now.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -760,13 +783,19 @@ export class Engine extends IEngine {
...
+        const topic = value;
+        if (this.client.session.keys.includes(topic)) {
+          await this.deleteSession(topic);
+          this.client.events.emit("session_expire", { topic });
+        } else if (this.client.pairing.keys.includes(topic)) {
+          await this.deletePairing(topic);
+          this.client.events.emit("pairing_expire", { topic });
+        }
+      } else if (type === "id" && typeof value === "number") {
```
bkrem, line 796: Do we need to check the fall through case here (i.e. if it matched neither `if` nor `else if` for some reason)?

bkrem (approved):

PR #1114, opened 2022-05-31
v2 Pass Storage Provider

0xBC84:
In a previous beta version we were able to pass our own storage provider, as long as it followed the `IKeyValueStorage` interface. This is desirable for when we want to use something other than SQLite, eg: a simple in memory DB or file system storage. This PR reimplements that ability.

### DIFF
--- a/packages/core/src/core.ts
+++ b/packages/core/src/core.ts
@@ -56,7 +56,9 @@ export class Core extends ICore {
     this.logger = generateChildLogger(logger, this.name);
     this.heartbeat = new HeartBeat();
     this.crypto = new Crypto(this, this.logger, opts?.keychain);
-    this.storage = new KeyValueStorage({ ...CORE_STORAGE_OPTIONS, ...opts?.storageOptions });
+    this.storage = opts?.storage ?
+      opts.storage :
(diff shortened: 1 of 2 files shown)


bkrem:
Hi @0xBC84,

You're right this was initially removed in the latest form of the v2.x SDK, since it is now no longer _required_ to explicitly pass the additional storage dependency (e.g. SQLite in a Node.js environment) to the `sign-client`, but as you already pointed out it may well be _desirable_ for consumers of `sign-client` to provide their own.

Thank you for this PR, we'll most likely move ahead with this once the team has had a chance to discuss the change together first thing tomorrow 👍 

In the meantime, could you give us some context on your use case and which alternative storage dependency/lib you'd be using here? It would help us a lot in terms of understanding additional use cases and possible edge cases 🙏

0xBC84:
Previously I was using this method to implement a file system based storage provider, I have since realised that `keyvaluestorage` actually uses `better-sqlite`, which doesn't require an SQLite client, and I was able to specify my own `data.db` table to use the FS, instead of using `:memory`. I'm building out a CLI, so FS based storage is required for me to persist data over a series of commands.

So I don't actually need to provide my own storage provider anymore, as the built in does what I need. However, this could be useful for leveraging a Redis instance or local storage if the client is in the browser.

bkrem (approved):

PR #1120, opened 2022-06-01
[v2] Add .getAll with optional filters to stores

### LINKED ISSUE #1102: Improve State Getters
State getters currently look like this

client.session.values
client.pairing.values

and these get all the values for these respective stores but we want to filter them by active and acknowledged, respectively, so instead we have

getAllSessions()
getAllPairings()

additionally we would also have one for single getters

getSession(topic: string)
getPairing(topic: string)

might as well add another one for proposals to avoid inconsistencies

getAllProposals()
and
getProposal(id: number)

xzilja:
closes #1102

However I implemented it in a slightly different way. I created a `getAll` method on store that takes in optional filter object in form of partial store data struct i.e. you can do things like `getAll({ active: true })` to only return active sessions for example.

I think this way will enable all use-cases for #1102 without us needing to change too many things around. And it is more flexible for extra filtering if we need it in the future.

### DIFF
--- a/packages/core/src/controllers/store.ts
+++ b/packages/core/src/controllers/store.ts
@@ -3,6 +3,7 @@ import { ICore, IStore, PairingTypes, ProposalTypes, SessionTypes } from "@walle
 import { ERROR, isProposalStruct, isSessionStruct } from "@walletconnect/utils";
 import { Logger } from "pino";
 import { CORE_STORAGE_PREFIX, STORE_STORAGE_VERSION } from "../constants";
+import isEqual from "lodash.isequal";
 
 type StoreStruct = SessionTypes.Struct | PairingTypes.Struct | ProposalTypes.Struct;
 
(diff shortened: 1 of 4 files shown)


packages/core/src/controllers/store.ts
```diff
@@ -87,6 +87,14 @@ export class Store<Key, Data extends StoreStruct> extends IStore<Key, Data> {
     return value;
   };
 
+  public getAll: IStore<Key, Data>["getAll"] = filter => {
+    if (!filter) return this.values;
+
+    return this.values.filter(value =>
+      Object.keys(filter).every(key => value[key] === filter[key]),
```
bkrem: Not a big concern right now, but this will only work with JS primitive types, not if users want to use object/array structures to filter by (e.g.  filtering pairings by `peerMetadata`)

bkrem (approved):
Nice, I like the more agnostic `getAll()` approach 👍

packages/core/src/controllers/store.ts
bkrem: Might be cleaner/less edge-casey to just compare the passed `filter` object as a whole with `lodash.deepEqual`, which then also covers cases like `.getAll({ active: true, peerMetadata: {name: "..."} })`

xzilja: Good point, I added `isEqual`, I think thats the one you meant?

bkrem: Yeah exactly `isEqual` my bad, sweet 👍

PR #1125, opened 2022-06-02
[v2] ensure update() adheres to requiredNamespaces

### LINKED ISSUE #1121: Handle requiredNamespaces in update method

xzilja:
closes #1121

### DIFF
--- a/packages/sign-client/test/validation.spec.ts
+++ b/packages/sign-client/test/validation.spec.ts
@@ -372,7 +372,7 @@ describe("Sign Client Validation", () => {
       );
     });
 
-    it("thorws when incompatible namespaces params are provided", async () => {
+    it("throws when incompatible namespaces params are provided", async () => {
       await expect(
         client.update({
           topic,
--- a/packages/sign-client/test/validation.spec.ts
+++ b/packages/sign-client/test/validation.spec.ts
@@ -13,6 +13,8 @@ import {
   TEST_RESPOND_PARAMS,
   TEST_NAMESPACES,
   TEST_REQUIRED_NAMESPACES,
+  TEST_NAMESPACES_INVALID_METHODS,
+  TEST_NAMESPACES_INVALID_CHAIN,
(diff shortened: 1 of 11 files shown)


packages/sign-client/test/validation.spec.ts
```diff
@@ -369,6 +371,21 @@ describe("Sign Client Validation", () => {
         "Missing or invalid update namespaces",
       );
     });
+
+    it("thorws when incompatible namespaces params are provided", async () => {
```
bkrem: Typo:
```suggestion
    it("throws when incompatible namespaces params are provided", async () => {
```

packages/utils/src/validators.ts
```diff
@@ -286,3 +286,10 @@ export function isValidNamespacesEvent(
   const events = getNamespacesEventsForChainId(namespaces, chainId);
   return events.includes(eventName);
 }
+
+export function isValidNamespacesChange(
+  requiredNamespaces: ProposalTypes.RequiredNamespaces,
+  namespaces: SessionTypes.Namespaces,
+) {
+  return !Object.keys(getDeletedDiff(requiredNamespaces, namespaces)).length;
```
bkrem: I think the simpler semantics would be to avoid the negation and explicitly do `Object.keys(...).length === 0`

bkrem (approved):
Nice! 💯 

Took my time to go over `getDeletedDiff` a few times because of `.reduce` + recursion but looks good to me 👍

xzilja:
> Took my time to go over getDeletedDiff a few times because of .reduce + recursion but looks good to me 👍

@bkrem yeh, implementation is adapted from https://www.npmjs.com/package/deep-object-diff but since we only needed that small bit, I though its best to make a small util instead of using lib

bkrem:
Ahh ok thanks for the link to that 👍 Yeah agreed, if we just need a small subset or an adaption of it pulling in the whole dep is overkill.

PR #1137, opened 2022-06-06
[v2] Move expirer target logic outside of engine

### LINKED ISSUE #1103: Improve expiry controller
- Logic could be extracted inside the expiry
- Make expiry controller smarter
- Better separation of concerns

xzilja:
closes #1103

### DIFF
--- a/packages/utils/src/misc.ts
+++ b/packages/utils/src/misc.ts
@@ -241,8 +241,10 @@ export function parseExpirerTarget(target: string) {
   const parsed: { id?: number; topic?: string } = { id: undefined, topic: undefined };
   if (type === "topic" && typeof value === "string") {
     parsed.topic = value;
-  } else if (type === "id" && typeof value === "number") {
-    parsed.id = value;
+  } else if (type === "id" && Number.isInteger(Number(value))) {
+    parsed.id = Number(value);
+  } else {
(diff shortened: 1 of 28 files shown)


packages/utils/src/misc.ts
```diff
@@ -236,6 +236,18 @@ export function formatIdTarget(id: number): string {
...
 
+export function parseExpirerTarget(target: string) {
+  const [type, value] = target.split(":");
+  const parsed: { id?: number; topic?: string } = { id: undefined, topic: undefined };
+  if (type === "topic" && typeof value === "string") {
+    parsed.topic = value;
+  } else if (type === "id" && typeof value === "number") {
+    parsed.id = value;
+  }
+
```
bkrem, line 249: Do we need to handle the unknown case here too? 

E.g. throw if `type` is a garbage value or `id` is a string and doesn't match the second statement, so the function calling `parseExpirerTarget` doesn't just proceed with an empty object.

bkrem (approved):

packages/utils/src/misc.ts
xzilja: Good idea, in fact it made me catch another bug with `typeof value === "number"` it will never be true because id here will always be string from `const [type, value] = target.split(":");` so needs to be converted to Number

PR #1147, opened 2022-06-10
fix falsy result values for respond

### LINKED ISSUE #1148: response validation
**Describe the bug**

This issue was reported in beta 100 by Neon Wallet:

I think I found a bug. Nothing happens if I send this response:

```
const response = {id: 1654795748359334, jsonrpc: '2.0', result: false}
await signClient.respond({topic, response})
```

but if I send anything different than `false` it works maybe you are checking if the result is not null like this:
`if (result) {`

**SDK Version (if relevant)**
 - Client: JS
 - Version v2-beta.100

pedrouid:


bkrem:
Relates to: #1148

PR #1151, opened 2022-06-11
Handle JSON-RPC Error data

### LINKED ISSUE #1000: Wallet Connect not exposing error.data when custom error is thrown
**Describe the bug**

Wallet Connect does not seem to return the original error from json rpc endpoint when the smart contract throws custom error which was introduced in Solidiy 0.8.4 https://blog.soliditylang.org/2021/04/21/custom-errors/ 
This leads to ethersjs unable to decode custom error [which ccip-read relies on showing ENS offchain record](https://medium.com/the-ethereum-name-service/upgrade-ethers-js-to-5-6-1-to-activate-ens-l2-offchain-integration-40ee1a0fdf2a)

**SDK Version (if relevant)**
 - Version 1.7.8

**To Reproduce**
1. Go to https://offchainexample2.surge.sh/name/1.offchainexample.eth/details
2. ENS has various record such as ETH/LTC/email/description
3. Click "Connect" on the left side
4. Click "Wallet Connect".
5. Then all the record disappears.
6. Click "Disconnect"
7. Then all the record reappear.

**Expected behavior**

When I put the following console.log at [`checkError` function of ethers js] (https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/json-rpc-provider.ts#L52), The provider via Infura (which we connect during read only mode) returned `error.data` on the left side of screenshot whereas wallet connect doesn't have any data on the right side of the screenshot below.

```
function checkError(method: string, error: any, params: any): any {
    console.log('***checkError', {
        method, error, params
    })
```

**Screenshots**

Screenshot_2022-04-26_at_20 15 45

pedrouid:
Fix #1000

### DIFF
--- a/packages/helpers/qrcode-modal/package.json
+++ b/packages/helpers/qrcode-modal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@walletconnect/qrcode-modal",
-  "version": "1.7.8",
+  "version": "1.8.0-rc.0",
   "description": "QR Code Modal for WalletConnect",
   "scripts": {
     "clean": "rm -rf dist",
@@ -68,9 +68,9 @@
     "webpack-cli": "3.3.10"
   },
   "dependencies": {
-    "@walletconnect/browser-utils": "^1.7.8",
+    "@walletconnect/browser-utils": "^1.8.0-rc.0",
     "@walletconnect/mobile-registry": "^1.4.0",
-    "@walletconnect/types": "^1.7.8",
+    "@walletconnect/types": "^1.8.0-rc.0",
     "copy-to-clipboard": "^3.3.1",
     "preact": "10.4.1",
     "qrcode": "1.4.4"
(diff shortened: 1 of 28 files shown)


pedrouid:
Published a release candidate under `next` tag with version `1.8.0-rc.0`

xzilja (approved):
Looks good. If you scroll further down in the pr, github picked up a bunch of unused vars. Not sure if they are needed, but if not they can prob be cleaned up.

packages/helpers/qrcode-modal/package.json
```diff
@@ -9,7 +9,7 @@
     "build:pre": "run-s clean build:svg build:css",
     "build:umd": "webpack",
     "build:cjs": "microbundle --no-compress -f cjs --alias react=preact/compat,react-dom/test-utils=preact/test-utils,react-dom=preact/compat --tsconfig ./tsconfig.cjs.json",
-    "build": "run-s build:pre build:cjs build:umd",
+    "build": "run-s build:pre build:cjs",
```
bkrem: We removed this to unblock the builds for the other modules, since the UMD build here was having cryptic TS issues.

@pedrouid we need to add this back before merging to avoid quietly discontinuing UMD builds:

```suggestion
    "build": "run-s build:pre build:cjs build:umd",
```

bkrem (changes requested):

bkrem (approved):

PR #1160, opened 2022-06-14
[beta.102] ethereum-provider package

### LINKED ISSUE #1117: Adjust Providers to work with 2.0-rc client (Ethereum, Cosmos, Solana)
- [x] Fix build
- [x] Fix types
- [x] Fix unit tests
- [ ] Update examples

xzilja:
closes [#1117](https://github.com/WalletConnect/walletconnect-monorepo/issues/1117)


bkrem:
> closes https://github.com/WalletConnect/walletconnect-monorepo/issues/1117

Quick project management note before I jump into code here: should we use "part of x" rather than "closes x" here since the issue specifies all providers, but this tackles specifically ethereum-provider as far as I can tell?

Or we can split the rest out into a new issue.

xzilja:
@bkrem I think I'll scope up an entirely new suite of issues for providers after work on ws auth is done. I want to re-work providers entirely, since scope is not that big and we can get few improvements there in terms of management and making it easier for other chains i.e. solana / cosmos / polkadot to integrate.

providers/ethereum-provider/src/index.ts
```diff
@@ -64,27 +65,21 @@ export interface EthereumProviderOptions {
 
 class EthereumProvider implements IEthereumProvider {
   public events: any = new EventEmitter();
-
-  private rpc: EthereumRpcConfig;
-
+  public rpc: EthereumRpcConfig;
```
bkrem: Is there a reason why `this.rpc` was rescoped from `private` to `public`? Couldn't find an example of it being being called from outside the class anywhere else in the diff

bkrem (approved):

bkrem:
> I think I'll scope up an entirely new suite of issues for providers after work on ws auth is done.

Gotcha, thanks for clarifying 👍

providers/ethereum-provider/src/index.ts
xzilja: I think we did this with pedro during quick fix sesh, but I just changed it back to private, ran tests and checked ts, all good. Nice catch 👌

providers/ethereum-provider/src/index.ts
```diff
@@ -229,11 +223,19 @@ class EthereumProvider implements IEthereumProvider {
     return Number(chainId.split(":")[1]);
   }
 
-  private setChainId(chains: string[]) {
```
pedrouid, line 232: I'm confused about the name change here

pedrouid: My main question is if this provider is still single-chain or is now multi-chain

xzilja: There are 2 methods `setChainId` and `setChainIds` now. One takes in array, the other single chain (string) and both perform different logic. 

`ethereum-provider` is not single chain, but it it is single namespace `eip155`. I mentioned in discord, that I want to come back to providers after we release beta 102 and re-work it based on what we discussed there as well. But it will require a lot of time, so for now I stopped at just fixing ethereum-provider

Issue #936, opened 2022-04-11
Update utils package to account for new changes


chadyj:
- [x] Update uri utils
- [x] Update validation utils

xzilja:
Done

PR #1176, opened 2022-06-21
[v2] Better error handling and feedback

### LINKED ISSUE #1139: Improved error handling
Received this feedback:

> regarding the @walletconnect/sign-client I can notice that the "5 mins timout pairing/session" error from (previously client.connect() now approval()) which was a string starting with "Pairing failed to settle after" (beta 26) changed to an undefined Error (beta 100). 
Does errors will raise a custom type/msg in later updates so that we can handle errors case differently if needed ?

We will need to delete/add some errors.

xzilja:
closes [#1139](https://github.com/WalletConnect/walletconnect-monorepo/issues/1139) and [#936](https://github.com/WalletConnect/walletconnect-monorepo/issues/936)

- More verbose / actionable error messages
- Added namespace conformity checks against requiredNamespaces
- Validating proposal id's
- New errors util, to simplify things and adhere to new error codes https://github.com/WalletConnect/walletconnect-specs/blob/main/sign/codes.md (had to add some internal ones in as well for better clarity)
- Re-worked few areas of core and engine as I was able to find few faulty cases with new validators
- Cleaned up few validators where previously `optional` argument was used, but is no longer necessary due to new rules around requiredNamespaces


packages/core/src/controllers/subscriber.ts
```diff
@@ -297,14 +296,12 @@ export class Subscriber extends ISubscriber {
...
   private async resubscribe(subscription: SubscriberTypes.Active) {
-    const { topic, relay } = subscription;
-    const params = { topic, relay };
-    this.pending.set(params.topic, params);
-    const id = await this.rpcSubscribe(params.topic, params.relay);
-    this.onResubscribe(id, params);
-    if (this.ids.includes(subscription.id)) {
-      const reason = ERROR.RESUBSCRIBED.format({ topic: subscription.topic });
-      this.deleteSubscription(subscription.id, reason);
+    if (!this.ids.includes(subscription.id)) {
```
bkrem, line 300: Could you outline what changed here, just simplified logic by inverting the if condition?

packages/utils/src/errors.ts
```diff
@@ -0,0 +1,149 @@
...
+    code: 9,
+  },
+};
+
+/**
+ * Utilities
+ */
+export function getInternalError(key: InternalErrorKey, context?: string | number) {
+  const { message, code } = INTERNAL_ERRORS[key];
+  return {
```
bkrem, line 142: Do we still get stack traces etc for these if we're simply throwing the returned object rather than actual instance of `Error`?

packages/core/src/controllers/subscriber.ts
xzilja: Essentially, before it subscribed, then checked if subscription.id is already included in subscription, if it was it then deleted subscription. I changed it so it checks if subscription.id is not already included and only if it is not, subscribes avoiding the need to delete / cleanup afterwards

packages/utils/src/errors.ts
```diff
@@ -0,0 +1,149 @@
...
+ * Types
+ */
+type SdkErrorKey = keyof typeof SDK_ERRORS;
+type InternalErrorKey = keyof typeof INTERNAL_ERRORS;
+
+/**
+ * Constants
+ */
+const SDK_ERRORS = {
+  INVALID_METHOD: {
```
bkrem, line 12: Style (optional):

Could be nice to label these by error class, similar to how they are in the spec, e.g.:

```js
  /* ----- INVALID (1xxx) ----- */
  INVALID_METHOD: {
  // ...
  /* ----- UNAUTHORIZED (3xxx) ----- */
  UNAUTHORIZED_METHOD: {
```

packages/utils/src/errors.ts
```diff
@@ -0,0 +1,149 @@
...
+  },
+  UNAUTHORIZED_UPDATE_REQUEST: {
+    message: "Unauthorized update request.",
+    code: 3003,
+  },
+  UNAUTHORIZED_EXTEND_REQUEST: {
+    message: "Unauthorized extend request.",
+    code: 3004,
+  },
+  USER_REJECED: {
```
bkrem: Typo:

```suggestion
  USER_REJECTED: {
```

packages/utils/src/errors.ts
```diff
@@ -0,0 +1,149 @@
...
+    message: "User rejected.",
+    code: 5000,
+  },
+  USER_REJECED_CHAINS: {
```
bkrem:
```suggestion
  USER_REJECTED_CHAINS: {
```

packages/utils/src/validators.ts
```diff
@@ -6,7 +6,46 @@ import {
...
+import { hasOverlap } from "./misc";
+
+export type ErrorObject = { message: string; code: number } | null;
+
+// -- types validation ----------------------------------------------------- //
+
+export function isValidArray(arr: any, itemCondition?: (item: any) => boolean) {
+  if (Array.isArray(arr)) {
+    if (typeof itemCondition !== "undefined" && arr.length) {
+      const matches = arr.filter(itemCondition);
```
bkrem: Can simplify this by using `Array#every` to check every array item meets the predicate/condition, then no length comparison needed:

```suggestion
      return arr.every(itemCondition);
```

packages/utils/src/validators.ts
```diff
@@ -6,7 +6,46 @@ import {
...
+      return matches.length === arr.length;
+    } else {
+      return true;
+    }
+  }
+  return false;
+}
+
+export function isValidObject(obj: any) {
+  return Object.getPrototypeOf(obj) === Object.prototype && Object.keys(obj).length;
```
bkrem, line 28: So we're not seeing empty object `{}` as a valid object, is that correct?

xzilja: I actually just moved this one while re-organising, but yeh `{}` not valid object in cases where this one is used

packages/utils/src/validators.ts
```diff
@@ -6,7 +6,46 @@ import {
...
+export function isValidString(input: any, optional: boolean) {
+  if (optional && isUndefined(input)) return true;
+
+  return typeof input === "string" && Boolean(input.trim().length);
+}
+
+export function isValidNumber(input: any, optional: boolean) {
+  if (optional && isUndefined(input)) return true;
+
+  return typeof input === "number";
```
bkrem: Another "fun" little JS nugget, we need to guard against `NaN` here because:

```
typeof NaN
-> 'number'
```

```suggestion
  return typeof input === "number" && !isNaN(input);
```

---

(😂 😭 )

bkrem: Lol I just realised this code just moved up in the file 🤦 

Anyways good to double check I guess 🤷

packages/utils/src/validators.ts
```diff
@@ -119,74 +125,220 @@ export function isSessionStruct(input: any): input is SessionTypes.Struct {
...
+  }
+
+  return error;
+}
+
+export function isValidNamespaceMethodsOrEvents(input: any): input is string {
+  let valid = true;
+  if (isValidArray(input)) {
+    if (input.length) {
+      input.forEach((item: any) => {
```
bkrem: Think this can also be simplified with `every`, which is doing basically the same thing under the hood:

```suggestion
      valid = input.every((item: any) => isValidString(item, false))
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -471,7 +476,7 @@ export class Engine extends IEngine {
       case "wc_sessionEvent":
         return this.onSessionEventRequest(topic, payload);
       default:
-        // TODO(ilja) throw / log unsuported event?
+        this.client.logger.info(`Unsuported request method ${reqMethod}`);
```
```suggestion
        this.client.logger.info(`Unsupported request method ${reqMethod}`);
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -497,12 +502,12 @@ export class Engine extends IEngine {
       case "wc_sessionRequest":
         return this.onSessionRequestResponse(topic, payload);
       default:
-        // TODO(ilja) throw / log unsuported event?
+        this.client.logger.info(`Unsuported response method ${resMethod}`);
```
```suggestion
        this.client.logger.info(`Unsupported response method ${resMethod}`);
```

bkrem (commented):
Took me a while but I've gone over everything, great work here this was an intense change 💯 

I want to have another pass at `validators.ts` at a later point since that's the most dense diff in terms of logic. 

Ping me for re-review once the error stack trace stuff is there, and then I'll approve from my side after I checked that and had another pass at `validators.ts` and the validators spec side by side 👍

bkrem (approved):
💪 :shipit:

pedrouid (approved):
👍

PR #1178, opened 2022-06-22
[v2] Client authentication

xzilja:
- Rename `waku` implementation to `iridium`
- Remove relayUrl formatting and fallback handling from core (it's all in one place now in relay controller)
- Add relay authentication to relay.init() method
- Added `node-fetch` package for fallback when fetch is not available
- Remove `relayProvider` option from opts
- Update `@walletconnect/utils` packages
- Update utils tests


ops/js/README.md
```diff
@@ -4,7 +4,7 @@
 2. Install redis and run it as a [brew service](https://gist.github.com/tomysmile/1b8a321e7c58499ef9f9441b2faa0aa8)
 3. Bootstrap lerna (install npm packages) `npm run bootstrap`
 4. Verify installation `npm run check`
-5. Start relay server `PORT=5555 npm run start --prefix=servers/relay`
+5. Start ts-relay server (separate repo) `PORT=5555 npm run start`
```
bkrem:
```suggestion
5. Start ts-relay server ([separate repo](https://github.com/WalletConnect/ts-relay)) `PORT=5555 npm run start`
```

packages/core/src/constants/core.ts
```diff
@@ -7,7 +7,6 @@ export const CORE_STORAGE_PREFIX = `${CORE_PROTOCOL}@${CORE_VERSION}:${CORE_CONT
 export const CORE_DEFAULT = {
   name: CORE_CONTEXT,
   logger: "error",
-  relayUrl: "wss://relay.walletconnect.com",
```
bkrem, line 10: Are we still defaulting to this endpoint somewhere else? Couldn't find with cmd+F in the diffs here, but I've probably just missed it somewhere.

xzilja: We had 2 fallbacks in relay and core, this one is from core. I simplified it a bit with most relay stuff now being in that respective controller

packages/core/src/controllers/relayer.ts
```diff
@@ -62,20 +64,22 @@ export class Relayer extends IRelayer {
...
-      typeof opts.relayProvider !== "string" && typeof opts.relayProvider !== "undefined"
-        ? opts.relayProvider
-        : new JsonRpcProvider(new WsConnection(rpcUrl));
     this.messages = new MessageTracker(this.logger, opts.core);
     this.subscriber = new Subscriber(this, this.logger);
     this.publisher = new Publisher(this, this.logger);
+
+    // re-assigned during init()
+    this.providerOpts = { ...opts };
+    this.provider = {} as IJsonRpcProvider;
```
bkrem, line 73: Is this needed from a class semantics perspective or can we leave `this.provider` as `undefined` until it is assigned during `init`? 

Avoids the empty assignment with type cast here and is technically more accurate for it to be undefined at the constructor stage with this new approach.

xzilja: Mainly to avoid checking for provider being undefined everywhere where it is used (a lot in relay). It becomes "defined" after initialisation and we already have guards for that

pedrouid (approved):
👍

ganchoradkov (approved):
LGTM 👍

bkrem (approved):
Big Diff Energy 👍

tfalencar:
- So this is the PR that removes the self-hosting relay server code. Now that we are almost one year after this, do we have any date estimates for open sourcing the backend code (specially given users were pushed to move to V2)?
- Also now that I see this, seeing waku being renamed to something else looks confusing - does it mean that walletconnect won't follow / contribute back to the waku protocol anymore?

Issue #1153, opened 2022-06-13
Extra v2 tests


chadyj:
- [ ] Most client tests only cover happy paths at the moment, these should be updated to cover as many edge cases as we can come up with. Eg
    -  happy path without validation, eg timeouts and updates
    - Expiries
    - Anything regarding time (expiring a pairing, a session). Mock time.
- [x] Validator tests will need to be updated after or alongside better error handling work

PR #1179, opened 2022-06-22
[v2] feat: sign-client events tests

ganchoradkov:
Tests added for the following events

- session_proposal
- session_update
- session_ping
- session_event
- session_delete

ref https://github.com/WalletConnect/walletconnect-monorepo/issues/1153


packages/sign-client/test/events.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+          
+          clients.A.disconnect({ topic: sessionA.topic, reason: getSdkError('USER_DISCONNECTED')});
+        } catch (e) {
+          reject(e);
+        }
+      })
+      deleteClients(clients);
+    });
+  });
+});
```
xzilja: Add new line here. This should be managed by prettier though, are you using eslint and prettier extensions in your ide?

packages/sign-client/test/events.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+      expect(pairingA.topic).to.eql(uriParams.topic);
+      expect(pairingA.relay).to.eql(uriParams.relay);
+
+      if (!pairingA) throw new Error("expect pairing A to be defined");
+
+      let sessionA: SessionTypes.Struct;
+      let sessionB: SessionTypes.Struct;
+
+      await Promise.all([
+        new Promise<void>((resolve, reject) => {
```
xzilja, line 56: Honestly, you probably don't need to specify return type `<void>` in tests

xzilja (approved):
Niiiice, left few minor suggestions

packages/sign-client/test/events.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+          } catch (error) {
+            reject(error);
+          }
+        })
+      ]);
+
+      deleteClients(clients)
+    });
+
+    it("session_update", async () => {
```
bkrem: Can we turn each event type into a `describe`, i.e. `describe("session_update", ...)` and then have this test inside with a description e.g. `it("emits and handles a valid session_update")`?

That way its more descriptive what the expectations are for this specific test, and lets us add regression tests or expected failure cases for each event type later.

bkrem (approved):

packages/sign-client/test/events.spec.ts
ganchoradkov: It's return `<void>` otherwise `resolve()` expects a value

PR #1185, opened 2022-06-23
add typed envelopes

### LINKED ISSUE #1161: Add envelope types to json rpc message serialization
From SDK meeting:
> Right now the encrypted message has the format of [ (nonce/iv) + encrypted message + tag ]. This is not possible in the Chat SDK method wc_chatInvite and we have to use the format of [publicKey + (nonce/iv) + encrypted message + tag]. To differentiate them easily we should add envelope type to message payload. For example, in Sign we would have [ 0 + (nonce/iv) + encrypted message ], and for Chat [ 1 +publicKey + (nonce/iv) + encrypted message]

pedrouid:


packages/utils/src/crypto.ts
```diff
@@ -70,12 +88,76 @@ export function decrypt(params: CryptoTypes.DecryptParams) {
...
+    }
+  }
+  return {
+    type,
+    senderPublicKey: opts?.senderPublicKey,
+    receiverPublicKey: opts?.receiverPublicKey,
+  };
+}
+
+export function isTypeOneEvelope(
```
bkrem: Typo:

```suggestion
export function isTypeOneEnvelope(
```

packages/core/test/crypto.spec.ts
```diff
@@ -136,98 +132,51 @@ describe("Crypto", () => {
     });
   });
 
-  describe("encrypt", () => {
```
bkrem, line 139: So `encrypt/decrypt` can go completely? @IljaDaderko are you aware of anywhere this may be relied on atm?

packages/core/test/crypto.spec.ts
```diff
@@ -136,98 +132,51 @@ describe("Crypto", () => {
...
-    });
-    it("encodes `payload` as hex string if the passed topic is not known", () => {
-      const topic = utils.generateRandomBytes32();
-      const result = crypto.encode(topic, payload);
-      expect(result).to.equal(encoding.utf8ToHex(safeJsonStringify(payload)));
+      await expect(invalidCrypto.encode("topic", payload)).to.eventually.be.rejectedWith(
+        "Not initialized. crypto",
+      );
     });
     it("encrypts `payload` if the passed topic is known", async () => {
```
bkrem:
```suggestion
    it.skip("encrypts `payload` if the passed topic is known", async () => {
```

Probably better to skip here for now if we don't have any `expect` assertions to avoid a false positive pass.

packages/core/test/crypto.spec.ts
```diff
@@ -136,98 +132,51 @@ describe("Crypto", () => {
...
-      crypto.decrypt = spy;
-      crypto.decode(topic, hexPayload);
-      const [calledTopic, calledEncoded] = spy.getCall(0).args;
-      expect(calledTopic).to.equal(topic);
-      expect(calledEncoded).to.equal(hexPayload);
+      // const spy = Sinon.spy(() => "message");
+      // crypto.decrypt = spy;
+      const decoded = await crypto.decode(topic, encoded);
+      expect(decoded).to.eql(payload);
+      // const [calledTopic, calledEncoded] = spy.getCall(0).args;
```
bkrem: Can we remove the commented out `expect`s and related calls here if not needed anymore?

bkrem (approved):
Looks good 👍 

One typo in `isTypeOneEnvelope` and a couple of small tweaks in the controller spec file

packages/core/test/crypto.spec.ts
xzilja: Noup, crypto is used within the engine that only accesses `encode/decode`

packages/core/test/crypto.spec.ts
pedrouid: I commented out because I wasn't sure if you felt it would be necessary or not

In this case Im going to guess its not longer necessary so I will remove them

packages/core/test/crypto.spec.ts
pedrouid: I double checked and they were only used internally by the Crypto controller

packages/utils/src/crypto.ts
pedrouid: LOL

xzilja (approved):

xzilja:
Good on my end, if its possible to get rid of commented out bits, we should prob try to do that 👍

PR #1191, opened 2022-06-25
fix(): signTypedData_v4 for metamask

midgerate:
Fixes issue where we are unable to sign transactions if its a typed array on metamask connected via walletconnect. 

https://github.com/MetaMask/metamask-mobile/issues/4441

### DIFF
--- a/packages/providers/web3-provider/src/index.ts
+++ b/packages/providers/web3-provider/src/index.ts
@@ -121,7 +121,12 @@ class WalletConnectProvider extends ProviderEngine {
       return;
     }
 
-    return this.sendAsyncPromise(payload.method, payload.params);
+    if(payload.method === "eth_signTypedData_v4" && this.walletMeta?.name === "MetaMask") {
+      const { result } =  await this.handleOtherRequests(payload);
+      return result
+    } else {
+      return this.sendAsyncPromise(payload.method, payload.params);
+    }
   };
 
   onConnect = (callback: any) => {


pedrouid:
Why would this method be an exception?

midgerate:
@pedrouid I could not test much. Here's what I know

1. Metamask defaults eth_signTypedData to eth_signTypedData_v3
2. Other wallets like rainbow, trust, argent defaults to eth_signTypedData_v4
3. When I tried to call eth_signTypedData_v4 with trust wallet, I was unable to sign the request (It seems (not confirmed) that trust does not allow eth_signTypedData_v4 being called directly)
4. I found that in https://github.com/MetaMask/metamask-mobile/issues/4441 using the `handleOtherRequests` with eth_signTypedData_v4 works for metamask and not for trust. 

So I created a specific condition for metamask and we have tested that it works now for metamask and all other wallets. 

It would be great if you could improve this PR since you understand the internals. 

I think this might already be solved in WC2 but it would take us a long time to integrate WC2.

itm-penny:
@midgerate Hi! I am searching for a solution to solve the same problem and I found this pull request, I wonder if I am able to test this locally in my project? 
My project is built in react app in javascript and I am using npm to install the walletconnect package, but I don't know how to change the code locally and to test it. Thanks for the help in advance!

midgerate:
@itm-penny At snapshot we use patch-package to make changes to all the build files and then make it work. You could technically use this fork that I created, publish it to npm and then use it.

itm-penny:
@midgerate Thanks! Will give it a try!

itm-penny:
@midgerate I tried changing the exact same part of the code and used patch-package to test it, and this solved my problem! Thank you so much for the solution, I've stuck in this problem for over a week. Hopefully this pull request gets merged ASAP, thanks again man!

Hmac512:
Thanks for making this PR @midgerate

This issue comes down to a simple question. Why exactly is walletconnect taking an explicitly versioned request like “eth_signTypedData_v4”, and removing the version before sending it to the wallet? “eth_signTypedData”


That seems like a bug. I can’t for the life of me figure out how one could decide to do that on purpose.

samuveth:
@pedrouid could we get this reviewed and merged please. We need this change so that we can move to EIP712 on snapshot.org

xanderdeseyn:
Any progress on this?

bkrem:
Published in https://github.com/WalletConnect/walletconnect-monorepo/releases/tag/1.8.0

PR #1196, opened 2022-06-27
[v2 experiment] new build tooling

xzilja:
## TODOs

- [x] Bundle npm deps in umd build
- [x] Ensure packages inherit deps from root in npm bundle
- [x] Add tool / command for version upgrade management instead of lernas one (ended up using lerna still)
- [x] Update readme's

## CHANGES
- Removed lerna in favour of npm workspaces (still using lerna for versioning and publish)
- Removed Mocha and Chai in favour of vitest
- Removed webpack and babel in favour of rollup / esbuild
- Removed `legacy` package
- Updated dependencies
- Moved shared devDependencies to root package.json (easier to update / manage)
- Fixed few / Disabled few typescript and eslint issues caused by package updates

## Out in the wild
- Example npm release https://www.npmjs.com/package/@walletconnect/sign-client/v/2.0.0-15eadf7
- Example wallet https://react-wallet-v2-git-ilja-200-1eeeacc0-walletconnect1.vercel.app/
- Example dapp https://react-dapp-v2-git-ilja-200-1eeeacc0-walletconnect1.vercel.app/


README.md
```diff
@@ -4,62 +4,33 @@ Open protocol for connecting Wallets to Dapps - https://walletconnect.com
...
-
-# staging server
-make test-staging
-
-# production server
-make test-production
+1. Install [nodejs and npm](https://nodejs.org/en/)
+2. Install python3 and ensure `python` cli is linked (required to build some npm modules)
+3. Install workspace dependencies i.e. run `npm install` from root folder
+4. Install redis. We recomend running it as a [brew service](https://gist.github.com/tomysmile/1b8a321e7c58499ef9f9441b2faa0aa8)
```
bkrem:
```suggestion
4. Install redis. We recommend running it as a [brew service](https://gist.github.com/tomysmile/1b8a321e7c58499ef9f9441b2faa0aa8)
```

package.json
```diff
@@ -17,17 +17,24 @@
...
+    "clean": "npm run clean --workspaces",
+    "lint": "npm run lint --workspaces",
+    "prettier": "npm run prettier --workspaces",
+    "build": "npm run clean --workspaces; npm run build --workspaces",
+    "test": "npm run test --workspaces",
+    "check": "npm run lint; npm run build; npm run test",
+    "reset": "npm run clean; npm run check",
     "new-version": "lerna version --no-git-tag-version",
-    "pre-publish": "run-s new-version reset commit-version",
+    "pre-publish": "npm run new-version; npm run reset",
```
bkrem, line 37: Why was the `commit-version` step removed here?

packages/core/src/constants/relayer.ts
```diff
@@ -32,4 +32,4 @@ export const RELAYER_STORAGE_OPTIONS = {
 
 // FIXME: Always manually change this after each release (hardcoded from package.json)
 // ... this will be temporarily hardcoded until we refactor our build chain!
-export const RELAYER_SDK_VERSION = "2.0.0-beta.102";
+export const RELAYER_SDK_VERSION = "2.0.0-rc.0";
```
bkrem, line 35: Again with the in-source version 😄 I think this needs to be either the canary version you published or we do rc.1 rather than merging a canary version to v2 after we already started RCs

xzilja: Yeh, my thinking here was basically that - update this when we release rc / prod

package.json
xzilja: I looked at this during my cleanup work and I don't think it ever worked i.e. I can't find these commits in git history for last few versions we pushed. Plus when we release we create version tags manually as well as push a separate pr with updated version. Next steps with this would be to actually not run releases manually, rather trigger github action when we "tag" a release here. I'll be looking into this after this pr is merged.

packages/types/src/sign-client/jsonrpc.ts
```diff
@@ -27,7 +27,7 @@ export declare namespace JsonRpcTypes {
       code: number;
       message: string;
     };
-    wc_pairingPing: {};
+    wc_pairingPing: Record<string, unknown>;
```
bkrem, line 30: Nice catch 🏌️‍♂️

bkrem (approved):
Looking great 💯 Most of the diff size is prettier/syntax trivialities so actual change set is a lot more contained than I expected

package.json
bkrem: > I can't find these commits in git history for last few versions we pushed.

Not sure what you mean, I think it was just a convenience script to commit the version after lerna updates everything like here (but I manually tagged it afterwards): https://github.com/WalletConnect/walletconnect-monorepo/commit/419ac1815ac93482b6a19274328af3c0416d5746

I'm not attached to it and agree we should replace with something that also does the tag for us automatically, just wasn't clear on why you thought it wasn't working 🤔

providers/signer-connection/test/index.spec.ts
```diff
@@ -1,3 +1,5 @@
+import { describe, it } from "vitest";
+
 // import "mocha";
 // import * as chai from "chai";
 // import { JsonRpcProvider } from "@walletconnect/jsonrpc-provider";
```
ganchoradkov, line 5: are these lines supposed to be removed or converted to use `vitest`?

ganchoradkov (approved):
Holy molly PR, awesome!  👍

package.json
xzilja: @bkrem yeh so basically that didn't happen to any of the releases I made i.e. lerna did it's thing, but I had to manually commit / pr them in still. And afterwards chad usually manually tagged them. So I got rid of it. My plan on what to do next is basically this

1. We tag version on `v2` branch
2. This triggers ci run that runs lerna with pre-defined version (i.e. tag)
3. This is then deployed to npm
4. Comited back to v2 branch

providers/signer-connection/test/index.spec.ts
xzilja: These are just commented out signer tests, honestly they won't work anymore anyway, but keeping them in just in case for future reference (these were commented out even before this tooling pr)

PR #1201, opened 2022-06-28
handle storage entry being null

StanislavSava:
This call will fail when the storage entry is null (for example, on React Native) because typeof null === "object"


pedrouid:
@StanislavSava why would this be `null`? Could you describe a scenario?

StanislavSava:
If you run a get on a key that has nothing under it, for example, if the setItem hasn't run before the get, on first load, you will get null, not undefined, and the check against === 'undefined" will fail

pedrouid:
Are you using your own custom KeyValueStorage? It should be compatible with this interface

```typescript
export class IKeyValueStorage {
  public getKeys(): Promise<string[]>;
  public getEntries<T = any>(): Promise<[string, T][]>;
  public getItem<T = any>(key: string): Promise<T | undefined>;
  public setItem<T = any>(key: string, value: T): Promise<void>;
  public removeItem(key: string): Promise<void>;
}
```

Source: https://github.com/WalletConnect/walletconnect-utils/tree/master/misc/keyvaluestorage#api

ganchoradkov:
closing as stale

PR #1205, opened 2022-06-29
remove auth-nonce

pedrouid:
Related to WalletConnect/walletconnect-specs#79


xzilja (approved):
👍 LGTM, not sure if we should keep it, but perhaps also remove `getHttpUrl` util?

bkrem (approved):

PR #1210, opened 2022-07-04
Beta.102

pedrouid:
Changes:
- All changes made with canary beta.101.2
- Update JWT payload params (WalletConnect/walletconnect-specs#81)
- Add Sign API tag to iridium_publish


packages/core/src/controllers/publisher.ts
```diff
@@ -78,11 +80,15 @@ export class Publisher extends IPublisher {
         message,
         ttl,
         prompt,
+        tag,
       },
     };
     if (typeof request.params?.prompt === "undefined") {
       delete request.params?.prompt;
     }
+    if (typeof request.params?.tag === "undefined") {
```
xzilja: Could cleanup these checks with `isUndefined` utility from `utils/misc`

xzilja (approved):

PR #1216, opened 2022-07-05
feat: builds sign client into docker

arein:
We currently use the sign client for testing and we consume it as a Docker container. In a previous commit this was broken as the client is no longer available via Docker.

This fixes that regression.

This also improves the previous solution as the Container is now built automatically into the registry by GitHub Actions.

# Testing
Sample CI run: https://github.com/WalletConnect/walletconnect-monorepo/actions/runs/2614023780


xzilja (approved):
LGTM 👍

ganchoradkov (approved):
LGTM

PR #1217, opened 2022-07-05
fix: tests have dependencies outside of the package

arein:
Currently the tests resolve the root directory via a roundtrip to the parent directory. This doesn't work in the docker context and is also unnecessarily complex.

# Testing

`docker run -e TEST_RELAY_URL="wss://dev.relay.walletconnect.com" sign-client npm run test`

### DIFF
--- a/packages/sign-client/test/client.spec.ts
+++ b/packages/sign-client/test/client.spec.ts
@@ -5,11 +5,12 @@ import {
   expect,
   initTwoClients,
   testConnectMethod,
-  TEST_SIGN_CLIENT_DATABASE,
   TEST_SIGN_CLIENT_OPTIONS,
   deleteClients,
 } from "./shared";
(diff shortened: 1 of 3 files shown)


packages/sign-client/test/client.spec.ts
```diff
@@ -149,6 +150,7 @@ describe("Sign Client Integration", () => {
         // delete
         deleteClients(beforeClients);
         // restart
+        console.log("Bang", TEST_SIGN_CLIENT_DATABASE);
```
xzilja: Assume this is no longer needed?

xzilja (approved):
looking good, need to get rid of that console log 👍

bkrem (approved):

packages/sign-client/test/client.spec.ts
arein: ups lol

PR #1220, opened 2022-07-06
feat: adds canary test

arein:
Implements a Canary test to be run in an interval and to depict properly if the infrastructure supporting the sign client is up and running.

Emits success, failure, and latency metrics which we can use to build monitors.

This change is mainly to introduce the backing logic and follow up PRs will improve the actual canary test.


xzilja (approved):
Looking good, could you pop `canary` folder inside `tests` please, should be good to merge afterwards

packages/sign-client/canary/canary.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const environment = process.env.ENVIRONMENT || 'dev';
+
+describe("Canary", function() {
+  describe("HappyPath", function() {
+    // TODO: implement a test that depicts
+    // the happy case better
+    it("connects", async function() {
+      const clients = await initTwoClients();
+      await testConnectMethod(clients);
```
bkrem: This feels like it's leaving the door open on false positive passes since we're not doing any concrete assertions here, i.e. relying on something to throw in these setup functions.

It's not a silver bullet but would be nice to additionally validate the end result with assertions:

```js
expect(clients.A instanceof SignClient).toBe(true);
expect(clients.B instanceof SignClient).toBe(true)
```

packages/sign-client/canary/canary.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+      deleteClients(clients);
+    });
+  });
+  afterEach(function (done) {
+    const cloudwatch = new CloudWatch({region: 'eu-central-1'});
```
bkrem: Can we move this callback out into its own function instead of inlining it, i.e.

```js
const uploadToCloudwatch = function (done) {...}

// ...

afterEach(uploadToCloudwatch)

```

packages/sign-client/canary/canary.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+          MetricName: `${metric_prefix}.latency`,
+          Unit: "Milliseconds",
+          Value: this.currentTest!.duration,
+          Timestamp: ts
+        },
+      ],
+      Namespace: `${environment}_Canary_SignClient`
+    };
+    cloudwatch.putMetricData(params, function(err: Error) {
+      if (err) {console.log(err, err.stack); done();}
```
bkrem: Using `console.error` should give us the stack automatically if it's defined:

```suggestion
      if (err) {console.error(err); done();}
```

bkrem (approved):

PR #1223, opened 2022-07-07
fix(crypto): fixes `generateSharedKey` param ordering in `decode`

bkrem:
## Context

- `crypto.decode` passes `params.senderPublicKey` and `params.receiverPublicKey` to `generateSharedKey` in the same order as `crypto.encode`.
- This is incorrect when decoding a type 1 envelope message, since in this case the `selfPublicKey === receiverPublicKey`.
- Attempting `decode` on a type 1 envelope message currently results in a `NO_MATCHING_KEY` error, since we're trying to look up the private key of the sender, rather than the receiver who is doing the decoding.

## Solution

- Fix parameter ordering in `decode > generateSharedKey`
- Explicitly label which publicKey is self and which one is peer in both `encode/decode` methods.

### DIFF
--- a/packages/core/src/controllers/crypto.ts
+++ b/packages/core/src/controllers/crypto.ts
@@ -102,7 +102,9 @@ export class Crypto implements ICrypto {
     const params = validateEncoding(opts);
     const message = safeJsonStringify(payload);
     if (isTypeOneEnvelope(params)) {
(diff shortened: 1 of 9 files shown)


pedrouid (approved):
Approved 👍 (also changed the target branch so we can publish this with beta.102)

xzilja (approved):
Seems like some changes from derek's pr are also in here, so this branch is not synced with v2 maybe?

PR #1243, opened 2022-07-13
fix: missing metadata

ganchoradkov:
Fixed a bug where `tag` property was not provided to `rpcPublish` method and sequentially was never received by the relay.

Also added the `tag` & `prompt` properties when the `checkQueue` method is invoked


xzilja (approved):
cc @pedrouid to review tag related bits

bkrem (approved):

PR #1244, opened 2022-07-13
Add Tags per RPC method for Sign Client

### LINKED ISSUE #1242: SDK granular tags
Implement spec https://github.com/WalletConnect/walletconnect-specs/issues/84

pedrouid:
Everything works for Granular tags but unfortunately had to comment out the TTL because it was causing tests to sporadically fail..

So this PR is good enough for RC.1 but we need review specs for very short TTL such as 30 seconds for ping methods

### DIFF
--- a/packages/sign-client/src/constants/engine.ts
+++ b/packages/sign-client/src/constants/engine.ts
@@ -1,3 +1,170 @@
+import { ONE_DAY, FIVE_MINUTES, THIRTY_SECONDS } from "@walletconnect/time";
+import { RelayerTypes, JsonRpcTypes } from "@walletconnect/types";
+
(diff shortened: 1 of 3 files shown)


xzilja (approved):
Clean af 👌

bkrem:
@pedrouid do we know why the tests were failing with the TTL defined or TBD?

bkrem (approved):

pedrouid:
> @pedrouid do we know why the tests were failing with the TTL defined or TBD?

I'm assuming that some of these TTLs are too short for publishing and introduce some race conditions on our side

Alternatively I recall Relay Server being limited to 6 hour TTLs which also could be causing issues with 24 hour TTLs

PR #1248, opened 2022-07-14
fix: sign-client build is broken

arein:
Build symlinks were broken. This is a hotfix as this might happen again. We have a plan to fix long-term.

# Testing

Tested locally


xzilja (approved):
🛥️ Dockerise it!

ganchoradkov (approved):

PR #1251, opened 2022-07-14
Add SDK user-agent support

### LINKED ISSUE #1241: SDK user-agent support
Implement the spec https://github.com/WalletConnect/walletconnect-specs/issues/85

pedrouid:
Fix #1241

### DIFF
--- a/packages/utils/src/misc.ts
+++ b/packages/utils/src/misc.ts
@@ -129,7 +129,6 @@ export function formatRelayRpcUrl({
   const splitUrl = relayUrl.split("?");
   const ua = formatUA(protocol, version, sdkVersion);
   const params = { auth, ua, projectId };
-  if (typeof params.projectId === "undefined") delete params.projectId;
   const queryString = appendToQueryString(splitUrl[1] || "", params);
   return splitUrl[0] + "?" + queryString;
 }
--- a/packages/utils/src/misc.ts
+++ b/packages/utils/src/misc.ts
@@ -98,6 +98,7 @@ export function getRelayClientMetadata(protocol: string, version: number): Relay
 
 export function getJavascriptOS() {
   const info = detect();
+  if (info === null) return "unknown";
   const os = info.os.replace(" ", "").toLowerCase();
   if (info.type === "browser") {
     return [os, info.name, info.version].join("-");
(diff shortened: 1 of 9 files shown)


pedrouid:
So unfortunately I was able to break my development build by doing a clean `git clone` on my machine which delayed my delivery with this PR

Still I created some mock data to test it externally I was able to format a UA string to comply with WalletConnect/walletconnect-specs#85

Screenshot 2022-07-14 at 15 54 11

xzilja (approved):
Nice, remind me to test this in react-native at some point to see what sort of data we get there. From memory, it had a bit of a different flow when it comes to user agents.

pedrouid:
> Nice, remind me to test this in react-native at some point to see what sort of data we get there. From memory, it had a bit of a different flow when it comes to user agents.

Yeah, I've some cases for it and I'm using this nice library I've used before called `detect-browser` which also works with react-native

packages/utils/src/misc.ts
```diff
@@ -86,23 +89,47 @@ export function getRelayClientMetadata(protocol: string, version: number): Relay
...
   relayUrl,
+  sdkVersion,
   auth,
   projectId,
 }: RelayerTypes.RpcUrlParams) {
   const splitUrl = relayUrl.split("?");
-  const metadata = { ...getRelayClientMetadata(protocol, version), auth };
-  const params = projectId ? { ...metadata, projectId } : metadata;
+  const ua = formatUA(protocol, version, sdkVersion);
+  const params = { auth, ua, projectId };
```
bkrem, line 131: Minor: Think we can save ourselves the inline `if` with object mutation if we keep the ternary style:

```suggestion
  const params = projectId ? { auth, ua, projectId } : { auth, ua };
```

bkrem (approved):

packages/utils/src/misc.ts
pedrouid: honestly we should have a util that sanitizes "undefined" values instead because what I don't like about the ternary approach is that you have to duplicate the object for the other values

However this delete might be unnecessary too... if the "query-string" package already sanitizes undefined values

bkrem: > what I don't like about the ternary approach is that you have to duplicate the object for the other values

Yeah I get that too, neither approach is ideal in the end. Let's just leave it as is, mostly stylistic anyways 👍

PR #1258, opened 2022-07-15
add reject reason default

pedrouid:


pedrouid:
On second thought.... I'm not sure this is ideal because Wallets should be explicit about the reasons they reject to ensure they are not just blindly rejecting as a user action when in reality it could be incompatible namespaces

PR #1271, opened 2022-07-20
fix: pending subscription resolve

ganchoradkov:
A fix for a rare bug where the session event e.g. `wc_sessionSettle` can be received before the subscription topic is resolved by the subscriber. This causes the event to be ignored.

The implementation adds a mechanism to wait for the subscription to resolve before processing the event


xzilja:
Code wise it's a 👍 , but I'm still not sure what the cause of the bug is, could you provide deeper explanation please? And cc @pedrouid into here as he has best understanding of subscriber controller.

ganchoradkov:
> Code wise it's a +1 , but I'm still not sure what the cause of the bug is, could you provide deeper explanation please? And cc @pedrouid into here as he has best understanding of subscriber controller.

In a nutshell, the session event can be received while the subscription topic is still pending [during this call](https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/core/src/controllers/subscriber.ts#L91)

This is causing [this check](https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/core/src/controllers/relayer.ts#L180) to return `true` and ignore the event while it **shouldn't**

packages/core/src/controllers/subscriber.ts
```diff
@@ -37,7 +39,7 @@ export class Subscriber extends ISubscriber {
 
   private cached: SubscriberTypes.Active[] = [];
   private initialized = false;
-
+  private pendingSubscriptionWatchLabel = PENDING_SUB_WATCH_LABEL;
```
bkrem: Do we need to define an shared external constant here if the label value is only used in the subscriber and is private?

packages/core/src/controllers/subscriber.ts
```diff
@@ -109,6 +111,33 @@ export class Subscriber extends ISubscriber {
...
+        }
+        if (
+          watch.elapsed(this.pendingSubscriptionWatchLabel) / 1000 >=
+          PENDING_SUB_RESOLUTION_TIMEOUT
+        ) {
+          clearInterval(interval);
+          watch.stop(this.pendingSubscriptionWatchLabel);
+          reject(false);
+        }
+      }, 20);
```
bkrem: Probably cleaner to formalise the interval into a constant too like with the timeout, to avoid a magic number here

packages/core/src/controllers/subscriber.ts
```diff
@@ -109,6 +111,33 @@ export class Subscriber extends ISubscriber {
...
+      watch.start(this.pendingSubscriptionWatchLabel);
+
+      const interval = setInterval(() => {
+        if (!this.pending.has(topic) && this.topics.includes(topic)) {
+          clearInterval(interval);
+          watch.stop(this.pendingSubscriptionWatchLabel);
+          resolve(true);
+        }
+        if (
+          watch.elapsed(this.pendingSubscriptionWatchLabel) / 1000 >=
```
bkrem: Would it make sense to define the `TIMEOUT` constant in ms (`FIVE_SECONDS * 1000`) to avoid the inline division here?

bkrem (approved):

arein (approved):
nit: as this is a hotfix it's ok to proceed without tests

But can you please add an issue such that we add a test covering this so we don't have regressions?

ganchoradkov:
> nit: as this is a hotfix it's ok to proceed without tests
> 
> But can you please add an issue such that we add a test covering this so we don't have regressions?

Will do in a new PR 👍

PR #1298, opened 2022-07-28
fix: loadtest timeout

ganchoradkov:
Added custom vitest config for sign-client so we can run longer loadtests

### DIFF
--- a/packages/sign-client/vitest.config.ts
+++ b/packages/sign-client/vitest.config.ts
@@ -1,0 +1,8 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    testTimeout: 600_000,
+    hookTimeout: 120_000,
+  },
+});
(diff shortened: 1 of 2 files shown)


packages/sign-client/vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    testTimeout: 600_000,
+    hookTimeout: 120_000,
+  },
```
arein: shouldn't this only affect the canary/load tests?

ganchoradkov: it should, but vitest supports global timeout only.. https://github.com/vitest-dev/vitest/issues/1598
Once they release something like that, we'll update

ganchoradkov: you're correct. Updated

arein (approved):

PR #1341, opened 2022-08-11
Make typing for store generic

devceline:
# Description

- Make typing for store generic to make extending it easier

The store controller is generic and should not make assumptions on the object
types stored by it

## How Has This Been Tested?

2 new tests in `store.spec.ts` to check for retrieving from cached data and duplicate keys 

## Due Dilligence

* [x] Breaking change
* [x] Requires a e2e/integration test update

### DIFF
--- a/packages/core/src/controllers/store.ts
+++ b/packages/core/src/controllers/store.ts
@@ -1,6 +1,11 @@
 import { generateChildLogger, getLoggerContext } from "@walletconnect/logger";
 import { ICore, IStore } from "@walletconnect/types";
-import { getInternalError, isProposalStruct, isSessionStruct } from "@walletconnect/utils";
+import {
+  getInternalError,
+  isProposalStruct,
+  isSessionStruct,
+  isUndefined,
+} from "@walletconnect/utils";
 import { Logger } from "pino";
 import { CORE_STORAGE_PREFIX, STORE_STORAGE_VERSION } from "../constants";
 import isEqual from "lodash.isequal";
@@ -52,11 +57,8 @@ export class Store<Key, Data extends Record<string, any>> extends IStore<Key, Da
         } else if (isSessionStruct(value)) {
           // TODO(pedro) revert type casting as any
           this.map.set(value.topic as any, value);
-        } else if (this.getKey) {
-          // Null checking to prevent undefined or null values to throw an error
-          if (value) {
-            this.map.set(this.getKey(value), value);
-          }
+        } else if (this.getKey && value !== null && !isUndefined(value)) {
+          this.map.set(this.getKey(value), value);
         }
       });
 
--- a/packages/core/src/controllers/store.ts
+++ b/packages/core/src/controllers/store.ts
@@ -5,7 +5,7 @@ import { Logger } from "pino";
 import { CORE_STORAGE_PREFIX, STORE_STORAGE_VERSION } from "../constants";
 import isEqual from "lodash.isequal";
 
-export class Store<Key, Data extends Object> extends IStore<Key, Data> {
+export class Store<Key, Data extends Record<string, any>> extends IStore<Key, Data> {
   public map = new Map<Key, Data>();
   public version = STORE_STORAGE_VERSION;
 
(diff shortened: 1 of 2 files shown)


bkrem:
Commenting what I mentioned on Slack here for visibility, caveats:

- ~Store assumes that the value is effectively an object because it's using spread-merging here for update: https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/core/src/controllers/store.ts#L103~ Edit: Resolved by the fact that value stays as `Record<K,V>` here 👍 

- When restoring from persistence into memory, there's a check for whether `value.proposer.publicKey` or `value.topic` exists: https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/core/src/controllers/store.ts#L36

packages/core/src/controllers/store.ts
```diff
@@ -40,6 +52,11 @@ export class Store<Key, Data extends StoreStruct> extends IStore<Key, Data> {
         } else if (isSessionStruct(value)) {
           // TODO(pedro) revert type casting as any
           this.map.set(value.topic as any, value);
+        } else if (this.getKey) {
+          // Null checking to prevent undefined or null values to throw an error
```
xzilja: Could all of this be combined into `else if (this.getKey && value)` ? ideally also used alongside `isUndefined` util to prevent truthy values slipping by

packages/core/src/controllers/store.ts
```diff
@@ -1,30 +1,42 @@
...
   private initialized = false;
 
+  /**
+   * Regenerates the value key to retrieve it from cache
+   */
+  private getKey: ((data: Data) => Key) | undefined;
+
   private storagePrefix = CORE_STORAGE_PREFIX;
 
+  /**
```
bkrem, line 27: I think we can remove this, only makes sense to add these JSDoc style comments if we do it everywhere.

xzilja (commented):
I assume this is needed to make store more reusable in other sdk's like chat and auth, correct?

packages/core/src/controllers/store.ts
```diff
@@ -1,30 +1,42 @@
...
+   * @param {Store<Key, Data>["getKey"]} getKey Regenerates the value key to retrieve it from cache
+   * @param {string} storagePrefix Prefixes value keys
+   */
   constructor(
     public core: ICore,
     public logger: Logger,
     public name: string,
-    storagePrefix = CORE_STORAGE_PREFIX,
+    storagePrefix: string = CORE_STORAGE_PREFIX,
+    getKey: Store<Key, Data>["getKey"] = undefined,
```
bkrem, line 39: Think this extra constructor param needs to be accounted for in `IStore`

devceline: I didn't want to make the assumption that all `IStore` implementers will need to have this functionality. One could argue that one could have an `IStore` implementation that does not use the same caching mechanisms, no?

packages/core/src/controllers/store.ts
devceline: Felt like it made sense specifically because `getKey` is not immediately straightforward.

Maybe I can remove the JSDoc and change the function's name to `getKeyForCache` or something, what do you think

bkrem:
> I assume this is needed to make store more reusable in other sdk's like chat and auth, correct?

Exactly, at the moment the accepted values are opinionated towards Sign.

packages/core/src/controllers/store.ts
bkrem: Yeah fair point, no need to add the assumption at this stage 👍

packages/core/src/controllers/store.ts
bkrem: Yeah assuming the type inference is meaningful enough when inspecting the constructor param, I think we can go with `getKeyForCache` or `getKeyForCacheRestore`.

If you feel like it's a lot better with the JSDoc comment feel free to leave it though 👍

devceline: I personally feel like it makes sense to leave JSDoc in, maybe in the future `getKey` can be expanded on and used for other reasons (eg: Logging) so just adding some documentation doesn't hurt anyone I think. But I do not feel _very_ strongly about this.

bkrem: Ok let's leave it there :)

bkrem (approved):
Thank you for this @devceline 🙌 Nice work on the tests too 💯

xzilja (approved):

PR #1343, opened 2022-08-12
feat: Reorder clientMeta `||`

HarryET:
# Description
When `clientMeta` is set it defaults to the meta that is found on the page rather than the provided meta

## How Has This Been Tested?
Ran `npm test`

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/clients/core/src/index.ts
+++ b/packages/clients/core/src/index.ts
@@ -118,7 +118,7 @@ class Connector implements IConnector {
   // -- constructor ----------------------------------------------------- //
 
   constructor(opts: IConnectorOpts) {
-    this._clientMeta = getClientMeta() || opts.connectorOpts.clientMeta || null;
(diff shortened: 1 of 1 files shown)


bkrem (approved):
## Context

Harry and I were both surprised when we realised that user-defined `clientMeta` is never applied as long as `getClientMeta` can grab something, which seems counter-intuitive since user-defined metadata should override defaults.

## Considerations

We can see this in two ways in my eyes:

- This is a bug and we're fixing it here
- This is a (breaking?) change from an intended feature

Think we need your input here @pedrouid since this is from the v1 days

pedrouid:
Actually this was intended for v1.0 as the goal was to have the metadata scrapped from the website to supersede user config

However this was very much redundant as eventually devs would override the private variable after init

That being said... it's pointless to change it now and we should only fix critical bugs for v1.0

HarryET:
Closing because this is intended behaviour.

PR #1346, opened 2022-08-15
fix: failed connections are still retried when closed

QingWei-Li:
# Description

If a wrong bridge is set, the `_nextSocket` will retry indefinitely. The `_nextSocket` does not stop when the connection is closed.

## Due Dilligence

* [ ] Breaking change


bkrem:
Closing. v1 has been deprecated and will only receive critical fixes going forwards.

PR #1348, opened 2022-08-15
V2.0.0 rc.2

### LINKED ISSUE #1338: React Native Sign SDK
## Description

Uppon testing sign SDK versions beta.102, rc.0 and rc.1 in latest react-native version (fresh vanilla app), all of these came up with errors like `new TextEncoder()` being undefined. This happens in both jsc and hermes runtimes and will only get worse as react-native is moving towards preferring hermes. This issue can be solved in 2 ways i.e. heavy js polyfilling or we create @walletconnect/react-native-sign sdk that utalises Swift and Kotlin SDK's. Latter is preferred way and will result in multiple benefits for react-native users i.e. logic will run on native thread and utilize native methods for storage / networking etc. This however needs a bit of research to understand if we can successfully implement Swift and Kotlin SDKs under one nice react-native js api similar to that of js package.

## Tasks

- [ ] Create test sdk package that exposes Kotlin sdk and test against it
- [ ] If task no.1 is successful, continue with Swift sdk implementation
- [ ] If task no.1 fails, continue with js sdk polyfilling

xzilja:
# Description

Created new package `@walletconnect/react-native-compat` that includes all dependencies and polyfills that are necessary to run our js sdk. (some dependencies were specified as peers as they include native code that needs to be linked thus users need to include it in their own package json's). After this install for react native will look something like this

```
npm i @walletconnect/sign-client @walletconnect/react-native-compat @react-native-async-storage/async-storage react-native-get-random-values
```

And users need to import compat package at their App component i.e.
```ts
import "@walletconnect/react-native-compat:

export default App() { }
```

In addition to above
- Updated dependencies
- Removed un-necessary deps like `events` that were needed for rn
- Tagged rc.2 (but not published yet)

Resolves #1338

## How Has This Been Tested?

- Ran build
- Ran tests (push ones are failing, but this has been an issue for me ever since they got added need to resolve this with Derek)
- Installed and tested latest canary that includes these changes in web-example wallet
- Installed and tested latest canary in react-native example (tested up to pairing proposal)

https://user-images.githubusercontent.com/3154053/184617706-e86a565e-71fb-42c5-ad40-2ae6d735e9a7.mov

## Due Dilligence

* [ ] Breaking change


ganchoradkov (approved):

rollup.config.js
```diff
@@ -1,16 +1,11 @@
...
 
 const input = "./src/index.ts";
 const plugins = [
-  nodeResolve({
-    preferBuiltins: false,
-    browser: true,
-  }),
+  nodeResolve({ preferBuiltins: false, browser: true }),
   commonjs(),
-  nodePolyfills(),
```
bkrem, line 13: Nice 🙌  are we sure we can remove these already, i.e. aren't they still needed for pino for example?

bkrem (approved):

devceline (approved):
LGTM

rollup.config.js
xzilja: I tested dapp example with this version and all loggs are still there, I assume thats pino working yeh?

bkrem: Ok let's go for it 👍

bkrem:
> (push ones are failing, but this has been an issue for me ever since they got added need to resolve this with Derek)

btw @IljaDaderko what seemed to resolve this for me was pulling the latest ts-relay

PR #1350, opened 2022-08-16
feat: universal provider

### LINKED ISSUE #1339: [Epic] v2 Provider
A universal provider has to be developed supporting v2 multi-chain
Ideally, It should support the following namespaces and their respective chains

- eip155
- solana
- cosmos 

General overview of the steps needed to accomplish this task

- [x] Create `UniversalProvider` class (naming could change) that can receives config object as constructor with the relevant namespaces, relay url etc 
- [x] Add general event listeners & handlers 
- [x] Initialize `sign-client` with the provided config
- [x] Create provider classes such as`Eip155Provider` , `SolanaProvider` that are responsible are initialized if their namespace is used and are responsible for handling their respective logic and rpc requests
- [x] `Eip155Provider` - Implement `request` & `requestAsync` methods in providers that handle json rpc methods & wallet communication
- [x] `Eip155Provider` - Implement namespace validations where the provided namespace/chainId as fx parameter is used in requests with fallback to the default initialized when no parameter is provided
- [x] `Eip155Provider` - Implement `http json rpc` provider to handle requests not requiring wallet forwarding
- [x] Create types & interfaces
- [x] Create test cases for each provider
- [ ] `SolanaProvider` - Implement `request` & `requestAsync` methods in providers that handle json rpc methods & wallet communication
- [ ] `SolanaProvider` - Implement namespace validations where the provided namespace/chainId as fx parameter is used in requests with fallback to the default initialized when no parameter is provided
- [ ] `SolanaProvider` - Implement `http json rpc` provider to handle requests not requiring wallet forwarding
- [ ] `CosmosProvider` - Implement `request` & `requestAsync` methods in providers that handle json rpc methods & wallet communication
- [ ] `CosmosProvider` - Implement namespace validations where the provided namespace/chainId as fx parameter is used in requests with fallback to the default initialized when no parameter is provided
- [ ] `CosmosProvider` - Implement `http json rpc` provider to handle requests not requiring wallet forwarding

ganchoradkov:
# Description
This draft PR is the initial implementation of the `universal provider` for walletconnect v2 protocol. 

## Usage

```typescript
import { ethers } from "ethers";
import UniversalProvider from "@walletconnect/universal-provider";

//  Initialize the provider
const provider = await UniversalProvider.init({
  logger: "info",
  relayUrl: "ws://<relay-url>",
  projectId: "12345678",
  metadata: {
    name: "React App",
    description: "React App for WalletConnect",
    url: "https://walletconnect.com/",
    icons: ["https://avatars.githubusercontent.com/u/37784886"],
  },
  client: undefined, // optional instance of @walletconnect/sign-client
});

//  create sub providers for each namespace/chain
await provider.connect({
    namespaces: {
      eip155: {
        methods: [
          "eth_sendTransaction",
          "eth_signTransaction",
          "eth_sign",
          "personal_sign",
          "eth_signTypedData",
        ],
        chains: ["eip155:80001"],
        events: ["chainChanged", "accountsChanged"],
        rpcMap: {
          80001: "https://rpc-mumbai.matic.today",
        },
      },
      pairingTopic: "<123...topic>", // optional topic to connect to
      skipPairing: false, // optional to skip pairing ( later it can be resumed by invoking .pair())
    },
  });

  //  Create Web3 Provider
  const web3Provider = new ethers.providers.Web3Provider(provider);
```

## Events

```typescript
// Subscribe for pairing URI
provider.on("display_uri", (uri) => {
  console.log(uri);
});

// Subscribe to session ping
provider.on("session_ping", ({ id, topic }) => {
  console.log(id, topic);
});

// Subscribe to session event
provider.on("session_event", ({ event, chainId }) => {
  console.log(event, chainId);
});

// Subscribe to session update
provider.on("session_update", ({ topic, params }) => {
  console.log(topic, params);
});

// Subscribe to session delete
provider.on("session_delete", ({ id, topic }) => {
  console.log(id, topic);
});
```

## Provider Methods

```typescript
interface RequestArguments {
  method: string;
  params?: any[] | undefined;
}

// Send JSON RPC requests

/**
 * @param payload
 * @param chain - optionally specify which chain should handle this request
 * in the format `<namespace>:<chainId>` e.g. `eip155:1`
 */
const result = await provider.request(payload: RequestArguments, chain: string | undefined);
```

## Due Dilligence

* [ ] Breaking change


providers/universal-provider/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "build:types": "tsc",
+    "build:source": "rollup --config rollup.config.js",
+    "build": "npm run build:pre; npm run build:source; npm run build:types",
+    "test:pre": "rm -rf ./test/test.db",
+    "test:run": "vitest run --dir test",
+    "test": "npm run test:pre; npm run test:run",
+    "lint": "eslint -c '../../.eslintrc' --fix './src/**/*.ts'",
+    "prettier": "prettier --check '{src,test}/**/*.{js,ts,jsx,tsx}'"
+  },
+  "dependencies": {
```
xzilja, line 34: I think some of these dependencies are only needed for testing purposes, correct? If that is the case, could you move relevant ones to `devDependencies` so they don't reflect as deps in npm registry. (re install them to update pakage lock file as well)

providers/universal-provider/README.md
```diff
@@ -0,0 +1,96 @@
...
+  logger: "info",
+  relayUrl: "ws://<relay-url>",
+  projectId: "12345678",
+  metadata: {
+    name: "React App",
+    description: "React App for WalletConnect",
+    url: "https://walletconnect.com/",
+    icons: ["https://avatars.githubusercontent.com/u/37784886"],
+  },
+  client: undefined, // optional instance of @walletconnect/sign-client
```
xzilja, line 22: What are the circumstances where optional instance of client might be needed?

providers/universal-provider/test/index.spec.ts
```diff
@@ -0,0 +1,329 @@
+import { describe, it } from "vitest";
+// import Web3 from "web3";
```
xzilja: Are tests coming latter on, or why are these commented out?

xzilja (commented):
Looking goood 🔥 Will need to wait for tests to get implemented for final approval. One question, what's the idea behind `subProviders` folder? Is it to simplify implementation even further so that we can create those small chain configs ourselves?

providers/universal-provider/package.json
ganchoradkov: > Looking goood fire Will need to wait for tests to get implemented for final approval. One question, what's the idea behind `subProviders` folder? Is it to simplify implementation even further so that we can create those small chain configs ourselves?

Think of `subProviders` as controllers. Each namespace will have its own implementation where logic can't be shared between chains.

providers/universal-provider/test/index.spec.ts
ganchoradkov: Tests are on the way 🚄

providers/universal-provider/README.md
```diff
@@ -0,0 +1,96 @@
...
+Universal Provider for WalletConnect Protocol
+
+## Usage
+
+```typescript
+import { ethers } from "ethers";
+import UniversalProvider from "@walletconnect/universal-provider";
+
+//  Initialize the provider
+const provider = await UniversalProvider.init({
```
arein, line 12: nit: why provide the rpc in `connect` not here?

providers/universal-provider/src/types/providers.ts
```diff
@@ -0,0 +1,36 @@
...
+import {
+  RpcProvidersMap,
+  RpcProviderMap,
+  RequestParams,
+  RequestArguments,
+  SessionNamespace,
+  NamespaceConfig,
+} from ".";
+
+export interface ISubProvider {
```
arein:
```suggestion
export interface IProvider {
```

arein (commented):
Awesome! minot nitpicks - Some tests and then LGTM

providers/universal-provider/README.md
ganchoradkov: the idea for the rpc config being in the `connect` fx and not in the `init` was that in some use-cases you might not know which chains will be used at the time of initialization

providers/universal-provider/test/index.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+        provider.on("session_delete", () => {
+          resolve();
+        });
+      }),
+      new Promise<void>(async (resolve) => {
+        await walletClient.disconnect();
+        resolve();
+      }),
+    ]);
+    // expect provider to be disconnected
```
xzilja: are these left over comments, or something you want to add in the future?

xzilja (approved):
Looks good, now that the tests are in. I think we can merge this (perhaps after you address those `FIXME` comments?). Would be good to make a release of this as well and perhaps create a pr to wagmi to add as v2 provider.

providers/universal-provider/README.md
```diff
@@ -0,0 +1,112 @@
...
+  logger: "info",
+  relayUrl: "ws://<relay-url>",
```
xzilja, line 14: Should this be wss?

providers/universal-provider/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "@walletconnect/jsonrpc-http-connection": "1.0.3",
```
xzilja, line 35: Before release, could you double check that these are all up to date

providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+            events: this.events,
+          });
+          break;
+        case "solana":
+          //TODO:
+          break;
+        case "cosmos":
+          //TODO:
+          break;
+        case "polygon":
```
xzilja: I think polygon is handled by Eip155Provider as well, correct?

providers/universal-provider/src/constants/values.ts
```diff
@@ -0,0 +1,5 @@
+export const LOGGER = "error";
+
+export const RELAY_URL = "ws://relay.walletconnect.com";
```
xzilja:
```suggestion
export const RELAY_URL = "wss://relay.walletconnect.com";
```

xzilja (approved):
Super nice! This will unlock so many cool integrations! Left few minor comments and in general if possible, could you look over typescript `any` and add appropriate types where possible, for nicer end user experience.

providers/universal-provider/README.md
```diff
@@ -0,0 +1,112 @@
...
+        "eth_sendTransaction",
+        "eth_signTransaction",
+        "eth_sign",
+        "personal_sign",
+        "eth_signTypedData",
+      ],
+      chains: ["eip155:80001"],
+      events: ["chainChanged", "accountsChanged"],
+      rpcMap: {
+        80001: "https://rpc-mumbai.matic.today",
```
arein:
```suggestion
        80001: "https://rpc.walletconnect.com?chainId=eip155:80001&projectId=<your walletconnect project id>",
```

arein (approved):

providers/universal-provider/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "wallet",
+    "walletconnect"
+  ],
+  "scripts": {
+    "clean": "rm -rf dist",
+    "build:pre": "npm run clean",
+    "build:types": "tsc",
+    "build:source": "rollup --config rollup.config.js",
+    "build": "npm run build:pre; npm run build:source; npm run build:types",
+    "test:pre": "rm -rf ./test/test.db",
```
bkrem: Think this can be removed since it's a side effect for running tests in sign-client specifically

providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -0,0 +1,267 @@
...
+  ConnectParams,
+  RequestArguments,
+  UniversalProviderOpts,
+  NamespaceConfig,
+} from "./types";
+
+import { RELAY_URL, LOGGER, STORAGE } from "./constants";
+
+export class UniversalProvider implements IUniversalProvider {
+  public client!: SignClient;
```
bkrem, line 23: How come we need the non-null assertion operator here but didn't seem to previously in SignClient and other similar classes?

providers/universal-provider/src/constants/values.ts
```diff
@@ -0,0 +1,5 @@
+export const LOGGER = "error";
+
+export const RELAY_URL = "wss://relay.walletconnect.com";
+
+export const STORAGE = "@universal_provider";
```
bkrem: Would it make sense to use a standardised format for the storage key prefixing like we do for Core, Sign etc? https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/core/src/constants/core.ts#L5

Also in terms of future-proofing/versioning the storage for possible breaking changes

providers/universal-provider/src/UniversalProvider.ts
ganchoradkov: Because the client is not initialized in the constructor and thus requiring adding `| undefinied`. 
This makes usage later on annoying due to constantly having to add checks such as `client?.`

Thought this solution is fine as attempting to use the client without it being initialized results in an exception such as
https://github.com/WalletConnect/walletconnect-monorepo/blob/gancho/feat/universal-provider/providers/universal-provider/src/UniversalProvider.ts#L180

bkrem: Yeah makes total sense 👍 

I'm more surprised we didn't run into this before in e.g. SignClient since the pattern is the same, but it seems like the way the type referencing there refers back to the abstract class every time (i.e. `public session: ISignClient["session"]`) doesn't evoke this issue in the TS compiler

providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -0,0 +1,267 @@
...
+        projectId: this.providerOpts.projectId,
+        metadata: this.providerOpts.metadata, // fetch metadata automatically if not provided?
+      }));
+
+    this.logger.trace(`SignClient Initialized`);
+  }
+
+  private createProviders(): void {
+    if (!this.client) {
+      throw new Error("Client not initialized");
```
bkrem: Consistency nitpick:

```suggestion
      throw new Error("Sign Client not initialized");
```

providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    Object.keys(this.namespaces).forEach((namespace) => {
+      switch (namespace) {
+        case "eip155":
+          this.rpcProviders[namespace] = new Eip155Provider({
+            // @ts-ignore
```
bkrem: Still needed?

providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -0,0 +1,267 @@
...
+        case "polkadot":
+          //TODO:
+          break;
+      }
+    });
+  }
+
+  private registerEventListeners(): void {
+    if (typeof this.client === "undefined") {
+      throw new Error("WalletConnect is not initialized");
```
bkrem: Consistency nitpick:

```suggestion
      throw new Error("Sign Client is not initialized");
```

bkrem (approved):

bkrem:
Amazing work on this @ganchoradkov 💯

PR #1393, opened 2022-08-29
fix: not filtering out unsuccessful values

arein:
# Description

We are filtering against `-1` not `0`

Towards https://github.com/WalletConnect/rs-relay/issues/153

## How Has This Been Tested?

Tested locally

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/sign-client/test/concurrency/concurrency.spec.ts
+++ b/packages/sign-client/test/concurrency/concurrency.spec.ts
@@ -145,7 +145,7 @@ describe("Sign Client Concurrency", () => {
           })
           .filter(
             (i: { handshakeLatencyMs: number; pairingLatencyMs: number }) =>
-              i.handshakeLatencyMs !== -1,
+              i.handshakeLatencyMs !== -1 && i.pairingLatencyMs !== -1,
           ),
       );
       const averagePairingLatency =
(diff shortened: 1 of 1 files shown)


packages/sign-client/test/concurrency/concurrency.spec.ts
```diff
@@ -123,8 +123,8 @@ describe("Sign Client Concurrency", () => {
               async (resolve) => {
                 const timeout = setTimeout(() => {
                   log(`Client ${i} hung up`);
-                  resolve({ handshakeLatencyMs: 0, pairingLatencyMs: 0 });
-                }, 90_000);
+                  resolve({ handshakeLatencyMs: -1, pairingLatencyMs: -1 });
```
xzilja, line 126: The only comment I'd have is to make these values constants i.e.  what does `-1` mean here?

xzilja (approved):

packages/sign-client/test/concurrency/concurrency.spec.ts
arein: I now I slipped this past you - will update later to something more meaningful as we keep iterating on the load test :)

PR #1397, opened 2022-08-30
fix(test): fixes separate clients using same persistent DB

bkrem:
# Description

- Ensures test clients are initialized with their own respective sqlite DBs when the relevant `storageOption` is passed.
- Also: fixes `test:pre` script not cleaning up test DBs as expected (creation and deletion paths were different)

### Context

@ennmichael discovered the following while running the sign-client tests against `rs-relay`:

> The `clients can ping each other after restart` test does in fact generate the same client IDs for client A and client B. The root cause of the issue is that `getClientSeed` returns the same value for both of them. 

- Since both clientA and clientB are initialised to use the same `test.db` reference and the key for the client seed is simply `"client_ed25519_seed"` (no namespacing), this resulted in both clients `get`'ing the same seed [here](https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/core/src/controllers/crypto.ts#L146) between them once one had been generated.

## How Has This Been Tested?

- Tested locally against rs-relay via `TEST_RELAY_URL=ws://127.0.0.1:8080 yarn test`

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,10 @@ zip
 tsconfig.tsbuildinfo
 test.db
 test.db-journal
+client_a.db
+client_a.db-journal
+client_b.db
+client_b.db-journal
 setup
 result*
 ops/grafana/grafana.ini
--- a/packages/sign-client/package.json
+++ b/packages/sign-client/package.json
@@ -22,7 +22,7 @@
     "build:types": "tsc",
     "build:source": "rollup --config rollup.config.js",
     "build": "npm run build:pre; npm run build:source; npm run build:types",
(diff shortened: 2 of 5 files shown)


xzilja (approved):
What a catch 👌 Also since client options for initTwoClients are always `{}` here, maybe move them to be last optional argument that defaults to `{}` so they don't need to be specified all the time.

arein (approved):
Tough one to sort out..

ganchoradkov (approved):

PR #1404, opened 2022-08-31
fix(test): avoids race condition when binding `session_proposal` listener

bkrem:
# Description

- Part of https://github.com/WalletConnect/rs-relay/issues/322
- Fixes the `connect (with old pairing)` test

Also:
- chore(test): skip currently moot `.extend` test until its refactored (WIP is here: https://github.com/WalletConnect/walletconnect-monorepo/commit/2612a126fb7078277eedd1bfbb2c40e9aaab9f32#)

## Context

From [Slack thread](https://walletconnect.slack.com/archives/C03SMNKLPU0/p1661954375463839?thread_ts=1661793364.945769&cid=C03SMNKLPU0):

> testConnectMethod first calls await A.connect() and only then further down binds the needed B.once("session_proposal") to listen to the incoming proposal on Client B.
> 
> - On ts-relay, we somehow get away with this, where both in the new pairing and known pairing scenarios the listener gets bound on Client B before it receives the actual proposal.
> - On rs-relay, this seems to work on the new pairing flow, but breaks on the known pairing flow; i.e. session_proposal is emitted on B before it has had the .once("session_proposal",...) listener set.

## How Has This Been Tested?

- Tested against rs-relay locally

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/sign-client/test/sdk/client.spec.ts
+++ b/packages/sign-client/test/sdk/client.spec.ts
@@ -34,7 +34,7 @@ describe("Sign Client Integration", () => {
         pairingTopic,
       });
       deleteClients(clients);
-    }, 120_000);
+    });
   });
 
   describe("disconnect", () => {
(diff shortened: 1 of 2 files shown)


arein (approved):
Sweet! thank you!

PR #1411, opened 2022-09-06
fix(engine): ensure requesting client sets own record pre `sendRequest`

bkrem:
# Description

* Fixes race condition in setting of JSON-RPC history records in engine.
* Part of https://github.com/WalletConnect/rs-relay/issues/322

## Context

* Scenario: clientA sends `wc_pairingPing` to clientB
* clientA was setting the history record _after_ it sent a JSON-RPC request to its peer rather than before.
* rs-relay seems to sometimes process the full request-response so much faster that clientB had responded before the record for its own request had even been set by clientA.
* This caused the "no record in history" error when the response from clientB came back and `onRelayEventResponse > this.client.history.get(...)` was called.

## How Has This Been Tested?

- Verified that race condition no longer appears locally by doing 10-15 runs.
- CI runs in `rs-relay` against fix with additional logging: https://github.com/WalletConnect/rs-relay/runs/8211851657?check_suite_focus=true

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -398,8 +398,8 @@ export class Engine extends IEngine {
     const payload = formatJsonRpcRequest(method, params);
     const message = await this.client.core.crypto.encode(topic, payload);
     const opts = ENGINE_RPC_OPTS[method].req;
-    await this.client.core.relayer.publish(topic, message, opts);
     this.client.history.set(topic, payload);
+    await this.client.core.relayer.publish(topic, message, opts);
 
     return payload.id;
   };


arein (approved):

bkrem:
Tests in `relayer.spec.ts` and `subscriber.spec.ts` all returning 400 bad request for some reason, unrelated to this change (also happens on `v2.0`).

Seems to be taking in the same env vars as the signClient tests though, not sure what this is about.

xzilja (approved):
I'd try to first wrap that `history.set` in 

```
setTimeout(() => this.client.history.set(topic, payload), 0)
```

which will basically call it in the next cpu "tick", maybe that will sort out race condition with rs relay.

If not, maybe cache previous history state and add `.catch` clause to publish to revert it back in case it fails, otherwise we end up with incorrect history state.

bkrem:
> If not, maybe cache previous history state and add .catch clause to publish to revert it back in case it fails, otherwise we end up with incorrect history state.

I don't see this, why incorrect history state? ClientB already sets the record immediately on the receiving end before it processes the incoming request, so this issue only occurs on the requesting side because the requester isn't setting the record up front.

bkrem:
Merging this since the failing tests in core are unrelated and are effectively already present on `v2.0`. Tried to debug what's wrong but haven't figured it out yet, so will keep looking tomorrow.

PR #1424, opened 2022-09-09
fix(engine): ensure proposal/session expiry is calculated dynamically

bkrem:
# Description

- Proposal/Session expiry was previously calculated on `constants/*` file load, rather than dynamically at call time like other `calcExpiry` calls.
- In a long running client this could lead to a a shorter and shorter expiry duration, even if it expiry is extended, since `calcExpiry` was fixed to the time of initialisation.
- This PR:
	- ensures expiry calculation is dynamic and done at call site.
	- Reimplements the associated unit test
- Follow-up to: https://github.com/WalletConnect/rs-relay/issues/322

## How Has This Been Tested?

- Tested locally
- In demo wallet/dapp via [canary release](https://www.npmjs.com/package/@walletconnect/sign-client/v/2.0.0-4a11a75.0)

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/sign-client/src/constants/proposal.ts
+++ b/packages/sign-client/src/constants/proposal.ts
@@ -1,6 +1,5 @@
-import { calcExpiry } from "@walletconnect/utils";
 import { THIRTY_DAYS } from "@walletconnect/time";
 
 export const PROPOSAL_CONTEXT = "proposal";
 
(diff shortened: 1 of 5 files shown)


xzilja (approved):
Great catch!

PR #1433, opened 2022-09-11
Set build target to ES2020

xzilja:
# Description

We are currently targeting `ESNext` as our build target, which is probably not a good idea. This has been changed to target `ES2020` according to platforms we target and their compatibility with feature set in this spec of js, for more info see https://kangax.github.io/compat-table/es2016plus/

## How Has This Been Tested?
Ran build with target versions set to `ES2015` and `ES2020` to confirm that esbuild is correctly picking these settings up from tsconfig at the root (compared output files and they were indeed different, with 2015 version having bigger size as expected). Final build was set to `ES2020` and is passing.

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,8 +12,9 @@
     "experimentalDecorators": true,
     "forceConsistentCasingInFileNames": true,
     "importHelpers": true,
-    "lib": ["ESNext", "DOM"],
-    "module": "ESNext",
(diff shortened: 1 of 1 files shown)


xzilja:
Looks like tests are failing due to hanging socket

bkrem (approved):

PR #1449, opened 2022-09-17
test: adds `test:ignoreUnhandled` script to unblock rs-relay pipeline

bkrem:
# Description

- Adds `test:ignoreUnhandled` script
- This is to unblock automatic deploys for rs-relay failing unnecessarily, but does not resolve the underlying issue/possible leftover race condition around `JsonRpcHistory` we still see only in the rs-relay `validate_js` pipeline: https://github.com/WalletConnect/rs-relay/actions/runs/3073322744/jobs/4965541078#step:4:105
- Once the cause of the unhandled `JsonRpcHistory` error in has been found, we should remove this script again.

## How Has This Been Tested?

- Tested locally

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "prettier": "npm run prettier --workspaces --if-present",
     "build": "npm run build --workspaces --if-present",
     "test": "npm run test --workspaces --if-present",
+    "test:ignoreUnhandled": "npm run test:ignoreUnhandled --workspaces --if-present",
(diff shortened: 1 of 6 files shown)


bkrem:
@arein could you update the `validate_js` job in rs-relay to use `npm run test:ignoreUnhandled` if/when this gets merged? Or even better if we can run against this branch first without merging to `v2.0`

arein (approved):

PR #1455, opened 2022-09-20
feat: log client ids when tests fail

arein:
# Description

These tests occasionally fail in the pipeline. In order to allow further debugging, log the client ids

## How Has This Been Tested?

Tested locally

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/sign-client/test/sdk/client.spec.ts
+++ b/packages/sign-client/test/sdk/client.spec.ts
@@ -138,7 +138,6 @@ describe("Sign Client Integration", () => {
         });
         afterEach(async (done) => {
           const { result } = done.meta;
-          console.log("dang", result?.state.toString());
           if (result?.state.toString() !== "pass") {
             console.log(
               `Test ${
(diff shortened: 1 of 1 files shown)


xzilja (approved):

packages/sign-client/test/sdk/client.spec.ts
```diff
@@ -131,21 +131,36 @@ describe("Sign Client Integration", () => {
...
-        await clients.B.ping({ topic });
-        deleteClients(clients);
+      describe("existing session", () => {
+        let clients;
+        beforeEach(async () => {
+          clients = await initTwoClients();
+        });
+        afterEach(async (done) => {
+          const { result } = done.meta;
+          console.log("dang", result?.state.toString());
```
bkrem: Needed? Can we use a different prefix if it is? 😄

bkrem:
Since this will only print clientIds if the issue happens in that specific sub-suite of tests, does it also make sense to put a `try/catch` in `testConnectMethod` and log this there whenever something fails?

bkrem (approved):

arein:
> Since this will only print clientIds if the issue happens in that specific sub-suite of tests, does it also make sense to put a `try/catch` in `testConnectMethod` and log this there whenever something fails?

Probably. We already have a timeout there. Printing the client id there makes sense.

Issue #1364, opened 2022-08-19
Unable to pair iOS Lobstr with `2.0.0-rc`


BenRacicot:
**Describe the bug**
When connecting my localhost app and scanning the QR URI with the Lobstr iOS app nothing happens for about 1 minute until (the timeout) errs:

```
wc:09d4...@2?relay-protocol=irn&symKey=0c8a...

context: "client/expirer" ERROR 'No matching key. expirer: topic:09d4...'
```
UPDATE:
There's also another version of this error seen in other scenarios:
```
No matching key. history:
```

**SDK Version (if relevant)**
    "@walletconnect/sign-client": "^2.0.0-rc.3",
    "@walletconnect/qrcode-modal": "^1.8.0",    
    "@walletconnect/utils": "2.0.0-rc.3",    
    "@walletconnect/types": "^2.0.0-rc.3",

**To Reproduce**
1. Run RC-3, init a connection to create a URI (pop the QR modal)
2. Open Lobstr's latest app and scan your app's QR/URI to connect WC
3. Nothing happens in Lobstr
4. Eventually the `No matching key` error shows in the RC-3 app console.

**Expected behavior**
The WC flow happens when the Lobstr app scans the URI

**Desktop**
 - OS: MacOS
 - Browser Chrome / 104
**Smartphone**
 - Device: iphone 12 / iOS 15.6.1
 - Lobstr App: 8.7.0_311

**Additional context**
Perhaps the RCs are not backwards compatible but it seems the URI should be recognized by Lobstr.

jmysliv:
hey! I have the same issue. Have you found what's wrong and how it can be fixed?

arein:
Apologies for this. We've occasionally seen this in internal benchmarks. @bkrem and I will investigate.

BenRacicot:
No worries! We're here to see WC be the best!
I've spoken to other devs about this and it would seem that as various pairings expire we get this error in the console? 
Not sure still but it does persist in RC-3

arein:
That's a good hypothesis. We'll chase it down. We also see it internally in our test suite which is probably another issue. We will chase down your hypothesis and update here.

arein:
@BenRacicot @jmysliv the issue you're seeing should have been fixed in https://github.com/WalletConnect/walletconnect-monorepo/releases/tag/2.0.0-rc.3 can you please upgrade? If you still see this after upgrading please reopen the ticket

BenRacicot:
[...]

PR #1459, opened 2022-09-20
fix(engine): avoids unwanted expirer deletion attempts post-expiry

bkrem:
# Description

- Mentioned in (but not cause of) #1364 
- We sometimes see expirer error logging (e.g. `context: "client/expirer" ERROR 'No matching key. expirer: topic:09d4...'`) when a pairing or proposal is being cleaned up on expiry of said proposal/pairing.
- The cause of this is:

1. proposal/pairing/session expiry is detected inside `expirer` via heartbeat checks, **`expirer` deletes the tracked expiry internally immediately**
2. `EXPIRER_EVENTS.expired` gets called in engine and triggers e.g. `deleteProposal`/`deletePairing`/`deleteSession`
3. The delete helper tries to re-delete the already removed expiry via `expirer.del` -> triggering a "key not found" error.

- In cases where the trigger for deletion is not the expirer events (e.g. when `this.deletePairing()` is called explicitly inside the engine) we DO want to explicitly call `expirer.del`
- To disambiguate these two scenarios handled by the delete helpers, I've added the optional `expirerHasDeleted` param.

## How Has This Been Tested?

- Tested locally against example dapp
- Unit tests

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -352,7 +352,7 @@ export class Engine extends IEngine {
     await this.setExpiry(topic, expiry);
   };
 
-  private deleteSession: EnginePrivate["deleteSession"] = async (topic) => {
+  private deleteSession: EnginePrivate["deleteSession"] = async (topic, expirerHasDeleted) => {
     const { self } = this.client.session.get(topic);
     // Await the unsubscribe first to avoid deleting the symKey too early below.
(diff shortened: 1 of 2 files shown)


xzilja (approved):

bkrem:
CI still choking on these darn `socket hang up` issues, but all fine locally for me on multiple runs 👍 Merging.

PR #1462, opened 2022-09-21
fix(test): ensure generated persistent DBs are always unique

bkrem:
# Description

- Ensure that clients never reuse persistent DB instances between tests
- Also: fix(test): avoids unwanted reassign of beforeClients

## How Has This Been Tested?

- Tested locally via unit tests

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,10 +30,7 @@ zip
 tsconfig.tsbuildinfo
 test.db
 test.db-journal
-client_a.db
-client_a.db-journal
-client_b.db
-client_b.db-journal
+**/tmp
 setup
 result*
 ops/grafana/grafana.ini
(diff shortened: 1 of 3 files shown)


bkrem:
Uhm no `socket hang up` in CIs at all on this one?? Either that was a massive fluke or this is finally resolved, rerunning and praying it's the latter.

ganchoradkov (approved):

bkrem:
@ganchoradkov the `after restart > clients can ping each other` test always seems to fail for me now both on session and pairing, also when I run against the `v2.0` branch (i.e. without these changes) with the new storage deps.

Was that passing for you on your PR locally?

arein (approved):

PR #1494, opened 2022-09-30
chore: socket hang up investigation

ganchoradkov:
# Description
PR to investigate `socket hang up` in GHA. 

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/core/src/core.ts
+++ b/packages/core/src/core.ts
@@ -104,7 +104,7 @@ export class Core extends ICore {
       this.initialized = true;
       this.logger.info(`Core Initilization Success`);
     } catch (error) {
-      this.logger.info(`Core Initilization Failure ${Date.now()}`);
+      this.logger.warn(`Core Initilization Failure at epoch ${Date.now()}`, error);
       this.logger.error((error as any).message);
       throw error;
     }
--- a/packages/core/src/core.ts
+++ b/packages/core/src/core.ts
@@ -104,7 +104,7 @@ export class Core extends ICore {
       this.initialized = true;
       this.logger.info(`Core Initilization Success`);
     } catch (error) {
-      this.logger.info(`Core Initilization Failure`);
+      this.logger.info(`Core Initilization Failure ${Date.now()}`);
       this.logger.error((error as any).message);
       throw error;
     }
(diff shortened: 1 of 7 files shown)


packages/core/src/core.ts
```diff
@@ -104,7 +104,7 @@ export class Core extends ICore {
       this.initialized = true;
       this.logger.info(`Core Initilization Success`);
     } catch (error) {
-      this.logger.info(`Core Initilization Failure`);
+      this.logger.info(`Core Initilization Failure ${Date.now()}`);
```
arein:
```suggestion
      this.logger.warn(`Core Initilization Failure at epoch ${Date.now()}`, error);
```

arein (approved):

packages/sign-client/src/controllers/engine.ts
```diff
@@ -400,8 +400,7 @@ export class Engine extends IEngine {
     const message = await this.client.core.crypto.encode(topic, payload);
     const opts = ENGINE_RPC_OPTS[method].req;
     this.client.history.set(topic, payload);
-    await this.client.core.relayer.publish(topic, message, opts);
-
+    this.client.core.relayer.publish(topic, message, opts);
```
bkrem, line 403: Sorry but only seeing this diff now. How come we don't await the `relayer.publish` step anymore here?

bkrem: `sendResult` and `sendError` still await `relayer.publish` so I'm not clear on whether this is accidental or intentionally different

ganchoradkov: This is the change I discussed regarding the occasional race conditions where awaiting the publish was causing the `response` event to be fired before the listener for it is initialized causing the workflow to hang

bkrem: Ok thanks for context 👍 

Does it make sense to align this for `sendResult` and `sendError` in terms of behaviour or is this unique to `sendRequest`? I think if this is the exception we should probably comment it in the code, since it looks like a mistake at first sight/someone might "fix" it down the line.

ganchoradkov: It's a unique case with `sendRequest` due to the above-mentioned issue happening only in this case. I think adding a comment would be adequate

PR #1516, opened 2022-10-07
chore: a comment to explain why await is ommited

ganchoradkov:
# Description
Added a comment to explain why the `await` was omitted for the publish call to avoid being unintentionally `fixed` in the future

## Due Dilligence

* [ ] Breaking change


packages/sign-client/src/controllers/engine.ts
```diff
@@ -400,6 +400,9 @@ export class Engine extends IEngine {
     const message = await this.client.core.crypto.encode(topic, payload);
     const opts = ENGINE_RPC_OPTS[method].req;
     this.client.history.set(topic, payload);
+
+    // await is intentionally ommited here because of a possible race condition
```
```suggestion
    // await is intentionally omitted here because of a possible race condition
```

bkrem:
Nice one thanks for adding context on this 💯

bkrem (approved):

PR #1531, opened 2022-10-13
fix: ignoring messages handled by external SDKs

ganchoradkov:
Added the possibility to define messages types that would be ignored by the `pairing`. 
This functionality is required as some SDKs such as `Auth` require custom/additional data when processing messages.

# Description

## How Has This Been Tested?
npm run test

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/core/src/controllers/pairing.ts
+++ b/packages/core/src/controllers/pairing.ts
@@ -57,7 +57,7 @@ export class Pairing implements IPairing {
 
   private initialized = false;
   private storagePrefix = CORE_STORAGE_PREFIX;
-  private messageTypesToIgnore = [TYPE_1];
+  private ignoredPayloadTypes = [TYPE_1];
 
   constructor(public core: ICore, public logger: Logger) {
     this.core = core;
--- a/packages/core/src/controllers/pairing.ts
+++ b/packages/core/src/controllers/pairing.ts
@@ -23,6 +23,7 @@ import {
   isValidString,
   isExpired,
   parseExpirerTarget,
+  TYPE_1,
 } from "@walletconnect/utils";
 import {
   formatJsonRpcRequest,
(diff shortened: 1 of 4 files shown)


packages/utils/src/crypto.ts
```diff
@@ -134,17 +134,8 @@ export function validateDecoding(
 }
 
 export function validateEncoding(opts?: CryptoTypes.EncodeOptions): CryptoTypes.EncodingValidation {
-  const type = opts?.type || TYPE_0;
```
bkrem: How come we're removing these checks? I think we should keep them since:

* there could be cases where `validateEncoding` is called without doing the manual message type check beforehand like we are here.
* If it's `TYPE_1` we should still validate that `senderPublicKey/receiverPublicKey` are present

packages/core/src/controllers/pairing.ts
```diff
@@ -56,6 +57,7 @@ export class Pairing implements IPairing {
 
   private initialized = false;
   private storagePrefix = CORE_STORAGE_PREFIX;
+  private messageTypesToIgnore = [TYPE_1];
```
bkrem: To align naming here to not have `messageTypes` and `payloadTypes` concepts side by side:

```suggestion
  private ignoredPayloadTypes = [TYPE_1];
```

packages/utils/src/crypto.ts
ganchoradkov: Good points. Thanks for bringing it up! Thought the validation would be handled by the external SDKs but makes sense to preserve it there

packages/utils/src/crypto.ts
```diff
@@ -143,6 +143,7 @@ export function validateEncoding(opts?: CryptoTypes.EncodeOptions): CryptoTypes.
       throw new Error("missing receiver public key");
     }
   }
+
```
bkrem: Removing accidental newline

```suggestion
```

bkrem (approved):

Issue #1507, opened 2022-10-04
[Feat/v2] Improve client and websocket lifecycle management


bkrem:
## Context

- Prior discussion thread and context from Fireblocks: https://walletconnect.slack.com/archives/C03RME3BS9L/p1663859970679629
- Similar functionality present in Sign v1: https://github.com/WalletConnect/walletconnect-monorepo/pull/480

## Description

- Currently there is no way to explicitly close/pause/reset the websocket connection managed by `core/relayer`
- We also don't currently have a well-defined way to tear down the SignClient and/or Core, e.g. in order to do a full restart without possibly leaving the previous SignClient in an active state and occupying memory.
- Relevant scenarios identified so far:
	- Fireblocks' use case as linked in the Slack discussion above.
	- Tearing down tests/closing socket connections gracefully.
	- Switching relay endpoint on an initialized client (see: https://github.com/WalletConnect/web-examples/issues/65)

PR #1540, opened 2022-10-25
feat: transport lifecycle 

ganchoradkov:
# Description

Resolves # [1507](https://github.com/WalletConnect/walletconnect-monorepo/issues/1507)

Added two new methods in `relayer` to explicitely stop/start WebSocket connection
-  `public abstract transportClose(): Promise<void>;`
-  `public abstract transportOpen(relayUrl?: string): Promise<void>;`

`transportOpen` takes an optional relay URL parameter, else it uses the one provided during the core's initialization

# Notable behavior
To preserve backward compatibility and avoid introducing breaking changes, calling `relayer.provider.request(args)` will restart the WebSocket connection automatically without needing to call `transportOpen()` beforehand 

## How Has This Been Tested?
Tested locally.
Tests can be created to validate relay URL update but that would require making `relayer.relayUrl` property public

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -136,8 +136,8 @@ export class Relayer extends IRelayer {
     await this.provider.connection.close();
   }
 
-  public async transportOpen(relayUrl: string = RELAYER_DEFAULT_RELAY_URL) {
-    this.relayUrl = relayUrl;
+  public async transportOpen(relayUrl?: string) {
+    this.relayUrl = relayUrl || this.relayUrl;
     this.provider = await this.createProvider();
     this.transportExplicitlyClosed = false;
   }
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -53,12 +53,12 @@ export class Relayer extends IRelayer {
   public subscriber: ISubscriber;
   public publisher: IPublisher;
   public name = RELAYER_CONTEXT;
+  public transportExplicitlyClosed = false;
 
   private initialized = false;
 
   private relayUrl: string;
   private projectId: string | undefined;
-  public transportExplicitlyClosed = false;
 
   constructor(opts: RelayerOptions) {
     super(opts);
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -58,6 +58,7 @@ export class Relayer extends IRelayer {
 
   private relayUrl: string;
   private projectId: string | undefined;
+  public transportExplicitlyClosed = false;
 
   constructor(opts: RelayerOptions) {
     super(opts);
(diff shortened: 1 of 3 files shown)


packages/core/src/controllers/relayer.ts
```diff
@@ -58,6 +58,7 @@ export class Relayer extends IRelayer {
 
   private relayUrl: string;
   private projectId: string | undefined;
+  public transportExplicitlyClosed = false;
```
bkrem: Since this is public (does it need to be, i.e. are we expecting people to check this?), can we move it up to live with the other public props?

packages/core/src/controllers/relayer.ts
```diff
@@ -131,9 +131,20 @@ export class Relayer extends IRelayer {
     this.events.removeListener(event, listener);
   }
 
+  public async transportClose() {
+    this.transportExplicitlyClosed = true;
+    await this.provider.connection.close();
+  }
+
+  public async transportOpen(relayUrl: string = RELAYER_DEFAULT_RELAY_URL) {
```
bkrem: I think the `relayUrl` param shouldn’t default to our constant here but should be undefined by default, because of this scenario:

1. Create an initial client with custom relayUrlA
2. Close transport
3. Reopen but don’t provide a relayUrl
4. -> transportOpen quietly overwrites our relayUrlA value with the default relayUrl

bkrem: I.e. only assign to default relayUrl if `this.relayUrl` is not already set

packages/core/src/controllers/relayer.ts
ganchoradkov: It's a flag to differentiate why the connection might be down.
Checking `relayer.connected()` is not sufficient as it might be down due to several reasons such as
- socket died/invalid relay url/other network issue
- closed via `transportClose`

bkrem: Yeah makes sense :) 👍

packages/core/src/controllers/relayer.ts
ganchoradkov: Good catch, thank you!

bkrem:
Super weird that the CI is taking 10+ minutes again now but it was around 5 mins yesterday 🤔 I ran both the `v2.0` branch and this one locally for tests and they weren't any slower than before. Maybe to do with which region we hit for CI runs.

bkrem (approved):
Great stuff, thanks for tackling this 🙌 

I think we should keep #1507 open for now though, since part of the scope there is also about a general way to tear down the client (i.e. disconnect, remove listeners, free up allocated memory etc),  which Fireblocks also specifically asked about.

ganchoradkov:
> Super weird that the CI is taking 10+ minutes again now but it was around 5 mins yesterday thinking I ran both the `v2.0` branch and this one locally for tests and they weren't any slower than before. Maybe to do with which region we hit for CI runs.

there seems to be an issue with the https://us-east-1.relay.walletconnect.com/ which is the endpoint the GHA is using. Might be a good idea to delay the merge until the issue is fixed

PR #1563, opened 2022-10-29
linting errors, async issues async not needed or await missing

bws9000:
# Description
Screen Shot 2022-10-29 at 5 13 13 PM

I was just interested in your project. I ran  `npm run check` and noticed some linting warnings probably overlooked while refactoring no be deal

Universal provider
- added await to line 53. This await probably not needed or async
- added await to line 100. It appears you might need one there
- line 139 removed async, no need for it there it looks like

eip155
- added await to line 51, i'm assuming you want one there

removed linting warnings

Resolves:
No related issue


ganchoradkov (approved):
Thanks for the PR, LGTM

PR #1590, opened 2022-11-03
fix: socket transport manual start/stop

### LINKED ISSUE #1578: SignClient not handling JSON-RPC responses after `transportClose/transportOpen`
**Describe the bug**
When explicitly disconnecting and reconnecting the SignClient with the new `core.relayer.transportOpen/transportClose` methods, the client seems to be able to send requests, but does not seem to register incoming responses.

**SDK Version (if relevant)**
 - sign-client@2.1.0

**To Reproduce**

Commit for the modified example dapp is here: https://github.com/WalletConnect/web-examples/commit/3832c78639a5f198e4073f4f88ee2bbcf7bcd021

1. Open the modified example dapp deployment: https://react-dapp-v2-git-debug-transportclose-walletconnect1.vercel.app/
2. Open https://react-wallet.walletconnect.com/ in another tab
3. Create a new session (this will trigger `transportClose` and then `transportOpen` - see commit above)
4. Click the Ping button -> wallet replies but dapp does not register response
5. OR: do a sign request -> approve in wallet -> response is not registered by dapp.

ganchoradkov:
# Description
With the addition of the `transportClose` & `transportOpen` a bug was introduced where after the socket was restarted, incoming relay messages were ignored.
This was due to the `core/subscriber` clearing its in memory cache/data while the socket was being closed and this caused the sequential `subscription topic` checks to fail

This PR adds a restart mechanism where when the socket is reconnected & the subscriber will resubscribe to the topics in its storage.
Also added tests to simulate transport open/close usage  

Resolves https://github.com/WalletConnect/walletconnect-monorepo/issues/1578

## How Has This Been Tested?
npm run test

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/core/src/controllers/subscriber.ts
+++ b/packages/core/src/controllers/subscriber.ts
@@ -20,11 +20,11 @@ import {
 } from "@walletconnect/utils";
 import {
   CORE_STORAGE_PREFIX,
-  RELAYER_PROVIDER_EVENTS,
   SUBSCRIBER_CONTEXT,
   SUBSCRIBER_EVENTS,
   SUBSCRIBER_STORAGE_VERSION,
   PENDING_SUB_RESOLUTION_TIMEOUT,
+  RELAYER_EVENTS,
 } from "../constants";
 import { SubscriberTopicMap } from "./topicmap";
 
@@ -51,8 +51,7 @@ export class Subscriber extends ISubscriber {
   public init: ISubscriber["init"] = async () => {
     if (!this.initialized) {
       this.logger.trace(`Initialized`);
-      await this.restore();
-      await this.reset();
+      await this.restart();
       this.registerEventListeners();
       this.onEnable();
     }
@@ -293,6 +292,11 @@ export class Subscriber extends ISubscriber {
     } as SubscriberEvents.Deleted);
   }
 
+  private restart = async () => {
+    await this.restore();
+    await this.reset();
+  };
+
   private async persist() {
     await this.setRelayerSubscriptions(this.values);
     this.events.emit(SUBSCRIBER_EVENTS.sync);
@@ -335,8 +339,7 @@ export class Subscriber extends ISubscriber {
   }
 
   private async onConnect() {
-    if (this.relayer.transportExplicitlyClosed) return;
-    await this.reset();
+    await this.restart();
     this.onEnable();
   }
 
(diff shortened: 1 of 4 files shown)


packages/sign-client/test/sdk/client.spec.ts
```diff
@@ -287,4 +288,68 @@ describe("Sign Client Integration", () => {
       await deleteClients(clients);
     }, 50_000);
   });
+  describe("transport", () => {
```
bkrem, line 291: Nice 👌 just ran these agains the current `v2.0` and they failed as expected, so the fact that it's passing here is the validation we need.

packages/core/src/controllers/subscriber.ts
```diff
@@ -357,10 +361,10 @@ export class Subscriber extends ISubscriber {
     this.relayer.core.heartbeat.on(HEARTBEAT_EVENTS.pulse, () => {
       this.checkPending();
     });
-    this.relayer.provider.on(RELAYER_PROVIDER_EVENTS.connect, async () => {
```
bkrem: Why are we moving the listener binding from `provider` to `relayer` directly?

bkrem (approved):
LGTM 🚀 

Can we do a quick canary for this to play around with it in a real world context before we merge and ship a patch?

packages/core/src/controllers/subscriber.ts
ganchoradkov: I moved them initially because the `relayer` fires exactly these events too so it reduced a bit of overhead and gives the relayer more control over the events but just tested and the fix of the PR doesn't need that change so I reverted to the original usage

ganchoradkov:
> LGTM rocket
> 
> Can we do a quick canary for this to play around with it in a real world context before we merge and ship a patch?

@bkrem definitely. 
Just published `core` & `sign-client` canary with version `2.1.0-433abac0-1`

I'm also going to dogfood them with some of the example apps 👍

bkrem:
> @bkrem definitely. Just published `core` & `sign-client` canary with version `2.1.0-433abac0-1`
> 
> I'm also going to dogfood them with some of the example apps 👍

Nice thank you! Will do some testing shortly too and report back here 👌

bkrem:
Ok played around with that canary on the example dapp + wallet, and couldn't find any problems using the repro I linked to in the issue for this 👍 

* Signing request-response E2e after `transportClose` -> `transportOpen` ✅ 
* Session ping E2E after close -> open ✅ 

All good from my side, thanks for fixing 🙌

PR #1598, opened 2022-11-04
fix: universal provider persistance

ganchoradkov:
# Description
While using the `universal-provider` in our example apps I discovered issues with storage persistence where on page reload the `universal-provider` couldn't restore completely its state

This update includes

- bug fix to the cache reloading
- addition constructor parameters such as  `storageOptions` so persisted storage is available 
- tests to cover persisted storage usage

## How Has This Been Tested?
npm run test

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/.github/workflows/pr_checks.yml
+++ b/.github/workflows/pr_checks.yml
@@ -33,6 +33,7 @@ jobs:
             packages/utils/dist
             providers/ethereum-provider/dist
             providers/signer-connection/dist
+            providers/universal-provider/dist
 
   code_style:
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


bkrem:
> While using the universal-provider in our example apps I discovered issues with storage persistence where on page reload the universal-provider couldn't restore completely its state

Dogfooding to the rescue, glad we uncovered these in that case 👍

bkrem:
Realised that we're not running the suite for the `universal-provider` pkg on CI, should probably configure it so we don't have false confidence from CI passing: https://github.com/WalletConnect/walletconnect-monorepo/blob/fix/universal-provider-persistance/.github/workflows/pr_checks.yml

bkrem (approved):
LGTM 👍 

Just need to make sure we run tests for this package on CI too as mentioned

PR #1599, opened 2022-11-04
feat(test): cover reconnection lifecycle

arein:
# Description

Tests that client reconnects automatically after the WS is being closed.

Towards https://github.com/WalletConnect/rs-relay/issues/587

## How Has This Been Tested?

Ran locally against staging.

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/sign-client/test/lifecycle/lifecycle.spec.ts
+++ b/packages/sign-client/test/lifecycle/lifecycle.spec.ts
@@ -110,6 +110,6 @@ describe("Lifecycle", () => {
     const { result } = done.meta;
     const nowTimestamp = Date.now();
     const latencyMs = nowTimestamp - (result?.startTime || nowTimestamp);
-    log(`Canary finished in state ${result?.state} took ${latencyMs}ms`);
+    log(`Lifecycle test finished in state ${result?.state} took ${latencyMs}ms`);
   });
 });
--- a/packages/sign-client/test/lifecycle/lifecycle.spec.ts
+++ b/packages/sign-client/test/lifecycle/lifecycle.spec.ts
@@ -1,0 +1,115 @@
+import { getSdkError } from "@walletconnect/utils";
+import {
+  initTwoClients,
+  testConnectMethod,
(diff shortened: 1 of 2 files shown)


packages/sign-client/test/lifecycle/lifecycle.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+  uploadCanaryResultsToCloudWatch,
+  throttle,
+  publishToStatusPage,
+} from "../shared";
+import { TEST_RELAY_URL } from "../shared/values";
+import { describe, it, expect, afterEach } from "vitest";
+
+const environment = process.env.ENVIRONMENT || "dev";
+
+const timeout = environment === "prod" ? 610_000 : 70_000;
```
bkrem, line 15: Think it makes sense to break these magic numbers out into their own labelled `const`, also since there's hardcoded references below that should vary by env, i.e.:

```ts
const PROD_TIMEOUT_IN_MS = 610_000
const STAGING_TIMEOUT_IN_MS = 70_000
```

packages/sign-client/test/lifecycle/lifecycle.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+            resolve();
+          });
+
+          await clients.A.ping({ topic: sessionA.topic });
+        } catch (e) {
+          reject(e);
+        }
+      });
+
+      log(`Going to wait for 70s until pinging again`);
```
bkrem:
```suggestion
      log(`Going to wait for ${timeout / 1000}s until pinging again`);
```

packages/sign-client/test/lifecycle/lifecycle.spec.ts
bkrem: Actually disregard, not that necessary anymore, found a simpler way to deal with the timeout reference further down 👍

bkrem (approved):
LGTM 🚀

PR #1643, opened 2022-11-15
fix: resubscribing after transport is reopened

ganchoradkov:
PR to address the possible race condition where `transportOpen` resolves before the client is in a ready state to receive messages. This happens because the client needs to resubscribe to its cached topics. 
Any messages sent to a topic the client has not yet resubscribed will not be received

# Description

## How Has This Been Tested?

- `npm run test`
- transport tests

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -142,9 +142,9 @@ export class Relayer extends IRelayer {
     this.relayUrl = relayUrl || this.relayUrl;
     this.transportExplicitlyClosed = false;
     await this.provider.connect();
-    // wait the subscriber to finish resubscribing to its topics
+    // wait for the subscriber to finish resubscribing to its topics
     await new Promise<void>((resolve) => {
-      this.subscriber.on(SUBSCRIBER_EVENTS.resubscribed, () => {
+      this.subscriber.once(SUBSCRIBER_EVENTS.resubscribed, () => {
         resolve();
       });
     });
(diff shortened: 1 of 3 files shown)


packages/core/src/controllers/relayer.ts
```diff
@@ -141,6 +142,12 @@ export class Relayer extends IRelayer {
     this.relayUrl = relayUrl || this.relayUrl;
     this.transportExplicitlyClosed = false;
     await this.provider.connect();
+    // wait the subscriber to finish resubscribing to its topics
```
arein:
```suggestion
    // wait for the subscriber to finish resubscribing to its topics
```

arein (approved):

packages/core/src/controllers/relayer.ts
```diff
@@ -141,6 +142,12 @@ export class Relayer extends IRelayer {
...
+    await new Promise<void>((resolve) => {
+      this.subscriber.on(SUBSCRIBER_EVENTS.resubscribed, () => {
```
bkrem: I think this needs to be `.once`, since the listener should only fire once right. Otherwise we risk leaking memory by adding listeners but not removing them each time `transportOpen` is called

packages/core/src/controllers/relayer.ts
```suggestion
    // wait for the subscriber to finish resubscribing to its topics
```

packages/core/src/controllers/relayer.ts
ganchoradkov: Good point! 🙏

bkrem (approved):

PR #1758, opened 2022-12-14
fix: stalled connection in CI

ganchoradkov:
# Description
Running `sign-client` tests in GHA has been problematic for a while due to socket connections issues such as

- `socket hang up` <- inability to capture the error causing the CI run to fail
- `stalled connections` <- A connection that is seemingly `active` in the SDK but on the relay, it has been closed with `104 connection reset by peer` 

Stalled connections/104 errors are happening on the client machine and are raised by the OS when it doesn't recognize an incoming packet. This can happen by any firewall/router etc and since we have no control over GHA runners a solutions is needed to handle such scenarios

Worth noting that the issues are so far  **only** reproducible during CI runs and not during normal SDK usage or local runs

### 1. Reducing overhead
Added `/integration` folder where full e2e test coverage is available without unnecessary duplicate tests and/or redundant pairings. We have 5+ `ping` tests that are testing the same thing. Moved the validation tests out of the integration pipeline as they don't test the relay, just create overhead connections
### 2. Better catching connection errors
Updated the `@walletconnect/jsonrpc-ws-connection` package to handle connection errors, also implemented reconnection on `socket hang up` so we can continue processing instead of failing the run
### 3. Proper disposal of connections after a test is done
Proper closing of connections and removing event listeners/heartbeats after a test is done 
### 4. Handling stalled connections
Added expiry timeout on subscrptions/payloads requests where if the request expires a reconnection is attempted. From the test runs, I'm seeing that a reconnection might attempted several times before it is restored and active again.
The mechanism for stalled requests works by utilizing the `queue` where the request is removed from the queue only after successful delivery

## How Has This Been Tested?
npm run test
tons of CI runs
canary has been running this code for ~2 weeks 

## Due Dilligence

* [ ] Breaking change


packages/core/src/controllers/relayer.ts
```diff
@@ -135,20 +148,50 @@ export class Relayer extends IRelayer {
...
+          });
+        });
+      }
+    } catch (e: unknown | Error) {
+      const error = e as Error;
+      if (!/socket hang up/i.test(error.message)) {
+        throw new Error(error.message);
+      }
+      // eslint-disable-next-line no-console
+      console.error(error);
```
xzilja: Should this be a log instead?

packages/core/src/controllers/relayer.ts
```diff
@@ -221,15 +264,20 @@ export class Relayer extends IRelayer {
     this.provider.on(RELAYER_PROVIDER_EVENTS.error, (err: unknown) =>
       this.events.emit(RELAYER_EVENTS.error, err),
     );
+
+    this.events.on(RELAYER_EVENTS.connection_stalled, async () => {
```
xzilja, line 268: Can probably simplify to

```
this.events.on(RELAYER_EVENTS.connection_stalled, this.restartTransport)
```

packages/core/src/controllers/subscriber.ts
```diff
@@ -40,7 +41,7 @@ export class Subscriber extends ISubscriber {
   private pendingSubscriptionWatchLabel = "pending_sub_watch_label";
   private pendingSubInterval = 20;
   private storagePrefix = CORE_STORAGE_PREFIX;
-
+  private subscribeTimeout = 10_000;
```
xzilja, line 44: This looks similar to `private publishTimeout = 10_000;` I saw previously. If they are aimed to be the same, perhaps it would be better to move them into a shared const file?

packages/core/src/controllers/subscriber.ts
```diff
@@ -362,10 +374,10 @@ export class Subscriber extends ISubscriber {
     this.relayer.core.heartbeat.on(HEARTBEAT_EVENTS.pulse, () => {
       this.checkPending();
     });
-    this.relayer.provider.on(RELAYER_PROVIDER_EVENTS.connect, async () => {
+    this.relayer.on(RELAYER_EVENTS.connect, async () => {
```
xzilja, line 377:
```
this.relayer.on(RELAYER_EVENTS.connect, this.onConnect)
```

packages/core/src/controllers/subscriber.ts
```diff
@@ -362,10 +374,10 @@ export class Subscriber extends ISubscriber {
...
       await this.onConnect();
     });
-    this.relayer.provider.on(RELAYER_PROVIDER_EVENTS.disconnect, () => {
+    this.relayer.on(RELAYER_EVENTS.disconnect, () => {
```
xzilja, line 380: this.relayer.on(RELAYER_EVENTS.disconnect, this.onDisconnect)

packages/sign-client/test/sdk/client.spec.ts
```diff
@@ -264,18 +145,17 @@ describe("Sign Client Integration", () => {
       const result = clients.A.session.get(topic).namespaces;
       expect(result).to.eql(namespacesAfter);
       await deleteClients(clients);
-    }, 50_000);
+    });
   });
 
   describe("extend", () => {
-    it("updates session expiry state", async () => {
+    it.skip("updates session expiry state", async () => {
```
xzilja: Double checking, should this remain skiped?

packages/sign-client/test/sdk/integration/integration.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+      try {
+        const settled: any = await createExpiringPromise(
+          testConnectMethod(clients),
+          TESTS_CONNECT_TIMEOUT,
+        );
+        pairingA = settled.pairingA;
+        sessionA = settled.sessionA;
+      } catch (e) {
+        // eslint-disable-next-line no-console
+        console.log("retrying", e);
```
xzilja: Perhaps `log` util here?

xzilja (commented):
Overall looking great! Grasped what you did here and it was definitely a needed addition. I left few comments on cleanup ideas.

packages/core/src/controllers/relayer.ts
```diff
@@ -135,20 +148,50 @@ export class Relayer extends IRelayer {
...
+          this.once(RELAYER_EVENTS.transport_closed, () => {
+            reject();
+          }),
+        ),
+      ]);
+
+      if (this.initialized) {
+        // wait for the subscriber to finish resubscribing to its topics
+        await new Promise<void>((resolve) => {
+          this.subscriber.once(SUBSCRIBER_EVENTS.resubscribed, () => {
```
bkrem, line 164: Is there a possibility of `SUBSCRIBER_EVENTS.resubscribed` firing before the listener is added here, i.e. causing the flow to hang?

packages/core/src/controllers/relayer.ts
```diff
@@ -135,20 +148,50 @@ export class Relayer extends IRelayer {
...
+      }
+      // eslint-disable-next-line no-console
+      console.error(error);
+      this.events.emit(RELAYER_EVENTS.transport_closed);
+    }
   }
+
+  public async restartTransport(relayUrl?: string) {
+    await this.transportClose();
+    await new Promise<void>((resolve) => setTimeout(resolve, RELAYER_RECONNECT_TIMEOUT));
```
bkrem, line 191: Can you explain this more please? Seems like it's effectively just doing a blocking `wait` for the duration of `RELAYER_RECONNECT_TIMEOUT`

packages/core/src/controllers/relayer.ts
ganchoradkov: I haven't seen that happening but as a precaution, I can move the event before the `.connect()` is called to be sure

packages/core/src/controllers/subscriber.ts
bkrem: Sorry but disagree, if these are two distinct variables unrelated to each other they shouldn't be unified because they happen to have the same value.

packages/core/src/controllers/relayer.ts
ganchoradkov: My reasoning was to give leeway for any replication delay in the relay and avoid the following scenario:
- initially connected to Node#1 
- disconnect transport
- reconnect transport immediately
- connected to node#2 before node#1 is done cleanup/replication etc

packages/sign-client/package.json
```diff
@@ -27,7 +27,8 @@
     "test:concurrency": "vitest run --dir test/concurrency",
     "test:xregion": "vitest run --dir test/xregion",
     "test": "npm run test:pre; npm run test:run",
-    "test:ignoreUnhandled": "npm run test:pre; npm run test:run -- --dangerouslyIgnoreUnhandledErrors --no-threads",
+    "test:integration": "vitest run --dir test/sdk/integration",
+    "test:ignoreUnhandled": "npm run test:pre; npm run test:integration -- --dangerouslyIgnoreUnhandledErrors --no-threads",
```
bkrem, line 31: Is this still being used btw? This was added as a workaround a while back and the unhandled error issue was resolved since then I think.

ganchoradkov: the `type bool of undefined` or something coming from some inner library still shows up occasionally

packages/core/src/controllers/relayer.ts
bkrem: Got it, thanks for context 👍 

I also just realised that `RELAYER_RECONNECT_TIMEOUT` is that standard one second we've always had for auto-reconnection attempts, so not as dramatic as it appeared at first.

packages/sign-client/test/sdk/integration/integration.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+        namespaces: namespacesAfter,
+      });
+      await acknowledged();
+      const result = clients.A.session.get(topic).namespaces;
+      expect(result).to.eql(namespacesAfter);
+    });
+  });
+
+  describe("extend", () => {
+    it.skip("updates session expiry state", async () => {
```
bkrem: Skipping here but I'm guessing this should be re-enabled if it was specifically moved to `integration.spec.ts`

packages/sign-client/test/sdk/persistance.spec.ts
```diff
@@ -0,0 +1,128 @@
+import { generateRandomBytes32 } from "@walletconnect/utils";
```
bkrem: Sorry to be that guy, but better we fix it up front: Should be `persistence.spec.ts` (all `e`) 😬

packages/sign-client/test/sdk/persistance.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+import { describe, it } from "vitest";
+import { initTwoClients, testConnectMethod, deleteClients, throttle } from "../shared";
+
+const generateClientDbName = (prefix: string) =>
+  `./test/tmp/${prefix}_${generateRandomBytes32()}.db`;
+
+describe("Sign Client Persistance", () => {
```
bkrem:
```suggestion
describe("Sign Client Persistence", () => {
```

packages/core/src/controllers/publisher.ts
```diff
@@ -39,7 +41,17 @@ export class Publisher extends IPublisher {
...
-      await this.rpcPublish(topic, message, ttl, relay, prompt, tag);
+      try {
+        const publish = await createExpiringPromise(
+          this.rpcPublish(topic, message, ttl, relay, prompt, tag),
+          this.publishTimeout,
+        );
+        await publish;
+      } catch (err) {
+        this.logger.debug(`Publishing Payload stalled`);
+        this.relayer.events.emit(RELAYER_EVENTS.connection_stalled);
```
arein, line 52: do you need to inspect the error here to tell for sure it's the `connection_stalled` one?

packages/sign-client/test/lifecycle/lifecycle.spec.ts
```diff
@@ -12,7 +12,7 @@ import { describe, it, expect, afterEach } from "vitest";
 
 const environment = process.env.ENVIRONMENT || "dev";
 
-const timeout = environment === "prod" ? 610_000 : 70_000;
+const timeout = environment === "prod" ? 610_000 : 30_000;
```
arein, line 15: feel free to link this against the server config https://github.com/WalletConnect/rs-relay/blob/master/terraform/vars/staging.tfvars#L9-L11

arein (approved):

packages/core/src/controllers/publisher.ts
ganchoradkov: This event is emitted when the `publish` request fails to resolve within the expiry (currently 10s). In normal circumstances, the expiry should be more than enough for a request to complete. If it doesn't, it is safe to assume the connection stalled. Also, the error here comes from the expiry rejecting the promise, not an actual network error

packages/core/src/controllers/relayer.ts
ganchoradkov: Using that approach leads to undefined errors unless this.restartTransport is used as arrow fx which looks weird as it breaks the standard for other fx
e.g. 
```
private restartTransport = () => {}
```

will prolly refactor at a later stage

bkrem (approved):

PR #1771, opened 2022-12-15
feat: Web3Wallet SDK

ganchoradkov:
# Description
`Web3Wallet` SDK to simplify `Sign` & `Auth` usage.
Interfaces: https://gist.github.com/pedrouid/4067a262550ffb80ea93ab21cedba2fb

## How Has This Been Tested?
integration tests for each method

## Due Dilligence

* [ ] Breaking change


packages/web3wallet/src/client.ts
```diff
@@ -0,0 +1,183 @@
...
+  public removeListener: IWeb3Wallet["removeListener"] = (name, listener) => {
+    return this.events.removeListener(name, listener);
+  };
+
+  // ---------- Engine ----------------------------------------------- //
+
+  public approveSession: IWeb3Wallet["approveSession"] = async (params) => {
+    try {
+      return await this.engine.approveSession(params);
+    } catch (error: any) {
```
xzilja, line 55: Given that we are catching and logging all errors in sub-sdks like sign and auth, do we need to re-do this here?

packages/web3wallet/src/controllers/engine.ts
```diff
@@ -0,0 +1,119 @@
...
+    super(client);
+    // initilized in init()
+    this.signClient = {} as any;
+    this.authClient = {} as any;
+  }
+
+  public init = async () => {
+    this.signClient = await SignClient.init({
+      core: this.client.core,
+      projectId: "",
```
xzilja: I couldn't find where project id is set for these latter on, should this come from some sort of options as opposed to being a string?

xzilja (commented):
This is shaping up really nicely 👍 Left few comments and one overall here: Do we need to manually re-declare some of the types inside new client and engine type files here? Or is it possible to re-assign these to already existing types from sign and auth sdks, thus keeping the in one place when we need to change them.

Similarly, some logic i.e catching errors and logging them is already done in those sdk's, so perhaps we don't need to repeat it here?

packages/web3wallet/src/controllers/engine.ts
ganchoradkov: `projectId` is set in the `core` instance that is passed in the `Web3Wallet` like so
https://github.com/WalletConnect/walletconnect-monorepo/blob/feat/web3wallet/packages/web3wallet/test/sign.spec.ts#L32

packages/web3wallet/src/client.ts
ganchoradkov: technically the SDK doesn't know that/if errors are being caught at an inner level so I don't think there is harm be on the safe side

packages/web3wallet/src/client.ts
```diff
@@ -0,0 +1,183 @@
...
+  static async init(opts: Web3WalletTypes.Options) {
+    const client = new Web3Wallet(opts);
+    await client.initialize();
+
+    return client;
+  }
+
+  constructor(opts: Web3WalletTypes.Options) {
+    super(opts);
+    this.metadata = opts.metadata as any; // TODO: automatic fetch of metadata when?
```
bkrem: Why the cast to `any` here?

Can probs remove the todo that is actually a question too 😄

packages/web3wallet/src/controllers/engine.ts
bkrem: Can't we simply omit `projectId` if it's already present via core? i.e. [SignClient.Options](https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/types/src/sign-client/client.ts#L46)

packages/web3wallet/src/controllers/engine.ts
```diff
@@ -0,0 +1,119 @@
...
+      return sessions;
+    }, {});
+  };
+
+  public getPendingSessionProposals: IWeb3WalletEngine["getPendingSessionProposals"] = () => {
+    return this.signClient.proposal.getAll();
+  };
+
+  public getPendingSessionRequests: IWeb3WalletEngine["getPendingSessionRequests"] = () => {
+    return this.client.pendingRequest.getAll();
```
bkrem, line 80: Not an item for now, but I think it would be interesting to discuss if this (`pendingRequest` store) should be a proper part of SignClient, rather than layering it on top here, so we have first class support for pending session requests in the client itself.

cc @pedrouid @0xAsimetriq

bkrem (approved):
Great work on this Gancho 💯 🚀 

My only real concern is how the treat `pendingRequests` in the mid-long term, since we seem to be layering sign-specific persistence/logic on top of the actual sign client here.

I think it's fine for now and we should be able to easily switch over to a signClient-level implementation if/when that ships, but definitely wondering if it makes more sense to make that a first-class citizen in SignClient itself.

packages/web3wallet/src/controllers/engine.ts
ganchoradkov: removing it in `sign` but it's required param in `auth` :+1:

packages/web3wallet/src/client.ts
ganchoradkov: making it required as it's required in `Auth`

PR #1795, opened 2022-12-20
feat: pending requests

### LINKED ISSUE #1793: pending requests
**Is your feature request related to a problem? Please describe.**
Currently `sign-client` lacks pending requests store. If a new session request is failed to be acted on during event emitting, the request is lost

**Describe the solution you'd like**
`pendingRequest` store

ganchoradkov:
# Description
Pending Request store implementation where new session requests are stored & can be retrieved at a later stage

Also, updated the `Web3Wallet` to use the pending requests from `sign-client`

Resolves https://github.com/WalletConnect/walletconnect-monorepo/issues/1793

## How Has This Been Tested?
Locally + additional integration test

## Due Dilligence

* [ ] Breaking change


packages/sign-client/src/controllers/pendingRequest.ts
```diff
@@ -0,0 +1,10 @@
+import { Store } from "@walletconnect/core";
+import { Logger } from "@walletconnect/logger";
+import { ICore, PendingRequestTypes } from "@walletconnect/types";
+import { REQUEST_CONTEXT, SIGN_CLIENT_STORAGE_PREFIX } from "../constants";
+
+export class PendingRequest extends Store<number, PendingRequestTypes.Struct> {
```
bkrem, line 6: Since we both already know this is a gotcha: are we sure that the `PendingRequest` store rehydrates from persistence/`localStorage` as expected without the `getKey` specification?

Should be fine since it has a `topic` field but worth making sure.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -674,6 +675,7 @@ export class Engine extends IEngine {
     const { id, params } = payload;
     try {
       this.isValidRequest({ topic, ...params });
+      await this.client.pendingRequest.set(payload.id, { id, topic, params });
```
bkrem: Lifecycle: Is there a risk of a bunch of pendingRequests being set but never cleaned up if they're never responded to? Could be used to create bloat either accidentally or maliciously.

packages/web3wallet/src/types/engine.ts
```diff
@@ -61,7 +66,7 @@ export abstract class IWeb3WalletEngine {
   public abstract getPendingSessionProposals(): Record<number, ProposalTypes.Struct>;
 
   // query all pending session requests (SIGN)
-  public abstract getPendingSessionRequests(): Record<number, Web3WalletTypes.SessionRequest>;
+  public abstract getPendingSessionRequests(): PendingRequestTypes.Struct[];
```
bkrem, line 69: If we're adding to SignClient we should also expose `getPendingSessionRequests` on the SignClient itself, rather than only here.

bkrem: At least that's how I understood the decision/intention of this during the Sync call, making it simply a wrapper/proxy call here. Let me know if you're seeing it differently please 🙏

ganchoradkov: makes sense, will update accordingly :+1:

packages/sign-client/src/controllers/engine.ts
ganchoradkov: Might be worth adding some kind of clean-up fx on start where expired requests are removed if not responded to just to be sure

packages/sign-client/src/controllers/pendingRequest.ts
ganchoradkov: wrote a test to confirm it works, so ye, no issues. Will add it to the persistence suite

packages/sign-client/src/controllers/engine.ts
bkrem: Hmm can we handle the same way we handle proposals, i.e. by setting it in the expirer which will then take care of it (once we make sure it's handled on `EXPIRER_EVENTS.expired`):
https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/sign-client/src/controllers/engine.ts#L344

It's the same scenario but it's a pending request instead of a pending session settlement.

packages/sign-client/src/controllers/pendingRequest.ts
bkrem: 🤘

packages/sign-client/src/controllers/engine.ts
ganchoradkov: added to the expirer flow. 
Question: If we are to add more stuff to the expirer might be worth adding new field `type` so the filtering in each listener  would be more clear/easier. Wdyt?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -350,6 +362,19 @@ export class Engine extends IEngine {
...
+    pendingRequest: PendingRequestTypes.Struct,
+  ) => {
+    const expiry = ENGINE_RPC_OPTS.wc_sessionRequest.req.ttl;
+    const { id, topic, params } = pendingRequest;
+    await this.client.pendingRequest.set(id, {
+      id,
+      topic,
+      params,
+    });
+    if (expiry) this.client.core.expirer.set(id, expiry);
```
bkrem, line 375: Do we need the `if` guard here if `expiry` is just referencing the static value for `wc_sessionRequest.req.ttl`, which is presumably always defined?

packages/sign-client/src/controllers/engine.ts
bkrem: Nice thank you 💯 

> Question: If we are to add more stuff to the expirer might be worth adding new field type so the filtering in each listener would be more clear/easier. Wdyt?

Yeah definitely, it was already overly implicit (deciding if its a session or proposal based on whether it has an id or topic). Great candidate for some refactoring since it's internal only 👍

bkrem (approved):
Thanks for the rapid turnaround incorporating this into SignClient directly 🙏

packages/sign-client/src/controllers/engine.ts
ganchoradkov: the `ttl` is an optional properly on type `RelayerTypes.PublishOptions.ttl?: number | undefined` so the `if` is to satisfy typescript. Its that or `const expiry = ENGINE_RPC_OPTS.wc_sessionRequest.req.ttl || 0`

bkrem: Ok I assumed it must be something like that, thanks for context 👍

PR #1832, opened 2023-01-04
feat: empty required namespaces

ganchoradkov:
# Description
Empty `requiredNamespaces` as per spec https://github.com/WalletConnect/walletconnect-docs/blob/a14e99fe3e42447ce892da34cc32b57b0e289bb8/docs/specs/clients/sign/session-namespaces.md#110-proposal-namespaces-may-be-empty

When a dapp does not provide `requiredNamespaces`, it will use the namespaces provided by the wallet when session is initiated  

The PR includes empty `requiredNamespaces` handling + test

- Canary `2.1.5-rc-135e5e70`
- Dapp example https://react-dapp-v2-4k4jtajfd-walletconnect1.vercel.app/
- Wallet example https://react-wallet-v2-oyxqd89lh-walletconnect1.vercel.app/

## How Has This Been Tested?
npm run test

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -868,11 +868,7 @@ export class Engine extends IEngine {
     if (!isUndefined(pairingTopic)) await this.isValidPairingTopic(pairingTopic);
 
     // validate required namespaces only if they are defined
-    if (
-      typeof requiredNamespaces === "object" &&
-      !Array.isArray(requiredNamespaces) &&
-      !Object.keys(requiredNamespaces).length
-    ) {
+    if (isValidObject(requiredNamespaces) === 0) {
       return;
     }
 
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -868,7 +868,7 @@ export class Engine extends IEngine {
     if (!isUndefined(pairingTopic)) await this.isValidPairingTopic(pairingTopic);
 
     // validate required namespaces only if they are defined
-    if (isValidObject(requiredNamespaces) === 0) {
+    if (!isUndefined(requiredNamespaces) && isValidObject(requiredNamespaces) === 0) {
       return;
     }
 
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -51,6 +51,7 @@ import {
   TYPE_1,
   getRequiredNamespacesFromNamespaces,
   isValidObject,
+  isValidArray,
 } from "@walletconnect/utils";
 import { SESSION_EXPIRY, ENGINE_CONTEXT, ENGINE_RPC_OPTS } from "../constants";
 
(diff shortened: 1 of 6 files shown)


packages/sign-client/src/controllers/engine.ts
```diff
@@ -148,7 +149,15 @@ export class Engine extends IEngine {
     this.isInitialized();
     await this.isValidApprove(params);
     const { id, relayProtocol, namespaces } = params;
-    const { pairingTopic, proposer, requiredNamespaces } = this.client.proposal.get(id);
+    const proposal = this.client.proposal.get(id);
+    let { pairingTopic, proposer, requiredNamespaces } = proposal;
+
+    if (!isValidObject(requiredNamespaces)) {
+      const required = getRequiredNamespacesFromNamespaces(namespaces, "approve()");
```
xzilja, line 156: it seems `getRequiredNamespacesFromNamespaces` is used only once, can `approve()` bit be hard coded into the method itself, or are you expecting this function to be used in different callers besides approve?

xzilja (approved):
Overall looks good 👍 left one small comment on possible simplification.

packages/sign-client/src/controllers/engine.ts
ganchoradkov: Technically it can be hardcoded but set it dynamic as a precaution when/if the fx is used elsewhere as is a useful utility especially now that we're supporting empty namespaces

packages/sign-client/src/controllers/engine.ts
```diff
@@ -856,6 +866,16 @@ export class Engine extends IEngine {
     }
     const { pairingTopic, requiredNamespaces, relays } = params;
     if (!isUndefined(pairingTopic)) await this.isValidPairingTopic(pairingTopic);
+
+    // validate required namespaces only if they are defined
+    if (
```
bkrem: Maybe I'm not seeing something, but couldn't we use `!isValidObject()` as above rather than inlining ad-hoc conditions here?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -856,6 +866,12 @@ export class Engine extends IEngine {
     }
     const { pairingTopic, requiredNamespaces, relays } = params;
     if (!isUndefined(pairingTopic)) await this.isValidPairingTopic(pairingTopic);
+
+    // validate required namespaces only if they are defined
+    if (!isUndefined(requiredNamespaces) && isValidObject(requiredNamespaces) === 0) {
```
bkrem, line 871: Sorry but I'm confused by the `=== 0` on what should be a boolean return. Can we fix `isValidObject` if it's returning a number? Probably comes from treating a number like a boolean in a boolean concatenation.

`@walletconnect/utils` is also monorepo so shouldn't cause any misalignment issues in terms of versioning.

bkrem (approved):
Nice work here 🚀 just looks like some weirdness still happening in the `isValidObject` validation helper

packages/sign-client/src/controllers/engine.ts
ganchoradkov: The fx `isValidObject` returns `false` on invalid variables & length of the keys on valid objects. I'm utilizing the weirdness (`=== 0`) as the value `{}` is technically a valid object but without any params.

In most other cases empty object `{}` shouldn't be considered valid but in this case, it makes sense.

PR #1834, opened 2023-01-05
feat: pending pairigs cleanup

ganchoradkov:
# Description

Pending pairings have a current lifetime of 5 minutes. Rapid creation of multiple pairings in a short period of time (before the expirer kick in) might lead to degraded performance, especially during development, thus an update is required to clean-up pending pairings when a new one is initiated.

- added auto clean-up when `connect()` is called
- added public API `cleanupPendingPairings` clean-up
- integration test

## How Has This Been Tested?

npm run test
dogfooding in demo apps

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/providers/universal-provider/src/UniversalProvider.ts
+++ b/providers/universal-provider/src/UniversalProvider.ts
@@ -151,13 +151,13 @@ export class UniversalProvider implements IUniversalProvider {
     this.logger.info("Cleaning up inactive pairings...");
     const inactivePairings = this.client.pairing.getAll({ active: false });
 
-    if (!isValidArray(invactivePairings)) return;
+    if (!isValidArray(inactivePairings)) return;
 
-    invactivePairings.forEach((pairing) => {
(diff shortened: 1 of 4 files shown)


providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -146,6 +147,19 @@ export class UniversalProvider implements IUniversalProvider {
     }
   }
 
+  public cleanupPendingPairings(): void {
+    this.logger.info("Cleaning up inactive pairings...");
+    const invactivePairings = this.client.pairing.getAll({ active: false });
```
bkrem: Typo:

```suggestion
    const inactivePairings = this.client.pairing.getAll({ active: false });
```

bkrem (approved):
Smooth ✨

antondalgren:
Does this include cleaning up pairings that have had a session before but has been disconnected prior to calling `connect()`? @ganchoradkov

PR #1838, opened 2023-01-05
fix(scripts): differentiate `sed` command by platform

bkrem:
# Description

- The empty string `""` parameter for `sed` in the `update_relayer_sdk_version.sh` script seems to be macOS/BSD specific.
- Adds control flow to call on macOS with extra parameter, without by default.

## How Has This Been Tested?

- Tested on macOS via parallel GNU `sed` installation.

## Due Dilligence

* [ ] Breaking change


ganchoradkov (approved):
Works awesome :100:  just as mentioned in private, on linux there's the warning for the incorrect syntax `==` should be `=`

PR #1839, opened 2023-01-05
fix: check if relay event is related to a pairing topic

pedrouid:
# Description

Currently if you use Core API without registering with the Pairing API you will get an error regarding History even if the incoming payload was not targeting a pairing topic.

For example, I'm using a random symKey and topic to message between two clients and you can see in the screenshot below that I'm using a method `wc_syncSet` and the Pairing controller is automatically rejecting it.\

Screenshot 2023-01-05 at 18 22 06

However I get an error message that JsonRpcHistory controller is not matching any keys to this request and this happens on `onRelayEventResponse` when the Pairing controller tries to reject with `unknown method`

Screenshot 2023-01-05 at 18 22 13

## How Has This Been Tested?

I'm creating a prototype for the Sync API on this repo: https://github.com/WalletConnect/sync-api-testing

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/core/src/controllers/pairing.ts
+++ b/packages/core/src/controllers/pairing.ts
@@ -252,6 +252,8 @@ export class Pairing implements IPairing {
     const { topic, payload } = event;
     const reqMethod = payload.method as PairingJsonRpcTypes.WcMethod;
 
+    if (!this.pairings.keys.includes(topic)) return;
+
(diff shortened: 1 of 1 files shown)


bkrem (approved):
Neat fix, thank you ✨ 

FYI @ganchoradkov

PR #1843, opened 2023-01-06
fix: expirer pending requests

ganchoradkov:
# Description

An issue was identified with pending requests where on approving/rejecting a request was logging`{context: 'core'} {context: 'core/expirer'} 'No matching key. expirer: topic:id:1672999835033891'`.
The problem was caused by `formatTarget` being called twice on the same key in `expirer.del` flow thus prepending the `topic:..` & `id:...` to the key

Additionally, the expiration calculation was incorrect for pending requests and is also resolved in the PR

## How Has This Been Tested?
dogfooding
additional test for the expirer

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/core/test/expirer.spec.ts
+++ b/packages/core/test/expirer.spec.ts
@@ -23,7 +23,6 @@ describe("Expirer", () => {
         resolve();
       });
     });
-    // core.expirer.set("test")
     await disconnectSocket(core.relayer);
   });
 });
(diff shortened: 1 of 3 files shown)


packages/core/test/expirer.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+    expect(core.expirer.length).to.eq(1);
+    await new Promise<void>((resolve) => {
+      core.expirer.on(EXPIRER_EVENTS.expired, (payload: any) => {
+        expect(payload.target).to.eq(formatExpirerTarget("topic", topic));
+        // confirm the expirer is empty again
+        expect(core.expirer.length).to.eq(0);
+        resolve();
+      });
+    });
+    // core.expirer.set("test")
```
bkrem: Needed?

```suggestion

```

bkrem (approved):
LGTM, thanks for identifying and adding initial expirer coverage 🙏 💯

Issue #1739, opened 2022-12-08
Add timeout option for signClient.request 


junyulah:
**Is your feature request related to a problem? Please describe.**

The default timeout for request sign is 5 minutes in hard code way, but in some scenarios, we need to customize the timeout value. One possible scenario is the wallet sign is an async process and may take more than 5 minutes. It's a better way to expose this option to Dapps.

**Describe the solution you'd like**

- add timeout option in params
```javascript
  public request: IEngine["request"] = async <T>(params: EngineTypes.RequestParams) => {
    this.isInitialized();
    await this.isValidRequest(params);
    const { chainId, request, topic, timeout } = params;
    const id = await this.sendRequest(topic, "wc_sessionRequest", { request, chainId });
    const { done, resolve, reject } = createDelayedPromise<T>(timeout);
    this.events.once<"session_request">(engineEvent("session_request", id), ({ error, result }) => {
      if (error) reject(error);
      else resolve(result);
    });
    return await done();
  };
```
- support timeout option in createDelayedPromise
```javascript
export function createDelayedPromise<T>(timeout?: number) {
  timeout = toMiliseconds(timeout || FIVE_MINUTES);
```

bkrem:
Hi @junyulah,

Thanks for raising this point regarding customisable timeouts for sign requests, and your detailed suggestion for the changed implementation!

> One possible scenario is the wallet sign is an async process and may take more than 5 minutes.

We’re curious about your use case here, could you expand on the scenario where 5 mins is insufficient to respond to the sign request?

junyulah:
Hi @bkrem,

Thanks for the response. Currently we are developing a MPC wallet, so to sign a transaction in our wallet will need multiple signers to sign it. And it's possible our users will need more than 5 minutes to contact different signers to sign the transaction. In our real testing scenarios, it's possible to take even longer than half hour. Get back to this timeout issue, it would be a great help if we can customize the timeout option, so the Dapps can have the flexibility to set timeout to support our MPC signing scenarios.

bkrem:
Thanks for the context 💯 

As a follow-up question: How does the dapp know which wallets need 30mins (or more) and which are fine with the 5min default in your scenario? Would you distinguish by the peer's (i.e. wallet's) metadata or just up the timeout in general for all requests?

The team is aware and we'll discuss at our Sign sync early next week, so hoping we can get back to you after that regarding this suggestion 👍

junyulah:
[...]

PR #1851, opened 2023-01-09
feat: optional request expiry

ganchoradkov:
# Description

Relates to #1739 

Added an optional `expiry` parameter to session requests where a number between `300` & `604800` seconds (5 minutes min - 7 days max) can be specified to extend a request timeout 
## How Has This Been Tested?

new validation + integration tests
## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/packages/sign-client/src/constants/engine.ts
+++ b/packages/sign-client/src/constants/engine.ts
@@ -1,4 +1,4 @@
-import { FIVE_MINUTES, ONE_DAY, THIRTY_SECONDS } from "@walletconnect/time";
+import { FIVE_MINUTES, ONE_DAY, SEVEN_DAYS, THIRTY_SECONDS } from "@walletconnect/time";
 import { EngineTypes } from "@walletconnect/types";
 
 export const ENGINE_CONTEXT = "engine";
@@ -104,6 +104,6 @@ export const ENGINE_RPC_OPTS: EngineTypes.RpcOptsMap = {
 };
 
 export const SESSION_REQUEST_EXPIRY_BOUNDARIES = {
-  min: 300,
-  max: 604800,
+  min: FIVE_MINUTES,
+  max: SEVEN_DAYS,
 };
--- a/packages/sign-client/src/constants/engine.ts
+++ b/packages/sign-client/src/constants/engine.ts
@@ -102,3 +102,8 @@ export const ENGINE_RPC_OPTS: EngineTypes.RpcOptsMap = {
     },
   },
 };
+
+export const SESSION_REQUEST_EXPIRY_BOUNDARIES = {
+  min: 300,
+  max: 604800,
+};
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -51,8 +51,14 @@ import {
   TYPE_1,
   getRequiredNamespacesFromNamespaces,
   isValidObject,
+  isValidRequestExpiry,
 } from "@walletconnect/utils";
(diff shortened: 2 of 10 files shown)


packages/sign-client/src/constants/engine.ts
```diff
@@ -102,3 +102,8 @@ export const ENGINE_RPC_OPTS: EngineTypes.RpcOptsMap = {
     },
   },
 };
+
+export const SESSION_REQUEST_EXPIRY_BOUNDARIES = {
+  min: 300,
```
bkrem: Can we use the labelled values from `@walletconnect/time` here instead of inline numbers as above? Should have both 5 mins and 7 days available already as far as I remember

packages/utils/src/validators.ts
```diff
@@ -489,3 +489,7 @@ export function isConformingNamespaces(
 
   return error;
 }
+
+export function isValidRequestExpiry(expiry: number, boudaries: { min: number; max: number }) {
```
bkrem: Param typo:

```suggestion
export function isValidRequestExpiry(expiry: number, boundaries: { min: number; max: number }) {
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -258,9 +264,9 @@ export class Engine extends IEngine {
   public request: IEngine["request"] = async <T>(params: EngineTypes.RequestParams) => {
     this.isInitialized();
     await this.isValidRequest(params);
-    const { chainId, request, topic } = params;
-    const id = await this.sendRequest(topic, "wc_sessionRequest", { request, chainId });
-    const { done, resolve, reject } = createDelayedPromise<T>();
+    const { chainId, request, topic, expiry } = params;
```
bkrem, line 267: Thought: we need to make sure we document this param properly regarding the expected value always being in seconds. Otherwise I think we'll end up with confused devs who aren't clear on what unit should be passed for `expiry`.

This already trips us up internally and tbh I'd prefer if we use labelling like `FIVE_MINUTES_IN_SECONDS`, but that's a whole other refactoring story 😅

packages/sign-client/test/sdk/client.spec.ts
```diff
@@ -227,4 +229,33 @@ describe("Sign Client Integration", () => {
...
+      const {
+        sessionA: { topic },
+      } = await testConnectMethod(clients);
+
+      const expiry = 10000;
+
+      await Promise.all([
+        new Promise<void>((resolve) => {
+          clients.A.core.relayer.once(
+            RELAYER_EVENTS.publish,
```
bkrem, line 245: Was `RELAYER_EVENTS.publish` specifically added for hooking into here in the test or am I missing it being listened to in the source somewhere? (cmd+f on the diff only showed me this).

bkrem (approved):
LGTM, thanks for getting this shipped 🚀 

Relates to #1739

packages/sign-client/src/controllers/engine.ts
ganchoradkov: Good idea. I tried to provide a descriptive error message in case an invalid value is provided and also updated it to include that the value is expected in seconds

packages/sign-client/test/sdk/client.spec.ts
ganchoradkov: This event was requested by the `Chat` team and was in the backlog to be added. I thought that was a good use case to utilize it in the expiry test and ship it

bkrem: Agreed in that case 👍 Thanks for context!

PR #1861, opened 2023-01-11
feat:cosmos provider

ganchoradkov:
# Description
Added `cosmos` support to the `universal-provider` where dapps can requests such as `cosmos_signDirect, cosmos_signAmino` to a wallet peer

## How Has This Been Tested?
new integration tests
dogfooding in cosmos example dapp

## Due Dilligence

* [ ] Breaking change


providers/universal-provider/src/providers/solana.ts
```diff
@@ -26,6 +32,25 @@ class SolanaProvider implements IProvider {
     this.namespace = Object.assign(this.namespace, namespace);
   }
 
+  public requestAccounts(): string[] {
```
bkrem, line 35: Intentional that the Solana provider is being updated in this branch/PR too?

bkrem (approved):
Clean ✨ 

Tagged @0xAsimetriq too in case he wants to take a look

xzilja (approved):
Nice! As we are now venturing into multi-chain support via single provider package, should we move things like `cosmos-wallet` (and sollana / ethereum) equivalents into `peerDependencies` and mark them as optional, as to not pre-bundle everything and make universal-provider leaner?

Therse peer dependencies should be mentioned in the docs as additional installs.

providers/universal-provider/src/providers/solana.ts
ganchoradkov: It's intentional as I've refactored the `requestAccounts` fx. Before, it was requesting accounts by sending a request with `eth_getAccounts` which doesn't work outside of evm

ganchoradkov:
>Nice! As we are now venturing into multi-chain support via single provider package, should we move things like cosmos-wallet (and sollana / ethereum) equivalents into peerDependencies and mark them as optional, as to not pre-bundle everything and make universal-provider leaner?

@0xAsimetriq the `cosmos-wallet` is just dev dependency to help sign/verify the tests. Shouldn't have an impact over prod builds

PR #1870, opened 2023-01-16
fix: Add reject value when rejecting connection promise on transportClose

amiryadid-fb:
# Description

On relayer `transportOpen`, when waiting for `provider.connect` to resolve, it might get stuck so there's a `Promise.race` to reject the promise if `transportClosed` is called during connection.
In the case of rejecting it like so, the reject is called with no value which causes the `catch` close to throw another exception (since it tried to access `message` of undefined) instead of handling the current one correctly.

## Due Diligence

- [ ] Breaking change


ganchoradkov (approved):
Good catch :100:  thanks for resolving this!
I'm seeing the CI is failing but that is due to no access to org ci params

PR #1876, opened 2023-01-19
fix: chainId incorrect in WalletConnect request

micwallace:
# Description

This fixes an issue where the requested chain or the default chain is not set correctly in the wc_sessionRequest when using the EIP155 provider. 

In the EIP155 provider, a request to wallet_switchEthereumChain sets the internal chainId property, but this value isn't used in subsequent client requests. 

Furthermore, the chainId supplied in the request parameters is forwarded to the wallet but isn't used in most cases, as WalletConnect has it's own chainId outside of the Ethereum request parameters:

```
{
  "request": {
    "method": "eth_sendTransaction",
    "params": [
      {
        "chainId": "0x5", // This parameter may be ignored by the wallet in preference of the top-level chainId
        "gas": "0x5208",
        "value": "0x5af3107a4000",
        "from": "0x8646df47d7b16bf9c13da881a2d8cdacda8f5490",
        "to": "0xcff805b714b24b3dd30cb4a1bea3745e5c5e73ef"
      }
    ]
  },
  // This parameter wasn't set correctly using the chainId in eip155.ts - it remained the first specified chain when connecting.
  "chainId": "eip155:5", 
  "topic": "a7f2fb29c8146ceb609d257c60f264d1d1c9214d08e10ed91212b1b03e73c578"
}
```

Resolves 

## How Has This Been Tested?

I have created a simple Dapp using universal provider and ethers.js. It has two buttons, each of which sends a transaction on a different chain. I have tested that both methods of specifying the correct chain work:

1. Call wallet_ethereumSwitchChain with the required chain OR
2. Specify hex chainId in the ethereum request parameters

Both of these methods now use the correct chain in AlphaWallet.

## Due Diligence

- [ ] Breaking change


ganchoradkov (commented):
Hey, thanks for the update but doesn't the `provider.request()` already accept an optional `chainid` that is added to the request object or I'm missing something?
https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/providers/universal-provider/src/UniversalProvider.ts#L48

```
 public async request<T = unknown>(
    args: RequestArguments,
    chain?: string | undefined,
  ): Promise<T> {
    const [namespace, chainId] = this.validateChain(chain);

    return await this.getProvider(namespace).request({
      request: {
        ...args,
      },
      chainId: `${namespace}:${chainId}`,
      topic: this.session?.topic,
    });
  }
```

micwallace:
It does, but if I don't specify the chainId then the chain will always be the first chain specified when connecting. wallet_switchEthereumChain sets the default chainId variable, but it's never used in subsequent requests. 

Ethers is calling the request method in this instance, so it's likely not setting this param. 

In saying that this PR needs some work in-case that variable is already set by the parameter you mentioned. I will take a closer look when I get back to my computer.

Another issue I encountered is switching to a chain that wasn't initially specified when pairing. This could be fixed by updating the WC session in the switchEthereumChain request.

ganchoradkov:
@micwallace I see your point, in that case shouldn't the check be
```
args.chainId = args.chainId || `${this.name}:${this.getDefaultChainId()}`;
```
i.e. to use the one provided as an argument OR the default

micwallace:
@ganchoradkov Yes indeed, I had my solution a bit to deep in the codebase and it needed some more extensive changes to make sure all bases are covered. Please see last commit message.

ganchoradkov:
Hey @micwallace could you please get the branch up-to-date with v2

micwallace:
Done, thanks @ganchoradkov

micwallace:
@ganchoradkov Any update on this?

micwallace:
@ganchoradkov This seems to be fixed in the latest version, thanks mate!

PR #1887, opened 2023-01-23
Port cacao utils from auth

devceline:
# Description
* Port a copy of auth cacao  formatting because it will be shared with chat for keyserver API

## Due Diligence

- [x] Breaking change (none)
- [x] Requires a documentation update (none)
- [x] Requires a e2e/integration test update (none)

### DIFF
--- a/packages/utils/src/cacao.ts
+++ b/packages/utils/src/cacao.ts
@@ -12,7 +12,6 @@ export interface CacaoPayload {
   resources?: string[];
 }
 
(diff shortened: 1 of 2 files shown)


ganchoradkov (changes requested):
Could you check what `prettier` is complaining about?

ganchoradkov (approved):

bkrem (approved):

PR #1894, opened 2023-01-24
fix(template): tweaks new-chains template w/ additional required data

bkrem:
# Description

- Adds some additional data points I realised we weren't specifying currently in the template, but are needed for us to add the new namespace/chain(s)


finessevanes (approved):
LGTM

PR #1895, opened 2023-01-25
refactor: disconnect pairings in universal-provider

ganchoradkov:
# Description
Reworked the pending pairings cleanup to unsubscribe the pairings instead of deleting them to avoid disruption of internal `SignClient/Core` flows

## How Has This Been Tested?
integration tests
dogfooding

## Due Diligence

- [ ] Breaking change


xzilja (approved):
Honestly, turned out even more elegant than before. Quick question, now that we are deleting subs, when expirer kicks in and deletes an outdated pairing, would it cause / throw an error due to associated subscription not existing?

ganchoradkov:
>now that we are deleting subs, when expirer kicks in and deletes an outdated pairing, would it cause / throw an error due to associated subscription not existing?

Tested that scenario and no exceptions were raised as the subscriber has built-in checks

PR #1918, opened 2023-01-31
Fixes how RPC URL is retrieved for Cardano Provider

ggaabe:
# Description

RPC Provider was returning undefined for certain calls on cardano. This is due to the chain ID being retrieved from the rpc map being incorrect.

We also now use an internal getCardanoRPCUrl, because the standard `getRPCUrl` method in universal-provider relies on `getChainId`, which attempts to parse the second half of the CAIP-2 chain as a number. This doesn't work for cardano, because cip-34 chain IDs have a `-` hyphen.


ganchoradkov:
Thanks @ggaabe, for the patch! :100: :100:

ganchoradkov (approved):

PR #1921, opened 2023-02-01
feat: verify-api

ganchoradkov:
# Description
`Verify API` implementation according to specs https://github.com/WalletConnect/walletconnect-docs/pull/436

## How Has This Been Tested?
tests
dogfooding

## Due Diligence

- [ ] Breaking change
- [x] Requires a documentation update
  - [] Related docs issue/PR (if docs are not included in this PR):


packages/core/src/controllers/verify.ts
```diff
@@ -0,0 +1,86 @@
...
+    this.logger = generateChildLogger(logger, this.name);
+    this.verifyUrl = VERIFY_SERVER;
+    this.abortController = new AbortController();
+  }
+
+  public init: IVerify["init"] = async (params) => {
+    // ignore on non browser environments
+    if (!isBrowser() || !document) return;
+
+    this.verifyUrl = params?.verifyUrl || VERIFY_SERVER;
```
xzilja, line 28: Since `this.verifyUrl` is set here with fallback, maybe safe to remove it from constructor?

xzilja (approved):
Looks good 👍 It seems like there are 2 approaches used to "abort" promises that are in flight, one is using abort controller in fetch and the other approach is a `Promise.race` for original promise and artificial delay one that rejects.

Abort one seems to be like a more correct approach here, so wanted to check if it can be used everywhere or are there any specifics here to why you used 2 approaches?

packages/core/src/controllers/verify.ts
ganchoradkov: it is, but typescript then complains about
- this.verifyUrl is not definitely assigned in the constructor
- requires everywhere to be checked that is assigned i.e.  `?.` on every usage

arein (approved):

pedrouid (approved):
👍

PR #1922, opened 2023-02-01
Fix/w3m universal provider

xzilja:
# Description

Fixes issue where in some cases (wagmi) a `setDefaultChain` function needs to be called before `connect`. `setDefaultChain` call will fail if it was called after `disconnect` due to rpc object being cleaed up (which we don't need to do).

## How Has This Been Tested?
Tested within wagmi walletconnect connector.

## Due Diligence

- [ ] Breaking change

### DIFF
--- a/providers/universal-provider/package.json
+++ b/providers/universal-provider/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@walletconnect/universal-provider",
   "description": "Universal Provider for WalletConnect Protocol",
-  "version": "2.3.2-858d3960",
+  "version": "2.3.2",
   "author": "WalletConnect, Inc. <walletconnect.com>",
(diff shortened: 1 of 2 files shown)


providers/universal-provider/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@walletconnect/universal-provider",
   "description": "Universal Provider for WalletConnect Protocol",
-  "version": "2.3.2",
+  "version": "2.3.2-858d3960",
```
ganchoradkov: I guess you don't want to commit this 😅

ganchoradkov (changes requested):

ganchoradkov (approved):

PR #1932, opened 2023-02-03
Add deeplink redirects for rpc requests

xzilja:
# Description

Client side libraries like web3modal / rainbowkit and connectkit set localstorage entry with deeplink or universal href for connected mobile wallets. We need to redirect to these during rpc requests (if such entry exists in localStorage)

## Due Diligence

- [ ] Breaking change


ganchoradkov (approved):
Awesome :100:

PR #1935, opened 2023-02-03
fix: allow `showQrModal` to be `false`

iamacook:
# Description

If `showQrModal` was set to `false`, it would be overwritten to `true`. It is now possible to set it to `false` and the setting remain.

## How Has This Been Tested?

Calling `EthereumProvider.init` with a `showQrModal` value returns `rpc.showQrModal` as specified.

## Due Diligence

- [ ] Breaking change


ganchoradkov:
Hey @iamacook, thanks for catching that :100:

ganchoradkov (approved):

PR #1940, opened 2023-02-06
fix: subscriber not initialized

ganchoradkov:
PR to address some edge cases in the `subscriber`
- subscriber not initialized
This happens when a subscription is attempted before the subscriber has finished initializing
- RESTORE_WILL_OVERRIDE
This happens when `RELAYER_EVENTS.connect` is emitted 2+ times in a short sequence before the previous restart has finished

Misc changes
-  subscription ID -> result of `hashMessage(topic + clientId)`
- skip contract deployment tests to unblock CI while relay msg payload size is increased again
# Description

## How Has This Been Tested?
integration test
dogfooding

## Due Diligence

- [ ] Breaking change

### DIFF
--- a/packages/core/src/controllers/subscriber.ts
+++ b/packages/core/src/controllers/subscriber.ts
@@ -412,7 +412,7 @@ export class Subscriber extends ISubscriber {
   }
 
   private async restartToComplete() {
-    if (this.restartInProgress) return;
+    if (!this.restartInProgress) return;
 
     await new Promise<void>((resolve) => {
       setInterval(() => {
--- a/packages/core/src/controllers/subscriber.ts
+++ b/packages/core/src/controllers/subscriber.ts
@@ -361,6 +361,7 @@ export class Subscriber extends ISubscriber {
   }
 
   private async onConnect() {
+    if (this.restartInProgress) return;
     await this.restart();
     this.onEnable();
   }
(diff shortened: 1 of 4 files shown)


packages/core/src/controllers/subscriber.ts
```diff
@@ -401,4 +410,16 @@ export class Subscriber extends ISubscriber {
...
     }
   }
+
+  private async restartToComplete() {
+    if (this.restartInProgress) return;
+
+    await new Promise<void>((resolve) => {
+      setInterval(() => {
+        if (!this.restartInProgress) {
+          resolve();
```
bkrem, line 421: Think we need to call `clearInterval(id)` here before resolving to avoid possible memory leakage from zombie intervals: https://developer.mozilla.org/en-US/docs/Web/API/setInterval#return_value

ganchoradkov: I thought to add it & forgot 🤦

xzilja (approved):
Looks good, double checking if ERC20 tests need to remain skipped?

packages/core/src/controllers/subscriber.ts
```diff
@@ -401,4 +410,16 @@ export class Subscriber extends ISubscriber {
       throw new Error(message);
     }
   }
+
+  private async restartToComplete() {
+    if (this.restartInProgress) return;
```
bkrem: Is there a missing negation `!` here or am I misunderstanding it? Seems like it returns immediately if a restart is in progress when the intention seems to be to create a locking mechanism to hold off on the `isInitialized` check being called if a restart is in progress.

bkrem (approved):

PR #1951, opened 2023-02-08
fix(core): ensures `core.relayUrl` property is assigned

bkrem:
# Description

## Context
- While dogfooding JS push + cast server interactions, I realised that `core.relayUrl` is defined as a property but never assigned. `opts.relayUrl` is simply passed through to `Relayer`.
- Accessing `core.relayer.relayUrl` also doesn't work because it's a private property there.

## Solution

- Ensure the already defined `core.relayUrl` prop is assigned properly, so it can be accessed/referenced from outside the client as expected.

## How Has This Been Tested?

- Unit tests

## Due Diligence

- [ ] Breaking change


bkrem:
@ganchoradkov @arein FYI also seeing that `Subscription not found` error on the CI here that I believe you guys already started to discuss:
https://github.com/WalletConnect/walletconnect-monorepo/actions/runs/4127743528/jobs/7131372171#step:6:26

arein (approved):

ganchoradkov (approved):

PR #1956, opened 2023-02-09
feat: optional namespaces in providers

### LINKED ISSUE #869: Walletconnect modal close not affect on provider event
**Describe the bug**
Potential memory leak on close Qr Modal on UI without success connection  

**SDK Version**
 - @walletconnect/ethereum-provider@2.3.3

**To Reproduce**
1. Import EthereumProvider
2. Store provider instance `const _provider = await EthereumProvider.init(options)` (showQrModal should be true or be undefined)
3. If the session is not defined try to connect `!_provider.session && (await _provider.connect())`
4. Click the cross button and close the modal window

**Expected behavior**
`await _provider.connect()` finished or throw error

**Actual behavior**
Promise never ended

**Additional context**
I don't know, maybe it could be handled in another way, but for me looks like we have bug with unhandled user modal close

ganchoradkov:
# Description

Update in `@walletconnect/universal-provider`:
- Implemented `optionalNamespaces` & `sessionProperties`

Update in `@walletconnect/ethereum-provider`:
- reduced default methods to `["eth_sendTransaction", "personal_sign"];`
- added init parameter called `optionalChains`
- util fx to build required & optional namespaces based on
       - providing more methods than the default two 
       - providing data in `optionalChains`
- throws an exception when web3modal is closed during pairing  

Init config of `ethereum-provider`
```
await EthereumProvider.init({
      projectId,
      chains: [1],
      methods: [
             "eth_sendTransaction",
             "personal_sign",
         ],
      optionalMethods: [
             "eth_signTransaction",
             "eth_signTypedData",
         ],
});
```
produces the following `requiredNamespaces` & `optionalNamespaces`
```
{  // required
  eip155: {
    chains: [ 'eip155:1' ],
    methods: [ 'eth_sendTransaction', 'personal_sign' ],
    events: [ 'chainChanged', 'accountsChanged' ],
    rpcMap: {
      '1': 'https://rpc.walletconnect.com/v1/?chainId=eip155:1&projectId=942cb705fcbddae9f61bc75f7fa14855'
    }
}

{  // optional
 eip155: {
    chains: [ 'eip155:1' ],
    methods: [
      'eth_sendTransaction',
      'personal_sign',
      'eth_signTransaction',
      'eth_signTypedData'
    ],
    events: [ 'chainChanged', 'accountsChanged' ],
    rpcMap: {
      '1': 'https://rpc.walletconnect.com/v1/?chainId=eip155:1&projectId=942cb705fcbddae9f61bc75f7fa14855'
    }
  }
}

```

## How Has This Been Tested?
integration tests
dogfooding

## Due Diligence

- [ ] Breaking change

closes https://github.com/WalletConnect/web3modal/issues/869

### DIFF
--- a/providers/ethereum-provider/src/EthereumProvider.ts
+++ b/providers/ethereum-provider/src/EthereumProvider.ts
@@ -5,22 +5,13 @@ import {
   ProviderAccounts,
   RequestArguments,
 } from "eip1193-provider";
-import { Metadata, UniversalProvider } from "@walletconnect/universal-provider";
+import { Metadata, Namespace, UniversalProvider } from "@walletconnect/universal-provider";
 import { Web3Modal } from "@web3modal/standalone";
 import { SignClientTypes } from "@walletconnect/types";
 
 export const RPC_URL = "https://rpc.walletconnect.com/v1/";
 
-export const signerMethods = [
-  "eth_requestAccounts",
-  "eth_accounts",
-  "eth_chainId",
-  "eth_sendTransaction",
-  "eth_signTransaction",
-  "eth_sign",
-  "eth_signTypedData",
-  "personal_sign",
-];
+export const signerMethods = ["eth_sendTransaction", "personal_sign"];
 
 export const signerEvents = ["chainChanged", "accountsChanged"];
 
@@ -35,6 +26,7 @@ export interface SessionEvent {
 
 export interface EthereumRpcConfig {
   chains: string[];
+  optionalChains?: string[];
   methods: string[];
   events: string[];
   rpcMap: EthereumRpcMap;
@@ -42,9 +34,9 @@ export interface EthereumRpcConfig {
   metadata?: Metadata;
   showQrModal: boolean;
 }
-
 export interface ConnectOps {
   chains?: number[];
+  optionalChains?: number[];
   rpcMap?: EthereumRpcMap;
   pairingTopic?: string;
 }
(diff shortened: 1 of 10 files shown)


providers/ethereum-provider/src/EthereumProvider.ts
```diff
@@ -64,9 +56,61 @@ export function getRpcUrl(chainId: string, rpc: EthereumRpcConfig): string | und
...
+    chains: requiredChains,
+    methods: requriedMethods,
+    events: requiredEvents,
+    rpcMap: requiredRpcMap,
+  };
+
+  const optionalMethods = methods.filter((method) => !requriedMethods.includes(method));
+  const optionalEvents = events.filter((method) => !requiredEvents.includes(method));
+
+  const additionalPermissionRequired = optionalMethods.length !== 0 || optionalEvents.length !== 0;
```
xzilja:
```suggestion
  const additionalPermissionRequired = optionalMethods.length || optionalEvents.length;
```

xzilja (commented):
Nice, few follow up questions

1. Are we allowing more than 1 required chain? If not, I think we should rename field to be `requiredChain` and make it a singular number instead of array

2. I'm not sure but it looks like logic inside `buildNamespaces` might be returning wrong data.

```ts
  const requriedMethods = methods.length > signerMethods.length ? signerMethods : methods;
  const requiredEvents = events.length > signerEvents.length ? signerEvents : events;
```
^ Above allows any methods to become required if 2 or less are passed in?

```ts
const optional: Namespace = {
    chains: additionalPermissionRequired ? requiredChains.concat(optionalChains) : optionalChains,
    methods,
    events,
    rpcMap,
  };
```
^ Above looks like it is including original unfiltered methods and events into optional namespace?

Would it be better to just hardcode required ones in and for optionals take user input and filter out required methods?

ganchoradkov:
>Are we allowing more than 1 required chain? If not, I think we should rename field to be requiredChain and make it a singular number instead of array

Is there particular reason why we shouldn't allow more chains? If you want to require just one, you're good, if you want more - you're good as well. I'm open for discussions.

>  const requriedMethods = methods.length > signerMethods.length ? signerMethods : methods;
  const requiredEvents = events.length > signerEvents.length ? signerEvents : events;
>^ Above allows any methods to become required if 2 or less are passed in?

fair point. I kinda see some benefit where a dap can explicitly require particular method this way but might be not the most intuitive usage. Because if all methods go directly to `optional` there is no way for you to require additional ones.


>const optional: Namespace = {
    chains: additionalPermissionRequired ? requiredChains.concat(optionalChains) : optionalChains,
    methods,
    events,
    rpcMap,
  };
^ Above looks like it is including original unfiltered methods and events into optional namespace?
Would it be better to just hardcode required ones in and for optionals take user input and filter out required methods?

Filtering the methods in theory is fine, but with the current structure it causes additional issues. In the example in the description, if we are to filter the methods, chain `'eip155:42'` is left without `sendTransaction` & `personal_sign`

pedrouid (approved):
👍

PR #1966, opened 2023-02-10
feat: session request sent

ganchoradkov:
# Description
Added new event emitted from `signClient` called `session_request_sent`.
Also updated the deep links handling in `universal-provider` to use that event instead of `relayer_events.publish` 

## How Has This Been Tested?
dogfooding

## Due Diligence

- [ ] Breaking change

### DIFF
--- a/providers/universal-provider/src/utils/deepLinks.ts
+++ b/providers/universal-provider/src/utils/deepLinks.ts
@@ -1,0 +1,31 @@
+import { SignClient } from "@walletconnect/sign-client";
+import { SignClientTypes } from "@walletconnect/types";
+import { RequestParams } from "../types";
+
+export const handleDeepLinks = (client: InstanceType<typeof SignClient>, params: RequestParams) => {
+  const handleSessionRequestSent = (
+    payload: SignClientTypes.EventArguments["session_request_sent"],
+  ) => {
+    // only handle the request if it matches the request and topic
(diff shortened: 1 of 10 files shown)


providers/universal-provider/src/utils/deepLinks.ts
```diff
@@ -0,0 +1,31 @@
...
+export const handleDeepLinks = (client: InstanceType<typeof SignClient>, params: RequestParams) => {
+  const handleSessionRequestSent = (
+    payload: SignClientTypes.EventArguments["session_request_sent"],
+  ) => {
+    // only handle the request if it matches the request and topic
+    if (payload.request !== params.request || payload.topic !== params.topic) return;
+    client.events.removeListener("session_request_sent", handleSessionRequestSent);
+    deeplinkRedirect();
+  };
+  client.on("session_request_sent", handleSessionRequestSent);
```
xzilja, line 14: Maybe we can use `.once` here instead of `.on`

xzilja (approved):
Nice, great idea with comparing args as well 👍

providers/universal-provider/src/utils/deepLinks.ts
ganchoradkov: `.once` will not work if a different request is received by the listener. In the fx, after confirming we've received the correct request, I'm removing the listener so we shouldn't have issues with memory leaks

PR #1985, opened 2023-02-14
export method and event constants, types

xzilja:
# Description

Exports method and event types alongside pre-defined constants to be used in downstream packages dependant on ethereum-provider

## How Has This Been Tested?
-

## Due Diligence

- [ ] Breaking change


providers/ethereum-provider/src/index.ts
```diff
@@ -1,3 +1,5 @@
 import { EthereumProvider as Provider } from "./EthereumProvider";
 export const EthereumProvider = Provider;
+export type { RpcEvent, RpcMethod } from "./EthereumProvider";
+export { OPTIONAL_METHODS, REQUIRED_EVENTS, REQUIRED_METHODS, OPTIONAL_EVENTS } from "./constants";
```
ganchoradkov: if you're exporting everything from `./constants/rpc` would it be better to export all? 
When we add new constants, we wouldn't have to update the `index.ts` as well 
```suggestion
export * from "./constants/rpc";
```

ganchoradkov (approved):
Proposed potential small change otherwise everything looks :100:

PR #1999, opened 2023-02-15
refactor: active chains persitence

### LINKED ISSUE #1994: ethereum-provider: the `disconnect` event is not emitted
**Describe the bug**
The `disconnect` event is not emitted when the session is closed on the wallet. The `session_delete` even is emitted, though, but I guess it should be re-mapped.

**SDK Version (if relevant)**
 - Client: ethereum-provider
 - Version 2.4.3

**Additional context**
Tested with a beta version of Bifrost wallet (Android)

### LINKED ISSUE #1995: ethereum-provider: calling `wallet_swithEthereumChain` succeeds even if the network is not supported
**Describe the bug**
The call to the  `wallet_swithEthereumChain` function with a chainId that is not supported by the wallet returns successfully and emits the chainChanged event.

**SDK Version (if relevant)**
 - Client: ethereum-provider
 - Version 2.4.3

**To Reproduce**
1. initialize the provider with 1 mandatory chain and a few optional chains, passing an explicit `rpcMap`
2. create the session
3. switch to one of the optional chains that is not supported by the wallet
4. chainid changed and event emitted

**Expected behavior**
I'd expect the  `wallet_swithEthereumChain` function to throw an exception, instead of switching to a RPC endpoint without signer support

**Additional context**
Reproduced with a beta version of Bifrost wallet (Android)

ganchoradkov:
# Description
PR to address several issues with active chain persistence in the providers and chain switching to unauthorized chains
The change log includes:
- persistence of active chainId on client reload
- throws an exception when attempting to switch chain that is not approved
- emits `disconnect` event on `session_delete`
- refactor/simplifies required & optional namespace usage

## How Has This Been Tested?
integration tests
dogfooding

## Due Diligence

- [ ] Breaking change

### DIFF
--- a/providers/ethereum-provider/src/EthereumProvider.ts
+++ b/providers/ethereum-provider/src/EthereumProvider.ts
@@ -1,5 +1,5 @@
 import { EventEmitter } from "events";
-import { getAccountsFromNamespaces, isValidArray } from "@walletconnect/utils";
+import { getAccountsFromNamespaces, getSdkError, isValidArray } from "@walletconnect/utils";
 import {
   IEthereumProvider as IProvider,
   ProviderAccounts,
@@ -340,6 +340,10 @@ export class EthereumProvider implements IEthereumProvider {
       (payload: SignClientTypes.EventArguments["session_delete"]) => {
         this.reset();
         this.events.emit("session_delete", payload);
+        this.events.emit("disconnect", {
+          ...getSdkError("USER_DISCONNECTED"),
+          data: payload.topic,
+        });
       },
     );
 
--- a/providers/ethereum-provider/src/EthereumProvider.ts
+++ b/providers/ethereum-provider/src/EthereumProvider.ts
@@ -321,9 +321,10 @@ export class EthereumProvider implements IEthereumProvider {
       this.events.emit("session_event", payload);
     });
 
-    this.signer.on("chainChanged", (chainId: number) => {
-      this.chainId = chainId;
-      this.events.emit("chainChanged", chainId);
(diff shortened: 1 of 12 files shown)


providers/ethereum-provider/src/EthereumProvider.ts
```diff
@@ -351,8 +356,10 @@ export class EthereumProvider implements IEthereumProvider {
   }
 
   private setHttpProvider(chainId: number): void {
-    const formattedChain = this.formatChainId(chainId);
-    this.signer.setDefaultChain(formattedChain, this.getRpcUrl(chainId));
+    this.request({
+      method: "wallet_switchEthereumChain",
+      params: [{ chainId: chainId.toString(16) }],
```
xzilja, line 361: What is `.toString(16)` used here for?

providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -20,7 +20,7 @@ import {
   PairingsCleanupOpts,
 } from "./types";
 
-import { RELAY_URL, LOGGER, STORAGE } from "./constants";
+import { RELAY_URL, LOGGER, STORAGE, PROVIVER_EVENTS } from "./constants";
```
xzilja: `PROVIVER_EVENTS` -> `PROVIDER_EVENTS`

xzilja (approved):
💯

providers/ethereum-provider/src/EthereumProvider.ts
ganchoradkov: the param is hexadecimal string

PR #2011, opened 2023-02-17
refactor: chainChanged in hex

ganchoradkov:
# Description
Updated `ethereum-provider` to emit `chainChanged` in hex instead of integer to be 1193 compliant.
Also mapped emit to proper event types

## How Has This Been Tested?
integration tests

## Due Diligence

- [ ] Breaking change

### DIFF
--- a/providers/ethereum-provider/src/EthereumProvider.ts
+++ b/providers/ethereum-provider/src/EthereumProvider.ts
@@ -81,6 +81,10 @@ export function getEthereumChainId(chains: string[]): number {
   return Number(chains[0].split(":")[1]);
 }
 
+export function toHexChainId(chainId: number): string {
+  return `0x${chainId.toString(16)}`;
+}
+
 export type NamespacesParams = {
   chains: EthereumRpcConfig["chains"];
   optionalChains?: EthereumRpcConfig["optionalChains"];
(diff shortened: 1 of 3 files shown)


providers/ethereum-provider/src/EthereumProvider.ts
```diff
@@ -316,15 +320,15 @@ export class EthereumProvider implements IEthereumProvider {
       } else if (event.name === "chainChanged") {
         this.setChainId(this.formatChainId(event.data));
       } else {
-        this.events.emit(event.name, event.data);
+        this.events.emit(event.name as any, event.data);
```
bkrem, line 323: Wouldn’t this always be `string`? Why the cast to any here?

bkrem (approved):

providers/ethereum-provider/src/EthereumProvider.ts
ganchoradkov: the `events.emit` is now typed to these events 
```
type Event =
    | "connect"
    | "disconnect"
    | "message"
    | "chainChanged"
    | "accountsChanged"
    | "session_delete"
    | "session_event"
    | "session_update"
    | "display_uri";
```
but technically you could also emit custom event

PR #2029, opened 2023-02-22
fix: pass dapp metadata to UniversalProvider

### LINKED ISSUE #2028: EthereumProvider is not passing the dapp metadata to its UniversalProvider instance
**Describe the bug**
The EthereumProvider does not pass down then dapp metadata to the UniversalProvider. In React Native, this causes the wallets to not show the dapp information and for some wallets it doesn't work at all without a dapp name and url (e.g. TrustWallet)

**SDK Version (if relevant)**
 - Version [2.4.5 of the EthereumProvider]

**To Reproduce**
- In this line: https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/providers/ethereum-provider/src/EthereumProvider.ts#L449 we should be passing the dapp metadata to the UniversalProvider

**Expected behavior**
- I expect the metadata to be passed to the UnivProvider.

iketw:
# Description
We need to pass the dapp metadata to the UniversalProvider. In React Native, we'd like to be able to show the dapp's name and url. Also, some wallets do not work without a name and url when connecting to them (namely: TrustWallet)

Resolves #2028

## How Has This Been Tested?
Used the new version in a small RN app to verify dapp metadata was being correctly pass through to the Universal Provider

## Due Diligence

- [ ] Breaking change

### DIFF
--- a/providers/ethereum-provider/src/EthereumProvider.ts
+++ b/providers/ethereum-provider/src/EthereumProvider.ts
@@ -446,7 +446,7 @@ export class EthereumProvider implements IEthereumProvider {
   private async initialize(opts: EthereumProviderOptions) {
     this.rpc = this.getRpcConfig(opts);
     this.chainId = getEthereumChainId(this.rpc.chains);
-    this.signer = await UniversalProvider.init({ projectId: this.rpc.projectId });
+    this.signer = await UniversalProvider.init({ projectId: this.rpc.projectId, metadata: opts.metadata });
     this.registerEventListeners();
     await this.loadPersistedSession();
     if (this.rpc.showQrModal) {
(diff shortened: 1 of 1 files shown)


ganchoradkov:
Hey @iketw, thanks for patching this up! :100:

ganchoradkov (approved):

iketw:
> Hey @iketw, thanks for patching this up! 💯

No problem, happy to help. Hey, I'm building a ReactNative package to support dapp creation and I want to use the `@walletconnect/legacy-provider` for WCV1. The issue is that that package is not reporting a `display_uri` event like the `@walletconnect/ethereum-provider` is doing. Do you know which package I can use for this? Alternatively I can propose a change for that legacy package

PR #2032, opened 2023-02-22
feat: `irn_batchSubscribe`

ganchoradkov:
# Description
Implemented `irn_batchSubscribe` with private `rpcBatchSubscribe` to handle subscriptions during initial initialization and after connection restart.

## How Has This Been Tested?
persistence tests in various packages are covering this 
dogfooding

## Due Diligence

- [ ] Breaking change

### DIFF
--- a/packages/core/src/controllers/subscriber.ts
+++ b/packages/core/src/controllers/subscriber.ts
@@ -227,6 +227,30 @@ export class Subscriber extends ISubscriber {
     return hashMessage(topic + this.clientId);
   }
 
+  private async rpcBatchSubscribe(subscriptions: SubscriberTypes.Params[]) {
+    if (!subscriptions.length) return;
+    const relay = subscriptions[0].relay;
+    const api = getRelayProtocolApi(relay.protocol);
+    const request: RequestArguments<RelayJsonRpc.BatchSubscribeParams> = {
+      method: api.batchSubscribe,
+      params: {
+        topics: subscriptions.map((s) => s.topic),
+      },
+    };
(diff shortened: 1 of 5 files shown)


packages/core/src/controllers/subscriber.ts
```diff
@@ -246,9 +270,12 @@ export class Subscriber extends ISubscriber {
     this.pending.delete(params.topic);
   }
 
-  private onResubscribe(id: string, params: SubscriberTypes.Params) {
-    this.addSubscription(id, { ...params, id });
-    this.pending.delete(params.topic);
+  private onBatchSubscribe(subscriptions: SubscriberTypes.Active[]) {
+    if (!subscriptions.length) return;
+    subscriptions.forEach((subscription) => {
+      this.setSubscription(subscription.id, { ...subscription });
```
bkrem, line 277: Bit worried about this becoming the new bottleneck but on the I/O-bound level instead of network-bound level. 

Let's see what kind of mileage we get out of this for now and then we can still talk about doing the I/O work (`subscriptions.set`/`topicMap.set` etc) in batch form too if necessary.

bkrem (approved):
👑

packages/core/src/controllers/subscriber.ts
ganchoradkov: @bkrem Fair point. Just tested with 3k active subs and performance is degraded _slightly_ on load approx ~1s load time. Also added batching on the batch subscribe for cases where we exceed the 500 max subs payload.

PR #2081, opened 2023-03-09
Add signature verification for cacao

devceline:
# Description

- Add `verifySignature` function to verify cacaos

Does not include `eip1271`. This will be used in chat client until auth team wants to adopt the functions from core.

## How Has This Been Tested?
Has been tested in auth client.

## Due Diligence

- [x] Breaking change
- [x] Requires a documentation update
- [x] Related docs issue/PR (if docs are not included in this PR):
- [x] Requires a e2e/integration test update

## Note
Maybe it'd be worth it to create a sub-package to prevent further enlarging core?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2332,6 +2332,8 @@
     },
     "node_modules/@ethersproject/hash": {
       "version": "5.7.0",
+      "resolved": "https://registry.npmjs.org/@ethersproject/hash/-/hash-5.7.0.tgz",
+      "integrity": "sha512-qX5WrQfnah1EFnO5zJv1v46a8HW0+E5xuBBDTwMFZLuVTx0tbU2kkx15NqdjxecrLGatQN9FGQKpb1FKdHCt+g==",
       "funding": [
         {
           "type": "individual",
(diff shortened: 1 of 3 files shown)


bkrem:
I saw @Elyniss seemed to have already commented this but it's gone (?), why only the partial implementation (without `eip1271`)?

Comparing to the "original": https://github.com/WalletConnect/auth-client-js/blob/main/packages/auth-client/src/utils/signature.ts#L19

> until auth team wants to adopt the functions from core.

Wouldn't be able to replace the existing implementation with this if there's no eip1271 case being handled.

devceline:
> I saw @Elyniss seemed to have already commented this but it's gone (?), why only the partial implementation (without `eip1271`)?
> 
> Comparing to the "original": https://github.com/WalletConnect/auth-client-js/blob/main/packages/auth-client/src/utils/signature.ts#L19
> 
> > until auth team wants to adopt the functions from core.
> 
> Wouldn't be able to replace the existing implementation with this if there's no eip1271 case being handled.

Wanted to see if we wanted to add this in a separate util or keep it in core first before adding that part, but now that it's moving to its own package, I'll add eip1271

devceline:
Closing this PR as a new package will be added to https://github.com/WalletConnect/walletconnect-utils instead.

devceline:
Closing this PR as a new package will be added to https://github.com/WalletConnect/walletconnect-utils instead.

PR #2082, opened 2023-03-09
chore(sign-client): added pairingTopic on session params

### LINKED ISSUE #2074: Add `pairingTopic` to `SessionStruct` to align with latest Sign spec
## Context

* `pairingTopic` was recently added to the existing[ data structure for Session in the specs](https://docs.walletconnect.com/2.0/specs/clients/sign/data-structures#session).
* The JS client's `SessionStruct` type needs to be extended to account for this and the implementation adjusted to also pass `pairingTopic` into newly created sessions going forward.

### Code references

* Definition of the `Session` controller/store: https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/sign-client/src/controllers/session.ts#L7
* Instantiation in the SignClient: https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/sign-client/src/client.ts#L48
* Examples of where `client.session.set` is called:
	* on approve (wallet side): https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/sign-client/src/controllers/engine.ts#L231
	* on `session_connect` (dapp side), i.e. after wallet has approved: https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/sign-client/src/controllers/engine.ts#L140

Cali93:
# Description
Added `pairingTopic` to `SessionStruct` to align with latest Sign spec 

Resolves #2074 

## How Has This Been Tested?
Not tested yet.

## Due Diligence

- [ ] Breaking change


bkrem:
Also adding @ganchoradkov here as the resident SignClient expert

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1,64 +1,64 @@
-import EventEmmiter from "events";
```
ganchoradkov, line 1: was this line removed by mistake? The build seems to be failing due to this

packages/types/src/sign-client/client.ts
```diff
@@ -22,6 +22,7 @@ export declare namespace SignClientTypes {
   interface BaseEventArgs<T = unknown> {
     id: number;
     topic: string;
+    pairingTopic: string;
```
ganchoradkov: The `pairingTopic` should be added to SessionTypes I think, otherwise, all SignClient Event types are failing as they are missing this prop. More logs in the CI build run

packages/sign-client/src/controllers/engine.ts
```diff
@@ -134,6 +134,7 @@ export class Engine extends IEngine {
           session.self.publicKey = publicKey;
           const completeSession = {
             ...session,
+            pairingTopic,
```
bkrem: The variable assignment flow here is unnecessarily confusing, but `pairingTopic` is only an optional parameter for `connect`, which is assigned to `let topic` above.

If `pairingTopic` isn't provided (i.e. we're connecting without a previously known pairing) we create a new one and assign that new topic to `topic` here: https://github.com/WalletConnect/walletconnect-monorepo/pull/2082/files#diff-04e05e19b9270d1267b68d324e16997e4465d6879324fe4427a6194c00410eabR108

Hence `topic` is the source of truth on the final pairing topic here:

https://github.com/WalletConnect/walletconnect-monorepo/pull/2082/files#diff-04e05e19b9270d1267b68d324e16997e4465d6879324fe4427a6194c00410eabR108

```suggestion
            pairingTopic: topic,
```

packages/types/src/sign-client/client.ts
bkrem: Doesn't belong in the Base RPC event args, since this is a generic interface to represent any/all types of RPC events, by passing along the JSON-RPC `id`, the `topic` the event occurred on and some generic payload `T`.

I.e. this interface isn't aware of implementation details like pairings, sessions, etc

```suggestion

```

ganchoradkov (approved):
Awesome 💯

PR #2108, opened 2023-03-15
fix: connection recovery

### LINKED ISSUE #2128: Pairing didn't respect error from the relay
**Describe the bug**
Test wallet with malformed URI for topic part. The relay return (Uncaught) error message, but the pairing still save the pairing.

**SDK Version (if relevant)**
 - Client: JS
 - Version: 2.4.5

**To Reproduce**
1. Go to [walletconnect tab on wallet demo](https://react-wallet.walletconnect.com/walletconnect).
2. Start pairing with uri `wc:testtopic@2?relay-protocol=irn&symKey=12345`
3. See error
4. Goto pairings tab, there's an `Unknown` pairing.
5. Refresh the page will stuck in loading page.

**Expected behavior**
- Caught the uncaught error from the provider request.
- Don't save pairing until the relay says ok.

**Screenshots**

image

image

 - OS: Win10
 - Browser: chrome

ganchoradkov:
# Description
PR to address an issue where sockets don't recover after network downtime/wifi restart etc
Changeset:
- all relay requests piped through `relayer`
- create a new provider instance on dead sockets instead of trying to restart them
- updated `json-rpc-provider`
- retry init/pair of clients in integration test suites

## How Has This Been Tested?
dogfooding
tests
rs-relay ci runs

## Due Diligence

- [ ] Breaking change


packages/core/src/controllers/publisher.ts
```diff
@@ -104,7 +105,7 @@ export class Publisher extends IPublisher {
     if (isUndefined(request.params?.tag)) delete request.params?.tag;
     this.logger.debug(`Outgoing Relay Payload`);
     this.logger.trace({ type: "message", direction: "outgoing", request });
-    return this.relayer.provider.request(request);
+    return this.relayer.request(request);
```
bkrem, line 108: Saw your bullet on this:
> all relay requests piped through relayer

I'm assuming this is simply removing unnecessary layering/indirection for `.request` by taking `provider` out of the equation or is there more to it I'm not seeing here at first glance?

packages/core/src/controllers/relayer.ts
```diff
@@ -168,15 +181,14 @@ export class Relayer extends IRelayer {
...
-
           // wait for the subscriber to finish resubscribing to its topics
           this.subscriber.once(SUBSCRIBER_EVENTS.resubscribed, () => {
             resolve();
           });
         }),
         await Promise.race([
           new Promise<void>(async (resolve) => {
-            await this.provider.connect();
+            await createExpiringPromise(this.provider.connect(), 5_000, "socket hang up");
```
bkrem: Are we shadowing the native/built-in `socket hang up` error with this and muddying the waters between whether the relayer triggered this or whether it came from the socket itself?

Might be less confusing for us to use a differentiated string here like `Provider failed to connect`. Lmk if I'm missing something here.

bkrem: Guessing we need to expand on https://github.com/WalletConnect/walletconnect-monorepo/pull/2108/files#diff-2666bbe86db232da1fe48efc66faddc72893a26741e59bcc7510ea3818bbd581R205 if we want this to throw the way `socket hang up` does but with a different message/cause.

packages/core/src/controllers/subscriber.ts
```diff
@@ -384,6 +384,7 @@ export class Subscriber extends ISubscriber {
   private async batchSubscribe(subscriptions: SubscriberTypes.Params[]) {
     if (!subscriptions.length) return;
     const result = (await this.rpcBatchSubscribe(subscriptions)) as string[];
+    if (!result) return;
```
bkrem: Empty array `[]` is still truthy so I think this is a fall through, better to go by `.length` assuming `rpcBatchSubscribe` will always return an array.
```suggestion
    if (!result.length) return;
```

bkrem (approved):
Thank you for the great rework here both on source and tests 💪 

One minor suggested fix on empty array check and a few contextual questions.

packages/core/src/controllers/publisher.ts
ganchoradkov: Having all requests go through the relayer gives more control such as awaiting the connection to open before submitting the request

packages/core/src/controllers/subscriber.ts
ganchoradkov: If the connection is down i.e. wifi switched off, the batch subscribe will resolve with `undefined` after the request timeouts. I will also add `.length` to be on the safe side 💯

PR #2135, opened 2023-03-22
fix: clients init retry

ganchoradkov:
# Description
Added a check in `initTwoPairedClients` catch statement if the clients are initialized before disconnecting after

## How Has This Been Tested?
dogfooding

## Due Diligence

- [ ] Breaking change


xzilja (approved):
No need to call `initTwoClients` again?

ganchoradkov:
> No need to call `initTwoClients` again?

It is called inside the loop correctly

PR #2137, opened 2023-03-22
feat: caip25 provider handling

ganchoradkov:
# Description
Implemented `caip25` handling in `universal-provider` to accept inline chain declarations e.g. `eip155:1` similarly as in `signClient`
Also fixed an issue where duplicate addresses were not filtered in `getAccounts()`

## How Has This Been Tested?
dogfooding
integration tests

## Due Diligence

- [ ] Breaking change


xzilja (approved):
Looks good, github picked up lint error about non-null assertion though

ganchoradkov:
> Looks good, github picked up lint error about non-null assertion though

There is a check just before the loop for the session but ts is not picking it up. Added another one tho to fix the warning 💯

providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -249,11 +254,25 @@ export class UniversalProvider implements IUniversalProvider {
...
+      throw new Error("Session not initialized. Please call connect() before enable()");
+    }
+
+    const providersToCreate = [
+      ...new Set(
+        Object.keys(this.session.namespaces).map((namespace) => parseNamespaceKey(namespace)),
+      ),
+    ];
+    providersToCreate.forEach((namespace) => {
+      if (!this.session) return;
```
bkrem, line 267: Saw the prior non null warning. Weird that this is still needed as an explicit check for TS to pick up that session is defined if you already have the `if (!this.session) ...` above?

providers/universal-provider/src/providers/cardano.ts
```diff
@@ -77,13 +77,15 @@ class CardanoProvider implements IProvider {
...
-        // remove namespace & chainId from the string
-        .map((account) => account.split(":")[2]) || []
-    );
+    return [
+      ...new Set(
+        accounts
+          // get the accounts from the active chain
+          .filter((account) => account.split(":")[1] === this.chainId.toString())
+          // remove namespace & chainId from the string
+          .map((account) => account.split(":")[2]) || [],
```
bkrem: Non-blocking: Realise you're just wrapping with `Set` here, but if `accounts` is always an array the `|| []` is redundant, since functional operations like filter/map will always yield at least an empty array back

bkrem (approved):

providers/universal-provider/src/UniversalProvider.ts
ganchoradkov: Yeah, seems redundant but at least ts is not raising warnings so no biggie

PR #2155, opened 2023-03-28
feat(core): adds `getPayloadSenderPublicKey` method to `core.crypto`

bkrem:
# Description
- Adds `core.crypto.getPayloadSenderPublicKey` method to return only the encoded payload's `senderPublicKey` (if present), similarly to what `getPayloadType` already does.
- This is necessary in Push (and probably useful in Chat where `TYPE_1` envelope messages are also used), as the response for `wc_pushRequest` no longer contains the `senderPublicKey` (see change [here](https://github.com/WalletConnect/walletconnect-docs/pull/502/files#diff-13680fc4bd48de744b4ba1a6debcf4f699f65a2abfcf54669e9d29cf527235aeR50)), leaving extraction from the `TYPE_1` envelope encoding as the only option.
- Also: formalises type reference for existing `getPayloadType`

## How Has This Been Tested?

- Pre-released as `core@2.5.2-4198a778` canary
- Integrated in PushClient via above canary and tested via unit tests + dogfooding in push dapp/wallet samples

## Due Diligence

- [ ] Breaking change

### DIFF
--- a/packages/core/src/controllers/crypto.ts
+++ b/packages/core/src/controllers/crypto.ts
@@ -16,7 +16,10 @@ import {
   isTypeOneEnvelope,
   deserialize,
   decodeTypeByte,
+  BASE16,
 } from "@walletconnect/utils";
+import { toString } from "uint8arrays";
+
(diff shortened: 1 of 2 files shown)


ganchoradkov (approved):
💯

PR #2161, opened 2023-03-29
feat: auto namespaces

ganchoradkov:
# Description
Implemented util that builds `namespaces` structure by provided supported `chains, methods, events & accounts`. 
interface
```
export function buildApprovedNamespaces(params: {
  requiredNamespaces: ProposalTypes.RequiredNamespaces;
  optionalNamespaces?: ProposalTypes.OptionalNamespaces;
  supportedNamespaces: Record<
    string,
    { chains: string[]; methods: string[]; events: string[]; accounts: string[] }
  >
}) 
```
It takes `required` & `optional` namespaces provided in the `session proposal` along with the supported chains/methods/events/accounts by the wallet and returns a structured `namespaces` object

## How Has This Been Tested?
multiple tests with different configuration

## Due Diligence

- [ ] Breaking change

### DIFF
--- a/packages/utils/src/namespaces.ts
+++ b/packages/utils/src/namespaces.ts
@@ -65,14 +65,25 @@ export function getRequiredNamespacesFromNamespaces(
   return required;
 }
 
-export function buildApprovedNamespaces(params: {
+export type BuildApprovedNamespacesParams = {
   requiredNamespaces: ProposalTypes.RequiredNamespaces;
   optionalNamespaces?: ProposalTypes.OptionalNamespaces;
   supportedNamespaces: Record<
     string,
     { chains: string[]; methods: string[]; events: string[]; accounts: string[] }
   >;
-}): SessionTypes.Namespaces {
+};
+
+/**
+ * util designed for Wallets that builds namespaces structure by provided supported chains, methods, events & accounts.
+ * It takes required & optional namespaces provided in the session proposal
+ * along with the supported chains/methods/events/accounts by the wallet and returns a structured namespaces object
+ * @param {BuildApprovedNamespacesParams} params
+ * @returns {SessionTypes.Namespaces}
+ */
+export function buildApprovedNamespaces(
+  params: BuildApprovedNamespacesParams,
+): SessionTypes.Namespaces {
   const { requiredNamespaces, optionalNamespaces = {}, supportedNamespaces } = params;
 
   const normalizedRequired = normalizeNamespaces(requiredNamespaces);
--- a/packages/utils/src/namespaces.ts
+++ b/packages/utils/src/namespaces.ts
@@ -169,18 +169,21 @@ export function parseNamespaceKey(namespace: string) {
 
 /**
  * Converts
+ * ```
  * {
  *  "eip155:1": {...},
  *  "eip155:2": {...},
  * }
+ * ```
  * into
+ * ```
  * {
  *  "eip155": {
  *      chains: ["eip155:1", "eip155:2"],
  *      ...
  *    }
  * }
- *
+ *```
  */
 export function normalizeNamespaces(
   namespaces: ProposalTypes.RequiredNamespaces,
--- a/packages/utils/src/namespaces.ts
+++ b/packages/utils/src/namespaces.ts
@@ -70,7 +70,7 @@ export function buildApprovedNamespaces(params: {
   optionalNamespaces?: ProposalTypes.OptionalNamespaces;
   supportedNamespaces: Record<
     string,
-    { chains?: string[]; methods: string[]; events: string[]; accounts: string[] }
+    { chains: string[]; methods: string[]; events: string[]; accounts: string[] }
   >;
 }): SessionTypes.Namespaces {
   const { requiredNamespaces, optionalNamespaces = {}, supportedNamespaces } = params;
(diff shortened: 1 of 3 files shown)


xzilja (approved):
This is a really good idea and a nice utility! 🚀

packages/utils/src/namespaces.ts
```diff
@@ -63,3 +64,140 @@ export function getRequiredNamespacesFromNamespaces(
   }
   return required;
 }
+
+export function buildApprovedNamespaces(params: {
```
bkrem: Since this is user facing should we add the context you put in the PR description here too as a JSDoc and point out it’s relevant for wallets specifically?

packages/utils/src/namespaces.ts
```diff
@@ -63,3 +64,140 @@ export function getRequiredNamespacesFromNamespaces(
...
+    { chains: string[]; methods: string[]; events: string[]; accounts: string[] }
+  >;
+}): SessionTypes.Namespaces {
+  const { requiredNamespaces, optionalNamespaces = {}, supportedNamespaces } = params;
+
+  const normalizedRequired = normalizeNamespaces(requiredNamespaces);
+  const normalizedOptional = normalizeNamespaces(optionalNamespaces);
+
+  const namespaces = {};
+  // build approved namespaces
```
bkrem, line 92: V minor: comment placement a bit odd here

packages/utils/test/validators.spec.ts
```diff
@@ -328,3 +328,914 @@ describe("Validators", () => {
...
+      cosmos: {
+        chains: chainsCosmos,
+        methods: methodsCosmos,
+        events: eventsCosmos,
+        accounts: accountsCosmos,
+      },
+    };
+    expect(approvedNamespaces).to.deep.eq(expected);
+  });
+  it.fails(
```
bkrem, line 930: TIL about there being an explicit `fails` method in vitest.

If we’re expecting specific error returns is there a way to check that to avoid false positives on these? Eg these throw but they actually all throw for the same reason due to a bug etc

packages/utils/test/validators.spec.ts
```diff
@@ -328,3 +328,914 @@ describe("Validators", () => {
...
+            methods,
+            events,
+            accounts: accountsEip,
+          },
+        },
+      });
+    },
+  );
+  it.fails(
+    "should throw while building namespaces (config 6 - partiall accounts for required chain)",
```
bkrem: Typo fix:

```suggestion
    "should throw while building namespaces (config 6 - partial accounts for required chain)",
```

packages/utils/src/namespaces.ts
```diff
@@ -63,3 +64,143 @@ export function getRequiredNamespacesFromNamespaces(
...
+  return namespace.includes(":");
+}
+
+export function parseNamespaceKey(namespace: string) {
+  return isCaipNamespace(namespace) ? namespace.split(":")[0] : namespace;
+}
+
+/**
+ * Converts
+ * ```
```
bkrem, line 185: Took the liberty of pushing a tiny change for these backticks so editors pick up the code blocks correctly:

Screenshot 2023-03-29 at 21 21 30

packages/utils/src/namespaces.ts
```diff
@@ -63,3 +64,143 @@ export function getRequiredNamespacesFromNamespaces(
...
+    );
+    const events = supportedNamespaces[requiredNamespace].events.filter((event) =>
+      normalizedRequired[requiredNamespace]?.events?.includes(event),
+    );
+
+    approvedNamespaces[requiredNamespace] = {
+      chains,
+      methods,
+      events,
+      accounts: chains
```
bkrem: Minor/style: is there a reason the `accounts` key is eval'd inline rather than before assignment like the others above?

packages/utils/test/validators.spec.ts
bkrem: Actually maybe overkill since the only cause of throwing is `isConformingNamespaces` which has its own tests.

bkrem (approved):
Epic 💯 

Often discussed but never tackled until now, especially with the tests and due diligence to back it up 💪

antondalgren:
I would like to see a test-case for this request: 
```typescript
    const required = {
      "eip155:1": {
        events: ["chainChanged"],
        methods: ["personal_sign", "eth_sendTransaction"],
      },
      "eip155:2": {
        events: ["chainChanged"],
        methods: ["personal_sign", "eth_sendTransaction"],
      },
    };
    const optional = {
      eip155: {
        chains: ["eip155:2"],
        events: ["chainChanged", "accountsChanged"],
        methods: [
          "personal_sign",
          "eth_sendTransaction",
          "eth_signTransaction",
          "eth_signTypedData",
        ],
      },
    };
```

I'd expect it to be 
```jsonc
{
   "eip155:1": {
        "events": ["chainChanged"],
        "methods": ["personal_sign", "eth_sendTransaction"],
        "accounts": ["eip155:1:0x...."]
    },
   "eip155:2": {
        "accounts": ["eip155:2:0x...."],
        "events": ["chainChanged", "accountsChanged"],
        "methods": [
          "personal_sign",
          "eth_sendTransaction",
          "eth_signTransaction",
          "eth_signTypedData"
        ]
      }
}
```

antondalgren:
Maybe the above is not what you are trying to achieve with this function. 🤔

ganchoradkov:
Hey @antondalgren, the above functionality tries to simplify the Wallet experience with parsing/validating required/optional namespaces. 
From our tests with dapps/wallets that already implemented v2, 99% do not provide such fine tuning so we've covered the vast majority of use cases in this first iteration.

Your test case is valid and I will experiment with how to preserve the fine details but also provide easy-to-parse final result

packages/utils/src/namespaces.ts
ganchoradkov: No reason, all props were inlined initially but after a few iterations the accounts were abandoned alone there

antondalgren:
Awesome @ganchoradkov! Just wanted to be sure that the granularity that is possible to achieve in V2 is actually preserved. I'm sure the majority of the dapps wouldn't implement this granularity, but for those that do, I believe that the tooling that majority wallet/dapps will use also have to support for the granularity options. Otherwise, we will end up with a community where you will not able to have the granularity that is achievable in the protocol.

PR #2175, opened 2023-04-03
Add Polkadot provider

marshacb:
# Description

This adds polkadot support to the UniversalProvider by creating a Polkadot http provider to ensure polkadot dApps can use WalletConnect relay servers successfully.

# How Has This Been Tested
Tested generating a modal / session ID

## Due Diligence

- [ ] Breaking change


providers/universal-provider/src/providers/polkadot.ts
```diff
@@ -0,0 +1,127 @@
...
+        .filter((account) => account.split(":")[1] === this.chainId.toString())
+        // remove namespace & chainId from the string
+        .map((account) => account.split(":")[2]) || []
+    );
+  }
+
+  private createHttpProviders(): RpcProvidersMap {
+    const http = {};
+    this.namespace.chains.forEach((chain) => {
+      http[chain] = this.createHttpProvider(chain);
```
ganchoradkov: In the `this.createHttpProvider` we should also pass the `rpcMap` url for that chain ID as in the other sub providers
```suggestion
      http[chain] = this.createHttpProvider(chain, this.namespace.rpcMap?.[chain]);
```

ganchoradkov (changes requested):
@marshacb  thank you for creating `Polkadot` sub-provider!
One small additional update is required and we are good to merge 💯 

edit: I'm seeing `prettier` is failling in the CI, `npm run prettier` should solve this

providers/universal-provider/src/providers/polkadot.ts
marshacb: thanks! I've updated this :)

ganchoradkov (approved):

PR #2182, opened 2023-04-05
fix(engine): handle payload that is not identifiable request/response

bkrem:
# Description

## Context
- Discovered while debugging x-platform for OpenSea (Kotlin dapp <-> JS wallet), where the incoming `eth_sendTransaction` request would simply halt in the SignClient without any error or response because the decoded payload was stringified instead of an expected request/response JSON-RPC object. See [here for full context](https://walletconnect.slack.com/archives/C02FJDVF6MS/p1680517342916879?thread_ts=1678981020.319979&cid=C02FJDVF6MS)
- Underlying issue is that if the engine decodes a payload that neither matches `isJsonRpcRequest` nor `isJsonResponse`, handling for the base case (i.e. unknown/unidentifiable payload) is currently missing.
- On the JS <-> JS client level this is forbidden by pre-flight validation on `client.request()`, but as shown in this case we can't assume that a malformed payload will never be received, and need to handle the case accordingly

## Solution

- Handle unidentifiable `payload` by logging error and returning a JSON-RPC Error response to the peer.

## Due Diligence

- [ ] Breaking change


JonathanConn (approved):
🚀

ganchoradkov (approved):

PR #2190, opened 2023-04-07
chore(generic-storage): add a generic storage

Cali93:
# Description

This adds a generic storage meaning that it can be used to store arbitrary data.
The first use case is for a util that will provide a shared storage between push and chat clients.

Related to [this issue](https://github.com/WalletConnect/walletconnect-utils/issues/79)
## How Has This Been Tested?

Current tests already cover such a functionality as we are just exposing an additional `KeyValueStorage`

## Due Diligence

- [ ] Breaking change


bkrem:
@Cali93 going to close this since we found a cleaner solution that supersedes this

PR #2199, opened 2023-04-11
fix: wrapps try-catch around `verify-api`'s resolve fx

ganchoradkov:
# Description
Added try-catch around `resolve` to avoid cases of invalid verify urls or rejected requests

## Due Diligence

- [ ] Breaking change


bkrem (approved):
Good call

PR #2327, opened 2023-05-09
feat: duplicate pairings cleanup

ganchoradkov:
# Description
Implemented duplicate pairings cleanup that removes old pairings when new one is created with the same peer.

## How Has This Been Tested?
dogfooding
tests

## Due Diligence

- [ ] Breaking change

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -361,7 +361,10 @@ export class Engine extends IEngine {
       const pairing = this.client.core.pairing.pairings.get(session.pairingTopic);
       const allPairings = this.client.core.pairing.pairings.getAll();
       const duplicates = allPairings.filter(
-        (p) => p.peerMetadata?.url === session.self.metadata.url && p.topic !== pairing.topic,
+        (p) =>
+          p.peerMetadata?.url &&
+          p.peerMetadata?.url === session.self.metadata.url &&
+          p.topic !== pairing.topic,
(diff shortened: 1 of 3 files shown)


packages/sign-client/src/controllers/engine.ts
```diff
@@ -354,6 +354,29 @@ export class Engine extends IEngine {
...
+      const duplicates = allPairings.filter(
+        (p) =>
+          p.peerMetadata?.url &&
+          p.peerMetadata?.url === session.self.metadata.url &&
+          p.topic !== pairing.topic,
+      );
+      if (duplicates.length === 0) return;
+      this.client.logger.info(`Cleaning up ${duplicates.length} duplicate pairing(s)`);
+      await Promise.all(
+        duplicates.map((p) => this.client.core.pairing.disconnect({ topic: p.topic })),
```
bkrem, line 372: Don’t think this is a huge concern with real world usage, but this could trigger a lot of `wc_pairingDelete` requests in quick succession where pairings with the same peer have accumulated.

bkrem (approved):
LGTM 💪

packages/sign-client/src/controllers/engine.ts
ganchoradkov: I've tested with ~20 active pairings and it was working just fine. After the initial cleanup - there will be 1 pairing to be removed only

JonathanConn (approved):

PR #2353, opened 2023-05-11
refactor: publisher queue

ganchoradkov:
# Description
- Reworked the publisher `queue` to remove items on relay ACK instead of when the request is responded to
- Removed unused deps from a few packages

## How Has This Been Tested?
Integration tests
dogfooding
canary `2.7.4-rc-a7ad32`

## Due Diligence

- [ ] Breaking change


packages/core/src/constants/relayer.ts
```diff
@@ -10,6 +10,7 @@ export const RELAYER_CONTEXT = "relayer";
 
 export const RELAYER_EVENTS = {
   message: "relayer_message",
+  message_ack: "message_ack",
```
bkrem: Nit: this value should be `relayer_` prefixed for the sake of consistency with the other values in the mapping.

packages/core/test/publisher.spec.ts
```diff
@@ -23,24 +28,130 @@ describe("Publisher", () => {
...
+      // Emit ACKs
+      items.forEach((item) =>
+        core.relayer.events.emit(RELAYER_EVENTS.message_ack, { id: item.opts.id }),
+      );
+
+      // -> Queue should clear after the ACKs.
+      expect(publisher.queue.size).to.equal(0);
+
+      // emit multiple pulses to ensure queue is cleared
+      await throttle(100);
```
bkrem: Why throttle before the first heartbeat pulse here rather than after?

bkrem (approved):
Nice, great to see the publisher/queue getting some love 💯 

This feels like a definite candidate for wider pre-release canary testing (e.g. in web3modal lab). Do you think it makes sense to release the smaller patches that were merged earlier separately and then do a release for this?

packages/core/test/publisher.spec.ts
ganchoradkov: left over from different iterations of the test. Its redundant and I removed it 💯

PR #2364, opened 2023-05-12
providers: universal-provider: fix wrong return type

### LINKED ISSUE #2363: Wrong return type of universal provider's disconnect function
**Describe the bug**
The universal provider's disconnect function has a wrong return type

https://github.com/WalletConnect/walletconnect-monorepo/blob/d0cf3999f294e3d02b6e369fa90761491766a39c/providers/universal-provider/src/types/providers.ts#LL42C10-L42C10

**Expected behavior**
The disconnect function should return a promise

Mostafatalaat770:
# Description
The disconnect function in the universal provider is async while in the interface it implements doesn't return a promise.

fixes #2363


bkrem (approved):
Looks like a good catch to me, thank you 🙏 will wait for @ganchoradkov for final confirmation/merge

ganchoradkov (approved):
Thanks for the catch & fix @Mostafatalaat770 💯

PR #2375, opened 2023-05-14
fix: delayed engine initialization in sign unless active sessions to …

JonathanConn:
PoC for optimizing sign client (NOT READY FOR MERGE)

# Description

## Due Diligence

- [ ] Breaking change

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,6 @@
-## Description
+> Creating a release? Please use the [Release PR Template](https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/.github/pr_template_release.md) instead.
 
-> Creating a release PR? Please use the [Release PR Template](./pr_template_release.md) instead.
+## Description
 
 Please include a brief summary of the change.
 


packages/sign-client/src/client.ts
```diff
@@ -85,6 +87,9 @@ export class SignClient extends ISignClient {
 
   public connect: ISignClient["connect"] = async (params) => {
     try {
+      if (!this.engineInitialized) {
```
arein: nit: maybe a comment here for posterity

Or over the variable name

packages/sign-client/src/client.ts
```diff
@@ -209,7 +214,12 @@ export class SignClient extends ISignClient {
       await this.session.init();
       await this.proposal.init();
       await this.pendingRequest.init();
-      await this.engine.init();
+
+      if (this.session.getAll().length > 0) {
```
arein: What does this mean?
nit: move into variable with a name to depict why you're checking this

arein (commented):
Nice! 

I know this is a draft
But how can this be tested?
Both manually and maybe in integ Tests?
Might be too hard
Just curious

packages/sign-client/src/client.ts
JonathanConn: I'll move it to a var so its more clear. When `session` is initialized, it loads the current sessions in, using `restore()`. I am checking to see if there are any active sessions

JonathanConn:
Need to consider 
- `find()` and `getPendingSessionRequests()` in client because they are not async
- testing for this, ideally with performance to create a before/after comparison

xzilja (commented):
LGTM 👍 will defer to Gancho for final approve

packages/sign-client/src/client.ts
```diff
@@ -218,4 +239,17 @@ export class SignClient extends ISignClient {
       throw error;
     }
   }
+
+  // if engine was not initialized during client initialization, initialize now
+  private async initEngine() {
+    if (!this.engineInitialized) {
```
arein: nitpick: this allows for race conditions. Not sure if it matters?

ganchoradkov (changes requested):
I would use a different approach instead of delaying the engine init, let me explain in private

PR #2400, opened 2023-05-16
fix: relay optimization 

JonathanConn:
## Description
Stop the relay socket after init if not in use (no active subscribed topics)
10 second delay on stop

## Type of change

## How has this been tested?
Two tests were added to the relay test suite

## Checklist

### DIFF
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -40,6 +40,7 @@ import {
   RELAYER_SUBSCRIBER_SUFFIX,
   RELAYER_DEFAULT_RELAY_URL,
   SUBSCRIBER_EVENTS,
+  RELAYER_TRANSPORT_CUTOFF,
 } from "../constants";
 import { MessageTracker } from "./messages";
 import { Publisher } from "./publisher";
@@ -95,7 +96,7 @@ export class Relayer extends IRelayer {
         await this.transportClose();
         this.transportExplicitlyClosed = false;
       }
-    }, 10_000);
+    }, RELAYER_TRANSPORT_CUTOFF);
   }
 
   get context() {
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -89,6 +89,13 @@ export class Relayer extends IRelayer {
     await Promise.all([this.messages.init(), this.transportOpen(), this.subscriber.init()]);
     this.registerEventListeners();
     this.initialized = true;
+    setTimeout(async () => {
+      if (this.subscriber.topics.length === 0) {
+        this.logger.info(`No topics subscribted to after init, closing transport`);
+        await this.transportClose();
+        this.transportExplicitlyClosed = false;
+      }
+    }, 10_000);
   }
 
   get context() {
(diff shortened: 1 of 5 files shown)


packages/core/src/controllers/relayer.ts
```diff
@@ -89,6 +89,13 @@ export class Relayer extends IRelayer {
     await Promise.all([this.messages.init(), this.transportOpen(), this.subscriber.init()]);
     this.registerEventListeners();
     this.initialized = true;
+    setTimeout(async () => {
+      if (this.subscriber.topics.length === 0) {
+        this.logger.info(`No topics subscribted to after init, closing transport`);
+        await this.transportClose();
+        this.transportExplicitlyClosed = false;
+      }
+    }, 10_000);
```
```suggestion
    }, RELAYER_TRANSPORT_CUTOFF);
```
could we create a `const` in the constants/relayer.ts like
`export const RELAYER_TRANSPORT_CUTOFF = 10_000`
and import it into the relayer and the tests?
It would solve the need to update the value in multiple places when it gets changed

packages/core/test/relayer.spec.ts
```diff
@@ -259,4 +262,37 @@ describe("Relayer", () => {
...
+
+    it("should close transport 10 seconds after init if NOT active", async () => {
+      relayer = new Relayer({
+        core,
+        logger,
+        relayUrl: TEST_CORE_OPTIONS.relayUrl,
+        projectId: TEST_CORE_OPTIONS.projectId,
+      });
+      await relayer.init();
+      await throttle(11_000); // +1 sec buffer
```
ganchoradkov:
```suggestion
      await throttle(RELAYER_TRANSPORT_CUTOFF + 1_000); // +1 sec buffer
```

packages/core/test/relayer.spec.ts
```diff
@@ -259,4 +262,37 @@ describe("Relayer", () => {
...
+        core,
+        logger,
+        relayUrl: TEST_CORE_OPTIONS.relayUrl,
+        projectId: TEST_CORE_OPTIONS.projectId,
+      });
+      const topic = "fake_topic";
+      await relayer.subscriber.init();
+      await relayer.subscriber.subscribe(topic);
+      await relayer.init();
+      await throttle(11_000); // +1 sec buffer
```
ganchoradkov:
```suggestion
      await throttle(RELAYER_TRANSPORT_CUTOFF + 1_000); // +1 sec buffer
```

ganchoradkov (changes requested):
Awesome! Small comment to improve the constants usage

arein (commented):
Observation: the original goal was for the socket to be only opened if it's actually needed eg lazy. Ilja Said this May have to be done on the Web3modal layer.

Looking at the data most of the Zombie sockets actually create a pairing and hence are subscribed to topics

So not sure if this change will work?

ganchoradkov:
>Observation: the original goal was for the socket to be only opened if it's actually needed eg lazy. Ilja Said this May have to be done on the Web3modal layer.
Looking at the data most of the Zombie sockets actually create a pairing and hence are subscribed to topics
So not sure if this change will work?

if even it's handled on the web3modal level (I think it shouldn't), people clicking `Connect` and leaving it there would be the same behavior as now. Lets investigate/test the different scenarios and go from there on the improvements

packages/core/src/controllers/relayer.ts
JonathanConn: Great suggestion, implemented

ganchoradkov (approved):
Awsome! 🥇

PR #2541, opened 2023-06-03
feat: add `eth_signTypedData_v4` to required methods

jxom:
Should `eth_signTypedData_v4` should be required?


pedrouid:
This is just going to break more connections than it will fix them

Our Wallet partners were given a NamespaceBuilder which will auto-resolve namespaces based on their supported methods overlapping with the proposed namespaces by the dapp

Thus if the wallet does support the method it would be included in optional namespaces... so making it required is just going to cause it to auto-reject instead

We need to identify the wallets that are causing issues with eth_signTypedData_v4 and assist them to re-configure their NamespaceBuilder properly

jxom:
Yeah, I guess this PR was more to cater for the out-of-the-box scenario where folks do not pass optional methods to ethereum-provider (unrelated to wagmi). A lot of DApps use eth_signTypedData, and if we are recommending DApps and libraries to switch to v2 WITHOUT including the optional methods (which I can see in the docs and many screenshots on Twitter), this will effectively break the DApp.

Maybe this should be made clear on the docs.

PR #2559, opened 2023-06-06
chore: move request deeplink handling to client level

xzilja:

## Description

- We were previously handling deeplink redirects on provider level, which meant that packages like `@web3modal/sign-react` or anyone using pure `sign-client` didn't benefit from this, thus deeplink handling was moved to sign client engine.

- In addition to above, I pulled this repo "fresh" and installed everything via npm@8, however it made some .lock file changes and I had to add extra dev packages like `esbuild` and `eslint-plugin-n` for everything to work. I believe it is due to some updates in rollup and eslint-standard deps that happened in .lock file where these were removed as devDependencies and made peerDeps.

- Added `update package-lock` step to github actions, it checks package.json and ensures associated lockfile is up to date, without installing anything

- Updated node target in github actions from 16.x to 18.x (also everything seems to work locally for me on node 20.x)

## Type of change

## How has this been tested?

A canary was cut for sign-client and applied in web3modal lab sign example: https://4ec50c56.web3modal-dev.pages.dev/with-sign-api/html

Then tested on mobile, with all redirects working as expected

## Fixes/Resolves (Optional)

N/A

N/A

## Checklist

N/A


packages/utils/src/misc.ts
```diff
@@ -321,3 +322,43 @@ export function engineEvent(event: EngineTypes.Event, id?: number | string | und
...
 }
+
+export async function handleDeeplinkRedirect({
+  id,
+  topic,
+  store,
+}: {
+  id: number;
+  topic: string;
+  store: IKeyValueStorage;
```
ganchoradkov:
```suggestion
  deepLink: string;
```
I think if we're to make this general util, that others can use too, might be better to pass the `deepLink` directly instead of `store`. This way, the util would work with any deep link and not be locked for `WALLETCONNECT_DEEPLINK_CHOICE` only. Wdyt?

package-lock.json
ganchoradkov, line 1: Any idea why the lock got restructured that much? 
May be the `"sideEffects": false`?

ganchoradkov (approved):
Awesome! 💯 
Just have to revert `2.7.9-7fa58fba` to `2.7.8`

PR #2571, opened 2023-06-07
fix: disconnect exceptions

### LINKED ISSUE #1772: Crash when deleting wallet connect session in wallet
**Describe the bug**
When I connect a dApp with the [Wallet Connect React Wallet](https://react-wallet.walletconnect.com) and delete the session in the wallet (in the sessions page), the dApp crashes. The call stack is:
```
Uncaught (in promise) Error: No matching key. keychain: 5aaaa520505c6774e5c3c7355f9a577caebe6b79aeb133af252eb7a4f1073a76
    at lt.get (index.es.js:1:16104)
    at dt.getSymKey (index.es.js:1:18645)
    at dt.encode (index.es.js:1:17891)
    at je.sendError (index.es.js:1:9128)
    at je.onSessionDeleteRequest (index.es.js:1:14521)
```

I have an event subscriber on my UniversalProvider (`_provider` in the code below) that never gets called:
```
// Subscribe to session delete
    _provider.on(
      'session_delete',
      ({ id, topic }: { id: number; topic: string }) => {
        console.log('session_deleted', id, topic);
      }
    );
```

The only place that the session is getting deleted is in the React wallet, not the dApp.

**SDK Version (if relevant)**
 - Client: TypeScript
 - SDK:
   - "@web3modal/standalone": "^2.0.0-beta.9"
   - "@walletconnect/universal-provider": "^2.1.4"

**To Reproduce**
1. Initializing a UniversalProvider
2. Calling connect
3. Initializing an instance of Web3Modal
4. `openModal` on the Web3Modal
5. Copy+Paste URI into [Wallet Connect React Wallet](https://react-wallet.walletconnect.com)
6. Successfully connect wallet to dApp
7. Delete session in the sessions page in wallet
8. Crash

**Expected behavior**
The session should be deleted and the `'session_delete'` handler should be called.

**Screenshots**
Screen Shot 2022-12-15 at 4 46 39 PM

 - OS: macOS
 - Browser: Chrome

ganchoradkov:
## Description
In rare cases when session disconnect is initiated from both peers could result in an invalid state where the session is active while keychain has been deleted. This blocks the client from deleting the session via `disconnect` method.

- Added checks for each data type before deleting as if such an invalid state is reached, the session is unusable anyway so we shouldn't prevent deletion.
- Removed redundant error response on session disconnect
- Added an await for relay ACK on the disconnect request before deleting the respective data
- Added optional ID parameter to `sendRequest`

## Type of change

## How has this been tested?
tests already covering disconnect functionality
dogfooding in example apps

## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -390,12 +390,14 @@ export class Engine extends IEngine {
     const { self } = this.client.session.get(topic);
     // Await the unsubscribe first to avoid deleting the symKey too early below.
     await this.client.core.relayer.unsubscribe(topic);
-    await Promise.all([
-      this.client.session.delete(topic, getSdkError("USER_DISCONNECTED")),
-      this.client.core.crypto.deleteKeyPair(self.publicKey),
-      this.client.core.crypto.deleteSymKey(topic),
-      expirerHasDeleted ? Promise.resolve() : this.client.core.expirer.del(topic),
-    ]);
+    this.client.session.delete(topic, getSdkError("USER_DISCONNECTED"));
+    if (this.client.core.crypto.keychain.has(topic)) {
+      await this.client.core.crypto.deleteKeyPair(self.publicKey);
+    }
+    if (this.client.core.crypto.keychain.has(topic)) {
+      await this.client.core.crypto.deleteSymKey(topic);
+    }
+    if (!expirerHasDeleted) this.client.core.expirer.del(topic);
   };
 
   private deleteProposal: EnginePrivate["deleteProposal"] = async (id, expirerHasDeleted) => {
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -801,7 +801,6 @@ export class Engine extends IEngine {
       ]);
       this.client.events.emit("session_delete", { id, topic });
     } catch (err: any) {
-      await this.sendError(id, topic, err);
       this.client.logger.error(err);
     }
   };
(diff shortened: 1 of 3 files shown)


packages/sign-client/src/controllers/engine.ts
```diff
@@ -390,12 +416,14 @@ export class Engine extends IEngine {
...
-      this.client.session.delete(topic, getSdkError("USER_DISCONNECTED")),
-      this.client.core.crypto.deleteKeyPair(self.publicKey),
-      this.client.core.crypto.deleteSymKey(topic),
-      expirerHasDeleted ? Promise.resolve() : this.client.core.expirer.del(topic),
-    ]);
+    this.client.session.delete(topic, getSdkError("USER_DISCONNECTED"));
+    if (this.client.core.crypto.keychain.has(topic)) {
+      await this.client.core.crypto.deleteKeyPair(self.publicKey);
+    }
+    if (this.client.core.crypto.keychain.has(topic)) {
```
xzilja, line 423: If conditions here are the same, is it intentional to have them separated this way or perhaps `await this.client.core.crypto.deleteSymKey(topic);` can be moved right after `await this.client.core.crypto.deleteKeyPair(self.publicKey);` bit?

xzilja (approved):
Looks good, left one comment / question

packages/sign-client/src/controllers/engine.ts
ganchoradkov: Good catch. Also added tests for the keychain 💯

JonathanConn (approved):
LGTM 🚀
Good tests

Issue #2392, opened 2023-05-16
Error: closeTransport called before connection was established


ChenHaoJie9527:
### Link to minimal reproducible example

Initialize web3Modal with an error!

### Summary

![image](https://github.com/WalletConnect/web3modal/assets/68054968/c04f0771-50fb-41b6-af2e-5eb794edfad8)

### List of related npm package versions

import {
  EthereumClient,
  w3mConnectors,
  w3mProvider,
} from "@web3modal/ethereum";
import { Web3Modal } from "@web3modal/html";
import { configureChains, createConfig, mainnet } from "@wagmi/core";
// import { arbitrum, mainnet, polygon } from "@wagmi/core/chains";

// const chains = [arbitrum, mainnet, polygon];
const projectId = "349dd70c82d590a2a01a5e31da8ca2b2";

const { publicClient, chains, webSocketPublicClient } = configureChains(
  [mainnet],
  [w3mProvider({ projectId })]
);
const wagmiConfig = createConfig({
  autoConnect: true,
  connectors: w3mConnectors({ projectId, version: 2, chains }),
  publicClient,
  webSocketPublicClient,
});
const ethereumClient = new EthereumClient(wagmiConfig, chains);
export const web3modal = new Web3Modal({ projectId }, ethereumClient);

xzilja:
Will transfer this to our other repository, where it is being addressed, ty for report

ganchoradkov:
Hey @ChenHaoJie9527, are you using some kind of VPN? 
Which region are you trying to connect from?

qqqzhch:
I also encountered this problem
```
const initializeWalletConnectClient = async () => {
      const core = new Core({
        projectId: WALLETCONNECT_V2_PROJECT_ID
      })

      const web3wallet = await Web3Wallet.init({
        core,
        metadata: Vault_METADATA
      })

      setWeb3wallet(web3wallet)
    }
```
Turning vpn on or off will cause this problem
![image](https://github.com/WalletConnect/walletconnect-monorepo/assets/3917994/7bc21be4-efdb-44db-88f0-3b2fe3e33463)

ganchoradkov:
>Turning vpn on or off will cause this problem

Does the connection recover by itself as it should?
Turning on/off the vpn might kill the socket connection silently without emitting disconnect events thus leading to the reconnection issue

ylmin:
I also encountered this problem

AdamK222:
[...]

PR #2644, opened 2023-06-15
chore(relayer): clarify error message for `rejectTransportOpen`

bkrem:

## Description

- Relates to #2392 

## Type of change

## How has this been tested?

- Unit tests

## Checklist

### DIFF
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -253,7 +253,7 @@ export class Relayer extends IRelayer {
   }
 
   private rejectTransportOpen() {
-    throw new Error("closeTransport called before connection was established");
+    throw new Error("Attempt to connect to relay via `transportOpen` has stalled.");
   }
 
   private async createProvider() {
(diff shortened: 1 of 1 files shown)


arein (commented):
Ideally the error message tells the user what to do. I assume we have a retry policy. So if the retry policy can catch this error and then let the user know when the next retry is that would be better

xzilja (approved):

packages/core/src/controllers/relayer.ts
```diff
@@ -253,7 +253,7 @@ export class Relayer extends IRelayer {
   }
 
   private rejectTransportOpen() {
-    throw new Error("closeTransport called before connection was established");
+    throw new Error("Attempt to connect to relay via `transportOpen` has stalled.");
```
```suggestion
    throw new Error("Attempt to connect to relay via `transportOpen` has stalled. Retrying...");
```

As Derek suggested, should we mention that auto retry is active?

ganchoradkov (approved):

PR #2647, opened 2023-06-15
remove deprecated eip1193-provider dependency

pedrouid:
This dependency is deprecated and it should be removed from universal-provider

NPM: https://www.npmjs.com/package/eip1193-provider


ganchoradkov (approved):
💯

PR #2781, opened 2023-06-26
feat:  providers to accept empty required chains/namespaces

ganchoradkov:
## Description
Implemented the following updates
- `universal-provider` -> to accept empty `namespaces` & `optionalNamespaces`
- `ethereum-provider` -> to accept empty `chains` OR `optionalChains` i.e. one of the params must be defined

## Type of change

## How has this been tested?
tests
canary -> `2.8.7-canary.8`
## Checklist

### DIFF
--- a/providers/ethereum-provider/src/EthereumProvider.ts
+++ b/providers/ethereum-provider/src/EthereumProvider.ts
@@ -178,6 +178,12 @@ export function buildNamespaces(params: NamespacesParams): {
     optional: optionalChains.length ? optional : undefined,
   };
 }
+
+// helper type to force setting at least one value in an array
+type ArrayOneOrMore<T> = {
+  0: T;
+} & Array<T>;
+
 /**
  * @param {number[]} chains - The Chains your app intents to use and the peer MUST support. If the peer does not support these chains, the connection will be rejected.
  * @param {number[]} optionalChains - The Chains your app MAY attempt to use and the peer MAY support. If the peer does not support these chains, the connection will still be established.
@@ -185,12 +191,12 @@ export function buildNamespaces(params: NamespacesParams): {
  */
 export type ChainsProps =
   | {
-      chains: number[];
+      chains: ArrayOneOrMore<number>;
       optionalChains?: number[];
     }
   | {
       chains?: number[];
-      optionalChains: number[];
+      optionalChains: ArrayOneOrMore<number>;
     };
 
 export type EthereumProviderOptions = {
@@ -460,9 +466,11 @@ export class EthereumProvider implements IEthereumProvider {
   }
 
   protected getRpcConfig(opts: EthereumProviderOptions): EthereumRpcConfig {
-    const requiredChains = opts.chains || [];
-    const optionalChains = opts.optionalChains || [];
+    const requiredChains = opts?.chains ?? [];
+    const optionalChains = opts?.optionalChains ?? [];
     const allChains = requiredChains.concat(optionalChains);
+    if (!allChains.length)
+      throw new Error("No chains specified in either `chains` or `optionalChains`");
     const requiredMethods = requiredChains.length ? opts?.methods || REQUIRED_METHODS : [];
     const requiredEvents = requiredChains.length ? opts?.events || REQUIRED_EVENTS : [];
     const optionalMethods = opts?.optionalMethods || [];
--- a/providers/ethereum-provider/src/EthereumProvider.ts
+++ b/providers/ethereum-provider/src/EthereumProvider.ts
@@ -10,7 +10,13 @@ import {
 import { Metadata, Namespace, UniversalProvider } from "@walletconnect/universal-provider";
 import type { WalletConnectModalConfig, WalletConnectModal } from "@walletconnect/modal";
 import { SessionTypes, SignClientTypes } from "@walletconnect/types";
-import { STORAGE_KEY, REQUIRED_METHODS, REQUIRED_EVENTS, RPC_URL } from "./constants";
+import {
+  STORAGE_KEY,
+  REQUIRED_METHODS,
+  REQUIRED_EVENTS,
+  RPC_URL,
+  OPTIONAL_METHODS,
+} from "./constants";
 
 export type QrModalOptions = Pick<
   WalletConnectModalConfig,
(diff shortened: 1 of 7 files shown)


providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -176,6 +177,12 @@ export class UniversalProvider implements IUniversalProvider {
       await approval()
         .then((session) => {
           this.session = session;
+          // assign namespaces from session if not already defined
```
bkrem, line 180: Been comparing code here but still a bit unclear on this step:
* I get that we need to call `setNamespaces` again after calling it initially in `connect` to account for any differences in the session vs the original proposal
* But how come we overwrite `this.namespaces` right afterwards and then re-persist manually?

providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -389,25 +398,27 @@ export class UniversalProvider implements IUniversalProvider {
...
-    this.optionalNamespaces = optionalNamespaces;
     this.sessionProperties = sessionProperties;
     this.persist("namespaces", namespaces);
     this.persist("optionalNamespaces", optionalNamespaces);
   }
 
   private validateChain(chain?: string): [string, string] {
     const [namespace, chainId] = chain?.split(":") || ["", ""];
-
+    if (!this.namespaces || !Object.keys(this.namespaces).length) return [namespace, chainId];
```
bkrem, line 414: Could you add context here please for the early return 🙏

providers/universal-provider/src/utils/misc.ts
```diff
@@ -101,3 +101,29 @@ export function normalizeNamespaces(namespaces: NamespaceConfig): NamespaceConfi
 export function parseCaip10Account(caip10Account: string): string {
   return caip10Account.includes(":") ? caip10Account.split(":")[2] : caip10Account;
 }
+
+/**
+ * Populates the chains array for each namespace with the chains extracted from the accounts if are otherwise missing
```
bkrem, line 106:
```suggestion
 * Populates the chains array for each namespace with the chains extracted from the accounts if they are otherwise missing
```

providers/universal-provider/src/utils/misc.ts
```diff
@@ -101,3 +101,29 @@ export function normalizeNamespaces(namespaces: NamespaceConfig): NamespaceConfi
...
+ */
+export function populateNamespacesChains(
+  namespaces: SessionTypes.Namespaces,
+): Record<string, SessionTypes.Namespace> {
+  const parsedNamespaces: Record<string, SessionTypes.Namespace> = {};
+  for (const [key, values] of Object.entries(namespaces)) {
+    const methods = values.methods || [];
+    const events = values.events || [];
+    const accounts = values.accounts || [];
+    const chains = isCaipNamespace(key)
```
bkrem, line 117: The shorthand via `isCaipNamespace` here is a bit confusing at first, since the intent is actually to see if the key is a full namespace + chainId definition right (e.g. `eip155:5`)?

```suggestion
    // If the key includes a CAIP separator `:` we know it's a namespace + chainId (e.g. `eip155:1`)
    const chains = isCaipNamespace(key)
```

providers/ethereum-provider/src/EthereumProvider.ts
```diff
@@ -121,18 +127,13 @@ export function buildNamespaces(params: NamespacesParams): {
...
   const required: Namespace = {
-    chains: requiredChains,
-    methods: requiredMethods,
-    events: requiredEvents,
-    rpcMap: requiredRpcMap,
+    chains,
+    methods: methods || REQUIRED_METHODS,
+    events: events || REQUIRED_EVENTS,
+    rpcMap: {
+      ...(chains.length && { [getEthereumChainId(chains)]: rpcMap[getEthereumChainId(chains)] }),
```
bkrem: Seems like this may attempt to spread `0` if `chains.length === 0`. Maybe safer with a ternary and always returning an object?

```suggestion
      ...(chains.length ? { [getEthereumChainId(chains)]: rpcMap[getEthereumChainId(chains)] } : {}),
```

providers/ethereum-provider/src/EthereumProvider.ts
```diff
@@ -160,31 +161,40 @@ export function buildNamespaces(params: NamespacesParams): {
   const optional: Namespace = {
     chains: [
       ...new Set(
-        shouldIncludeRequiredChains ? requiredChains.concat(optionalChains || []) : optionalChains,
+        shouldIncludeRequiredChains ? required.chains.concat(optionalChains || []) : optionalChains,
+      ),
+    ],
+    methods: [
+      ...new Set(
+        required.methods.concat(optionalMethods?.length ? optionalMethods : OPTIONAL_METHODS),
```
bkrem, line 172: Not sure if intentional so flagging: seems like we're now always adding `OPTIONAL_METHODS` to methods if `optionalMethods` is not defined. 

Makes sense to me but is a deviation from the previous `[]`

bkrem: Question for me is: shouldn't we then do the same with `OPTIONAL_EVENTS`?

bkrem (approved):
💪 💯

providers/universal-provider/src/UniversalProvider.ts
ganchoradkov: removed the redundant `this.setNamespaces({ namespaces: session.namespaces as NamespaceConfig });`

providers/universal-provider/src/UniversalProvider.ts
ganchoradkov: so we don't process the logic after such as `validating namespaces` & trying to return `defaultChain`

providers/ethereum-provider/src/EthereumProvider.ts
ganchoradkov: good suggestion

wildlifechorus:
Please merge this we are dying over here: https://github.com/orgs/WalletConnect/discussions/2957

ccxdev:
@0xAsimetriq @ignaciosantise @JonathanConn 

Review it please, because we facing problem https://github.com/orgs/WalletConnect/discussions/2957

bkrem (approved):

PR #2841, opened 2023-06-28
fix: history records expiry

ganchoradkov:
## Description
Implemented a default expiry of history records of 30 days. 
Additionally, history records are removed on request-response after they're no longer needed.

## Type of change

## How has this been tested?
tests
dogfooding

## Checklist

### DIFF
--- a/packages/core/src/controllers/history.ts
+++ b/packages/core/src/controllers/history.ts
@@ -220,7 +220,7 @@ export class JsonRpcHistory extends IJsonRpcHistory {
   private cleanup() {
     try {
       this.records.forEach((record: JsonRpcRecord) => {
-        if (!record.expiry || record.expiry < Date.now()) {
+        if (!record.expiry || record.expiry <= Date.now()) {
           this.logger.info(`Deleting expired history log: ${record.id}`);
           this.delete(record.topic, record.id);
         }
--- a/packages/core/src/controllers/history.ts
+++ b/packages/core/src/controllers/history.ts
@@ -1,8 +1,10 @@
 import { formatJsonRpcRequest, isJsonRpcError } from "@walletconnect/jsonrpc-utils";
 import { generateChildLogger, getLoggerContext, Logger } from "@walletconnect/logger";
 import { IJsonRpcHistory, JsonRpcRecord, RequestEvent, ICore } from "@walletconnect/types";
-import { getInternalError } from "@walletconnect/utils";
+import { calcExpiry, getInternalError } from "@walletconnect/utils";
 import { EventEmitter } from "events";
+import { THIRTY_DAYS } from "@walletconnect/time";
+import { HEARTBEAT_EVENTS } from "@walletconnect/heartbeat";
 import {
   CORE_STORAGE_PREFIX,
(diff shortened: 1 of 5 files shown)


packages/core/src/controllers/history.ts
```diff
@@ -208,6 +211,23 @@ export class JsonRpcHistory extends IJsonRpcHistory {
...
+
+    this.core.heartbeat.on(HEARTBEAT_EVENTS.pulse, () => {
+      this.cleanup();
+    });
+  }
+
+  private cleanup() {
+    try {
+      this.records.forEach((record: JsonRpcRecord) => {
+        if (!record.expiry || record.expiry < Date.now()) {
```
xzilja:
```suggestion
        if (!record.expiry || record.expiry <= Date.now()) {
```

xzilja (approved):

JonathanConn (approved):

JonathanConn (approved):

packages/core/src/controllers/history.ts
```diff
@@ -80,6 +82,7 @@ export class JsonRpcHistory extends IJsonRpcHistory {
       topic,
       request: { method: request.method, params: request.params || null },
       chainId,
+      expiry: Date.now() + calcExpiry(THIRTY_DAYS),
```
bkrem: Don't need the explicit `Date.now()` call, handled by `calcExpiry` internally if not passed as second param: https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/utils/src/misc.ts#L313

```suggestion
      expiry: calcExpiry(THIRTY_DAYS),
```

bkrem (approved):

packages/core/src/controllers/history.ts
ganchoradkov: I tried it and it didn't calculate the correct expiry. It was calculating 30 days in milliseconds rather than `Date.now() + calcExpiry(THIRTY_DAYS)` may be bug in the util?

packages/core/src/controllers/history.ts
```diff
@@ -208,6 +211,23 @@ export class JsonRpcHistory extends IJsonRpcHistory {
...
+
+    this.core.heartbeat.on(HEARTBEAT_EVENTS.pulse, () => {
+      this.cleanup();
+    });
+  }
+
+  private cleanup() {
+    try {
+      this.records.forEach((record: JsonRpcRecord) => {
+        if (!record.expiry || record.expiry <= Date.now()) {
```
bkrem: ~Wait doesn't this need to be `>=` (if expiry is greater than current timestamp)?~ Disregard 🤦 but point regarding comparison of seconds to ms still seems to stand:

We normally store `expiry` in seconds but `Date.now()` returns milliseconds

bkrem (changes requested):

packages/core/src/controllers/history.ts
ganchoradkov: the tests confirm the expected behaviour

bkrem (approved):

PR #3111, opened 2023-07-12
feat: user action feedback

ganchoradkov:
## Description

- implemented `isOnline` util for browser & react-native platforms to check network state before attempting to publish
- implemented `subscribeToNetworkChange` util that calls a callback when the network changes - online/offline
- reworked set of requests to bubble exceptions to the top so users can be notified when publishing failed

## Type of change

## How has this been tested?
tests
dogfooding 
canaries  - monorepo`2.9.2-canary-eb8558a`, web3wallet `1.8.8-canary-eb8558a`

## Checklist


packages/react-native-compat/package.json
```diff
@@ -21,6 +21,7 @@
     "prettier": "prettier --check './**/*.{js,ts,jsx,tsx}'"
   },
   "dependencies": {
+    "@react-native-community/netinfo": "^9.4.1",
```
ignaciosantise: we need to add this as a peerDependency and tell the user to install it separately to make react native link it correctly

Similar issue with linking modules: https://stackoverflow.com/a/59777212/6562073

ganchoradkov: Yeah, additionally we have to update the docs 👍

packages/sign-client/test/sdk/integration/push.spec.ts
```diff
@@ -10,7 +10,7 @@ import {
 import { describe, it, expect, afterEach, beforeAll } from "vitest";
 import { PairingTypes, SessionTypes } from "@walletconnect/types";
 
-describe("Push", () => {
+describe.skip("Push", () => {
```
bkrem: Debug override?

```suggestion
describe("Push", () => {
```

packages/sign-client/package.json
```diff
@@ -27,7 +27,7 @@
     "test:run": "vitest run --dir test/sdk",
     "test:concurrency": "vitest run --dir test/concurrency",
     "test:xregion": "vitest run --dir test/xregion -- --dangerouslyIgnoreUnhandledErrors --segfault-retry=3",
-    "test": "npm run test:pre; npm run test:run",
+    "test": "npm run test:pre; npm run test:run -- --allowOnly",
```
bkrem: Debug override or want to keep this?

providers/universal-provider/test/index.spec.ts
```diff
@@ -76,6 +76,13 @@ describe("UniversalProvider", function () {
     await provider.client.core.relayer.transportClose();
     await walletClient.client?.core.relayer.transportClose();
   });
+  beforeEach(() => {
```
bkrem: Debug code or want to keep it?

packages/utils/src/network.ts
```diff
@@ -0,0 +1,70 @@
...
+      break;
+    case ENV_MAP.node:
+      // wip: need to implement
+      break;
+    default:
+      break;
+  }
+}
+
+export function subscribeToBrowserNetworkChange(callBackHendler: (connected: boolean) => void) {
```
bkrem:
```suggestion
export function subscribeToBrowserNetworkChange(callbackHandler: (connected: boolean) => void) {
```

packages/utils/src/network.ts
```diff
@@ -0,0 +1,70 @@
...
+  // fallback to true if global.NetInfo is undefined, meaning an older version of react-native-compat is used
+  return true;
+}
+
+export function getNodeOnlineStatus() {
+  // wip: may be send quick request to check?
+  return true;
+}
+
+export function subscribeToNetworkChange(callBackHendler: (connected: boolean) => void) {
```
bkrem:
```suggestion
export function subscribeToNetworkChange(callbackHandler: (connected: boolean) => void) {
```

packages/utils/src/network.ts
```diff
@@ -0,0 +1,70 @@
...
+  if (isReactNative() && typeof global !== "undefined" && (global as any)?.NetInfo) {
+    const state = await (global as any)?.NetInfo.fetch();
+    return state?.isConnected;
+  }
+  // fallback to true if global.NetInfo is undefined, meaning an older version of react-native-compat is used
+  return true;
+}
+
+export function getNodeOnlineStatus() {
+  // wip: may be send quick request to check?
```
bkrem: Could we add a proper todo instead?

```suggestion

```

packages/core/src/controllers/relayer.ts
```diff
@@ -392,8 +438,9 @@ export class Relayer extends IRelayer {
   }
 
   private async toEstablishConnection() {
-    if (this.connected) return;
-    if (this.connecting) {
+    await this.confirmOnlineStateOrThrow();
+    if (this.connected && (await isOnline())) return;
```
bkrem: Seems like this is duplicating the `isOnline` check done inside `confirmOnlineStateOrThrow` one line above?

packages/utils/src/network.ts
```diff
@@ -0,0 +1,70 @@
...
+  return isBrowser() && navigator?.onLine;
+}
+
+export async function getReactNativeOnlineStatus(): Promise<boolean> {
+  // global.NetInfo is set in react-native-compat
+  if (isReactNative() && typeof global !== "undefined" && (global as any)?.NetInfo) {
+    const state = await (global as any)?.NetInfo.fetch();
+    return state?.isConnected;
+  }
+  // fallback to true if global.NetInfo is undefined, meaning an older version of react-native-compat is used
```
bkrem, line 32: Is this still needed/wanted given the polyfill you added in `react-native-compat`? I.e. should this always be able to resolve `NetInfo`?

packages/utils/src/network.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export function subscribeToBrowserNetworkChange(callBackHendler: (connected: boolean) => void) {
+  if (isBrowser()) {
+    window.addEventListener("online", () => callBackHendler(true));
+    window.addEventListener("offline", () => callBackHendler(false));
+  }
+}
+
+// global.NetInfo is set in react-native-compat
+export function subscribeToReactNativeNetworkChange(callBackHendler: (connected: boolean) => void) {
```
bkrem:
```suggestion
export function subscribeToReactNativeNetworkChange(callbackHandler: (connected: boolean) => void) {
```

bkrem (commented):
Review batch 1/2

Getting these out there already, will finish review once I'm back online 👍

packages/utils/src/network.ts
```diff
@@ -0,0 +1,70 @@
...
+      // wip: need to implement
+      break;
+    default:
+      break;
+  }
+}
+
+export function subscribeToBrowserNetworkChange(callBackHendler: (connected: boolean) => void) {
+  if (isBrowser()) {
+    window.addEventListener("online", () => callBackHendler(true));
```
bkrem: Is there a risk of us creating an event listener leak here if they are never removed?

packages/core/src/controllers/publisher.ts
```diff
@@ -46,19 +45,27 @@ export class Publisher extends IPublisher {
       const id = opts?.id || (getBigIntRpcId().toString() as any);
       const params = { topic, message, opts: { ttl, relay, prompt, tag, id } };
       // delay adding to queue to avoid cases where heartbeat might pulse right after publish resulting in duplicate publish
-      const queueTimeout = setTimeout(() => this.queue.set(id, params), this.queueTimeout);
+      const queueTimeout = setTimeout(() => this.queue.set(id, params), this.publishTimeout);
```
bkrem, line 48: What's the reason for the change/aligning with the publish timeout?

packages/core/src/controllers/publisher.ts
```diff
@@ -46,19 +45,27 @@ export class Publisher extends IPublisher {
...
         );
         await publish;
-        clearTimeout(queueTimeout);
+        this.removeRequestFromQueue(id);
         this.relayer.events.emit(RELAYER_EVENTS.publish, params);
       } catch (err) {
         this.logger.debug(`Publishing Payload stalled`);
         this.needsTransportRestart = true;
+        if (opts?.internal?.throwOnFailedPublish) {
+          // remove the request from the queue so it's not retried automatically
```
bkrem, line 62: Can you explain why we want to remove the request from queue if it's seemingly due to a connection issue? Wouldn't we want to retry when the connection comes back online?

packages/core/src/controllers/relayer.ts
```diff
@@ -88,6 +95,7 @@ export class Relayer extends IRelayer {
 
   public async init() {
     this.logger.trace(`Initialized`);
+    this.registerEventListeners();
```
bkrem, line 98: Does moving this up in the sequence have a specific effect?

packages/core/src/controllers/relayer.ts
```diff
@@ -190,86 +197,90 @@ export class Relayer extends IRelayer {
 
   public async transportClose() {
     this.transportExplicitlyClosed = true;
-    if (this.connected) {
+    /**
+     * if there was a network disruption like restart of network driver, the socket is most likely stalled and we can't rely on it
+     * in such case provider.disconnect() is not reliable,it might resolve after a long time or not emit disconnect event at all
```
```suggestion
     * as in this case provider.disconnect() is not reliable, since it might resolve after a long time or not emit disconnect event at all.
```

packages/core/src/controllers/relayer.ts
```diff
@@ -190,86 +197,90 @@ export class Relayer extends IRelayer {
...
+     */
+    if (this.hasExperiencedNetworkDisruption && this.connected) {
+      await createExpiringPromise(this.provider.disconnect(), 1000, "provider.disconnect()").catch(
```
bkrem, line 205: Checking if I'm seeing this right: we attempt to call `provider.disconnect()` and call `onProviderDisconnect` regardless, i.e. it either resolves or call it after 1 second.

bkrem (approved):

packages/utils/src/network.ts
ganchoradkov: This is to safeguard against cases where projects use an outdated version of `react-native-compat`. We want to avoid  a case where isOnline returns incorrectly `false` permanently

packages/utils/src/network.ts
ganchoradkov: This util is called once per relayer(core) initialization so shouldn't be the case but I can add an unsubscribe just in case

packages/core/src/controllers/publisher.ts
ganchoradkov: Adding items to the queue is directly related to the publish timeout so using single value makes sense as they have to be aligned always

packages/core/src/controllers/publisher.ts
ganchoradkov: There are two cases where payloads can fail to publish
1. no internet connection
2. no relay ACK
We need to mark some payloads such as user-initiated ones and they must be retried by the user manually.

packages/core/src/controllers/relayer.ts
ganchoradkov: Just to be sure all events are subscribed by the time connection is attempted

packages/core/src/controllers/relayer.ts
ganchoradkov: we call `onProviderDisconnect` manually only if `provider.disconnect()` fails to resolve within the timeout. As pointed out by the comment above, when the socket is stalled, disconnect is unreliable

PR #3137, opened 2023-07-13
fix(relayer): prevents forwarding empty/invalid message payloads to clients

bkrem:
## Description

- Pairing: ignores non-pairing topics earlier on incoming provider message to avoid redundant controller work
- Relayer: Ensures we don't emit invalid/empty/garbage messages to clients
- Context: https://walletconnect.slack.com/archives/C03RWGSBCUU/p1689247549066609

## Type of change

## How has this been tested?

- Unit tests
- Canary (`2.9.0-e4636394`)

## Checklist

### DIFF
--- a/packages/core/src/controllers/pairing.ts
+++ b/packages/core/src/controllers/pairing.ts
@@ -240,10 +240,11 @@ export class Pairing implements IPairing {
     this.core.relayer.on(RELAYER_EVENTS.message, async (event: RelayerTypes.MessageEvent) => {
       const { topic, message } = event;
 
+      // Do not handle if the topic is not related to known pairing topics.
+      if (!this.pairings.keys.includes(topic)) return;
+
       // messages of certain types should be ignored as they are handled by their respective SDKs
-      if (this.ignoredPayloadTypes.includes(this.core.crypto.getPayloadType(message))) {
-        return;
-      }
+      if (this.ignoredPayloadTypes.includes(this.core.crypto.getPayloadType(message))) return;
 
       const payload = await this.core.crypto.decode(topic, message);
       if (isJsonRpcRequest(payload)) {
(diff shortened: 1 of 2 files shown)


packages/core/src/controllers/relayer.ts
```diff
@@ -292,10 +292,28 @@ export class Relayer extends IRelayer {
...
+    }
+
+    // Ignore if `topic` is not subscribed to.
+    if (!(await this.subscriber.isSubscribed(topic))) {
+      this.logger.debug(`Ignoring message for non-subscribed topic ${topic}`);
+      return true;
+    }
+
+    // Ignore if `message` is a duplicate.
     const exists = this.messages.has(topic, message);
```
ganchoradkov, line 313: nice! This should actually solve `no matching key: history` on duplicates that were already processed 💯 💯

ganchoradkov (approved):

packages/core/src/controllers/pairing.ts
```diff
@@ -240,10 +240,11 @@ export class Pairing implements IPairing {
     this.core.relayer.on(RELAYER_EVENTS.message, async (event: RelayerTypes.MessageEvent) => {
       const { topic, message } = event;
 
+      // Do not handle if the topic is not related to known pairing topics.
+      if (!this.pairings.keys.includes(topic)) return;
```
devceline, line 244: What about operations that don't generate pairings ? Or under the hood everything generates a pairing? Thinking push_message and chat_invite etc

devceline (approved):
Approved, just left a question so I understand core better

packages/core/src/controllers/pairing.ts
bkrem: The pairing controller is only able to handle messages on topics that pertain to pairings, since the only RPC methods handled here are `wc_pairingPing` and `wc_pairingDelete`.

This isn't directly related to the message fix, but I moved this logic up in the controller's incoming message handling since we were still doing unnecessary/unwanted work like `core.history.set` here before the pairing controller would exit due to this condition

PR #3220, opened 2023-07-19
feat: session requests queue

ganchoradkov:
## Description
Implemented `session_request` queue where pending requests are emitted - one after another instead of immediately when received by the SDK

## Type of change

## How has this been tested?
tests
canary - `2.10.0-canary-af1c2bee`
## Checklist


packages/sign-client/src/controllers/engine.ts
```diff
@@ -90,6 +99,11 @@ export class Engine extends IEngine {
       this.registerExpirerEvents();
       this.client.core.pairing.register({ methods: Object.keys(ENGINE_RPC_OPTS) });
       this.initialized = true;
+
+      setTimeout(() => {
```
bkrem, line 103: Context on why we need a timeout here?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -878,6 +891,43 @@ export class Engine extends IEngine {
     }
   };
 
+  private addRequestToQueue = (request: PendingRequestTypes.Struct) => {
+    this.requestQueue.requests.push(request);
+  };
+
+  private afterResponse = (params: EngineTypes.RespondParams) => {
+    this.deletePendingSessionRequest(params.response.id, { message: "fulfilled", code: 0 });
+    // intentionally delay the emitting of the next pending request a bit
```
bkrem, line 904: Why?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -878,6 +891,43 @@ export class Engine extends IEngine {
     }
   };
 
+  private addRequestToQueue = (request: PendingRequestTypes.Struct) => {
+    this.requestQueue.requests.push(request);
+  };
+
+  private afterResponse = (params: EngineTypes.RespondParams) => {
```
bkrem: Would try to keep naming conventions in line with `<verb>...`

```suggestion
  private cleanupAfterResponse = (params: EngineTypes.RespondParams) => {
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -878,6 +891,43 @@ export class Engine extends IEngine {
...
+    this.deletePendingSessionRequest(params.response.id, { message: "fulfilled", code: 0 });
+    // intentionally delay the emitting of the next pending request a bit
+    setTimeout(() => {
+      this.requestQueue.state = REQUEST_QUEUE_STATES.idle;
+      this.processRequestQueue();
+    }, toMiliseconds(this.requestQueueDelay));
+  };
+
+  private processRequestQueue = async () => {
+    if (this.requestQueue.state === REQUEST_QUEUE_STATES.active) {
```
bkrem, line 912: Q: Trying to fully grok what we're doing here: this is it effectively creating a lock state where queue state remains active until the oldest pending request has been responded to?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -878,6 +891,43 @@ export class Engine extends IEngine {
...
+      this.client.logger.info("session request queue is already active.");
+      return;
+    }
+
+    const request = this.requestQueue.requests[0];
```
bkrem, line 917:
```suggestion
    // Select the first/oldest request in the array to ensure last-in-first-out (LIFO)
    const request = this.requestQueue.requests[0];
```

bkrem (approved):

packages/sign-client/src/controllers/engine.ts
ganchoradkov: Timeout is needed to allow wallets to subscribe to `session_request` after `SignClient.init` resolves. The 1s delay is an extra precaution so the first emitted event is not missed due to app not being in a fully ready state to handle events. We can adjust though

packages/sign-client/src/controllers/engine.ts
ganchoradkov: same reasoning as above, to allow the app to process the previous request's views/modals/state

packages/sign-client/src/controllers/engine.ts
ganchoradkov: Yes, or expires, this way we ensure all requests are responded to

Issue #2970, opened 2023-07-03
{context: 'core/relayer'}: socket stalled  at eval


mqklin:
**Describe the bug**
`open` function call from `useWeb3Modal()` just caused `{context: 'core/relayer'}: socket stalled` error.

**SDK Version (if relevant)**
"@web3modal/ethereum": "^2.4.7",
"@web3modal/react": "^2.4.7",

**To Reproduce**

**Smartphone (please complete the following information):**
- Device: iPhone 12 mini
- OS: iOS 16.5.1
- App Version: 7.0.1 (MetaMask)
![image](https://github.com/WalletConnect/walletconnect-monorepo/assets/9456433/33549157-c6fd-4e21-b8a8-3c87a098ce50)

ganchoradkov:
Hey @mqklin this error means the client can't establish a connection with the relay. 
Are you using some VPN or proxy that might be causing this?

mqklin:
Hey @ganchoradkov 
no I don't. 
I just tested WC2 the whole day today, and got this error 2 times.

ric344:
{"context": "core"} {"context": "core/relayer"} [Error: WebSocket connection failed for URL: wss://relay.walletconnect.com?auth=eyJhbGciOiJFZERTQSIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJkaWQ6a2V5Ono2TWtpajZ5WE1IYzdaSnhWdjd1QW5ENzdWb3Myc1ZDaUc2M05EM1d3Y29iZXptWiIsInN1YiI6ImI2ZmE2OWRiMzM1OGY5OGE2Y2VlZjdiYzlkNGU2NGI2OWI1MjBhZDE0NTZjMzQwOGNjMWJmM2NjNjNjMGNmOGIiLCJhdWQiOiJ3c3M6Ly9yZWxheS53YWxsZXRjb25uZWN0LmNvbSIsImlhdCI6MTY4ODUwNjc4OSwiZXhwIjoxNjg4NTkzMTg5fQ.BvTRkiCVmeyi-HBs5ufdAA4NS4bF0B-ahg4jQaF53x8ermz69IjiHVNuAgkFJYSU2UL_hXeGX4e8NV4nb-8cBQ&projectId=761c84c077c0d977b8f4e20fbd4eae64&ua=wc-2%2Fjs-y%2Funknown-%2Freact-native&useOnCloseEvent=true]

roman-utila:
[...]

Issue #3065, opened 2023-07-10
Our QA team and sentry summarize the issues that occur while using walletconnect


siosio34:
**Describe the bug**
Our QA team and sentry summarize the issues that occur while using walletconnect.

1. switchChain does not work properly in mobile browser using walletconnect. Similar issues include:
https://github.com/MetaMask/metamask-mobile/issues/6655
https://github.com/wagmi-dev/references/issues/393

2. To use QR modal, please install @walletconnect/modal package 
-The QR module appears properly, but the error seems to occur intermittently.
- As a result of tracking through sentry, it mostly occurred in IOS 16 or higher.

3. Attempt to connect to relay via `transportOpen` has stalled. Retrying...
- The error below occurs intermittently and gets caught by the sentry, and I don't know when or why it happens.

4. WebSocket connection failed for URL
- The error below occurs intermittently and gets caught by the sentry, and I don't know when or why it happens.

**SDK Version (if relevant)**
 - Client: JS, metamask
 - Version
walletconnect - 2.8.4
wagmi - 1.3.2
viem - 1.0.0

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

zhy023:
The same problem happened to me...

imaksp:
Is this metamask app issue or walletconnect?

siosio34:
@imaksp 
I'm not sure exactly what the problem is.

However, it does not happen in the Metamask Chrome extension extension, and it seems that the bugs occur when connecting Metamask Mobile through Wallet Connect.

N3ST3DLabs:
having the same problem with Metamask on chrome and brave ....... Viem has been a nightmare upgrade .........Even bugs in   usereadContract Allowance function returns ON instead of a UINT256 .  :(  i guess moving back to old version with ether is much better

frankye23:
> 3. Attempt to connect to relay via transportOpen has stalled. Retrying...

I also run into this error. In my case, in order to reproduce, i have to send multiple request to my own api. And the error randomly show in developer console. But i wonder why axios request will be stalled by wc socket stalled.

kdenhartog:
I believe this can be closed now that #3230 has been merged

glitch-txs:
1. These are MetaMask related issues. They are being worked on atm.
2. So far I've seen this issue and was related to code implementation. The error message seems off though.

[...]

PR #3230, opened 2023-07-20
Up the timeout to 10 seconds rather than 5 seconds

kdenhartog:


kdenhartog:
closes #2970 #2392 #3065

The timeout got dropped from 10 seconds to 5 seconds in this commit https://github.com/WalletConnect/walletconnect-monorepo/commit/9c3fb2e6fc28b563ed87c7a93937d021e9dbb49c#diff-2666bbe86db232da1fe48efc66faddc72893a26741e59bcc7510ea3818bbd581R193 which made it so that it will non-deterministically fail if the page is executing other scripts at the time of trying to setup the web socket. Switching this back to 10 seconds is a bit of a hack to avoid the problem and realistically the better solution is to find a way to determine if the connection has stalled without a timeout which was added in https://github.com/WalletConnect/walletconnect-monorepo/pull/1758.

kdenhartog:
cc @ganchoradkov since you're the one who's interacted with this code most recently

ganchoradkov (approved):
Thank you for the PR 💯  Awesome timing as it was on our radar to increase the timeout 🥇

PR #3266, opened 2023-07-25
refactor: wrap try/catch on payload processing

ganchoradkov:
## Description
Wrapped Pairing API & Sign Client payload processing in try/catch

## Type of change

## How has this been tested?
dogfooding 

## Checklist

### DIFF
--- a/packages/core/src/controllers/pairing.ts
+++ b/packages/core/src/controllers/pairing.ts
@@ -246,14 +246,18 @@ export class Pairing implements IPairing {
       // messages of certain types should be ignored as they are handled by their respective SDKs
       if (this.ignoredPayloadTypes.includes(this.core.crypto.getPayloadType(message))) return;
 
-      const payload = await this.core.crypto.decode(topic, message);
-      if (isJsonRpcRequest(payload)) {
-        this.core.history.set(topic, payload);
-        this.onRelayEventRequest({ topic, payload });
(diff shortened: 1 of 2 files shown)


packages/core/src/controllers/pairing.ts
```diff
@@ -246,14 +246,18 @@ export class Pairing implements IPairing {
...
+        if (isJsonRpcRequest(payload)) {
+          this.core.history.set(topic, payload);
+          this.onRelayEventRequest({ topic, payload });
+        } else if (isJsonRpcResponse(payload)) {
+          await this.core.history.resolve(payload);
+          await this.onRelayEventResponse({ topic, payload });
+          this.core.history.delete(topic, payload.id);
+        }
+      } catch (error) {
+        this.logger.error(error);
```
bkrem, line 261: What are we hoping to achieve with the extra try/catch? Could be cleaner to try catch the `decode` specifically and the request/response branching separately?

Like this it’s one big catch so it may be unclear what the cause was as-is if the error is unclear.

bkrem (approved):

packages/core/src/controllers/pairing.ts
ganchoradkov: the `decode` has its own try-catch internally so it doesn't really throw anything. Might make sense to move it out of this try-catch so it's not confusing when reading the code

PR #3410, opened 2023-08-09
feat: verify fallback

ganchoradkov:
## Description
Implemented fallback to `.org` on Verify Iframe & Attestation resolve.

Additionally, updated iframe `onload` behavior to be triggered by a message ping by the enclave JS instead of relaying on iframe `onload/onerror` events.

## Type of change

## How has this been tested?
dogfooding on Chome/Mozilla/Safari

## Checklist

### DIFF
--- a/packages/core/src/controllers/verify.ts
+++ b/packages/core/src/controllers/verify.ts
@@ -71,13 +71,13 @@ export class Verify extends IVerify {
     const mainUrl = params?.verifyUrl || VERIFY_SERVER;
     let result = "";
     try {
-      result = await this.fetch(params.attestationId, mainUrl);
+      result = await this.fetchAttestation(params.attestationId, mainUrl);
     } catch (error) {
       this.logger.warn(
         `failed to resolve attestation: ${params.attestationId} from url: ${mainUrl}`,
       );
       this.logger.warn(error);
-      result = await this.fetch(params.attestationId, VERIFY_FALLBACK_SERVER);
+      result = await this.fetchAttestation(params.attestationId, VERIFY_FALLBACK_SERVER);
     }
     return result;
   };
@@ -86,7 +86,7 @@ export class Verify extends IVerify {
     return getLoggerContext(this.logger);
   }
 
-  private fetch = async (attestationId: string, url: string) => {
+  private fetchAttestation = async (attestationId: string, url: string) => {
     this.logger.info(`resolving attestation: ${attestationId} from url: ${url}`);
     // set artificial timeout to prevent hanging
     const timeout = this.startAbortTimer(ONE_SECOND * 2);
--- a/packages/core/src/controllers/verify.ts
+++ b/packages/core/src/controllers/verify.ts
@@ -33,7 +33,21 @@ export class Verify extends IVerify {
       this.removeIframe();
     }
     this.verifyUrl = verifyUrl;
-    await this.createIframe();
+
+    await this.createIframe().catch((error) => {
+      this.logger.error(`Verify iframe failed to load: ${this.verifyUrl}`);
+      this.logger.error(error);
+    });
+
+    if (this.initialized) return;
+
+    this.removeIframe();
+    this.verifyUrl = VERIFY_FALLBACK_SERVER;
+
+    await this.createIframe().catch((error) => {
+      this.logger.error(`Verify iframe failed to load: ${this.verifyUrl}`);
+      this.logger.error(error);
+    });
   };
 
   public register: IVerify["register"] = async (params) => {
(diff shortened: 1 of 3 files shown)


packages/core/src/controllers/verify.ts
```diff
@@ -47,21 +61,35 @@ export class Verify extends IVerify {
 
   public resolve: IVerify["resolve"] = async (params) => {
     if (this.isDevEnv) return "";
+    const mainUrl = params?.verifyUrl || VERIFY_SERVER;
```
ignaciosantise, line 75: if the user defines a `verifyUrl` by param, should we use VERIFY_FALLBACK_SERVER in case of failure too? Or in that case just fail without a fallback

ganchoradkov: atm is an immediate fallback to `.org` consistently to `relayUrl` fallback

bkrem:
> Additionally, updated iframe onload behavior to be triggered by a message ping by the enclave JS instead of relaying on iframe onload/onerror events.

What was the context for this change? Was this necessary/fixing an issue or an improvement?

packages/core/src/controllers/verify.ts
```diff
@@ -33,7 +38,23 @@ export class Verify extends IVerify {
       this.removeIframe();
     }
     this.verifyUrl = verifyUrl;
-    await this.createIframe();
+
+    await this.createIframe().catch((error) => {
```
bkrem: Generally not a huge fan of mixing `async/await` and `then/catch` in functions that are already procedural (i.e. not a functional pipe), but if you have a strong preference here OK by me 👍

bkrem: (The alternative being a `try/catch` statement)

packages/core/src/controllers/verify.ts
```diff
@@ -33,7 +38,23 @@ export class Verify extends IVerify {
...
+      this.logger.error(`Verify iframe failed to load: ${this.verifyUrl}`);
+      this.logger.error(error);
+    });
+
+    if (this.initialized) return;
```
bkrem, line 49: So `this.initialized` is a proxy for the initial `createIframe` having succeeded on `.com`? Wouldn't it be easier to do the handling below in the `catch` of the first attempt and avoid this check on `this.initialized`? 

E.g. `.warn` on initial failure and then log `.error` like [here](https://github.com/WalletConnect/walletconnect-monorepo/pull/3410/files#diff-adfd0925fb8e8e5ac3dfdf97b5e40ddf5957b1bc86ef2e065bc88f60d887ea82R53) once the fallback fails too.

At least that's how we're doing it for the WS connection atm.

packages/core/src/controllers/verify.ts
```diff
@@ -47,21 +68,35 @@ export class Verify extends IVerify {
...
+    }
+    return result;
+  };
 
-    this.logger.info(`resolving attestation: ${params.attestationId}`);
+  get context(): string {
+    return getLoggerContext(this.logger);
+  }
+
+  private fetch = async (attestationId: string, url: string) => {
```
bkrem: Nit: not a big deal but feel this would be more accurate and avoids kind of shadowing the `fetch` keyword (even though it's a class method here)

```suggestion
  private fetchAttestation = async (attestationId: string, url: string) => {
```

packages/core/src/controllers/verify.ts
```diff
@@ -81,40 +116,40 @@ export class Verify extends IVerify {
...
+        iframe.src = `${this.verifyUrl}/${this.projectId}`;
+        iframe.style.display = "none";
+        document.body.append(iframe);
+        this.iframe = iframe;
+        iframeOnLoadResolve = resolve;
+      }),
+      new Promise((_, reject) =>
+        setTimeout(() => {
+          window.removeEventListener("message", onMessage);
+          reject("iframe load timeout");
```
bkrem:
```suggestion
          reject("verify iframe load timeout");
```

bkrem (approved):
💪

packages/core/src/controllers/verify.ts
ganchoradkov: It's done so intentionally as we have to catch both attempts (main url & fallback) unlike in the WS connection where the fallback rejection is unhandled on purpose

ganchoradkov:
>What was the context for this change? Was this necessary/fixing an issue or an improvement?

@bkrem Different browsers are inconsistent with onload/onerror events making it very difficult to actually detect when the iframe fails to load due to it being cross-origin. 
For example in Chrome, onload is emitted even if you use `non-existent-url` because the iframe is not considered a resource like `img`

PR #3553, opened 2023-08-28
chore: expose EthereumProviderOptions interface

Jurshsmith:
## Description

This will enable consumers to extend this interface for other use cases aside from initializing the EthereumProvider in the consumer's application context.

## Type of change

## How has this been tested?

No tests since it is a simple change.

## Checklist

Just a question: I am curious if there is any non-obvious underlying issue if this interface gets exposed (?)

### DIFF
--- a/providers/ethereum-provider/src/index.ts
+++ b/providers/ethereum-provider/src/index.ts
@@ -1,5 +1,5 @@
 import { EthereumProvider as Provider } from "./EthereumProvider";
 export const EthereumProvider = Provider;
-export type { RpcEvent, RpcMethod } from "./EthereumProvider";
+export type { EthereumProviderOptions, RpcEvent, RpcMethod } from "./EthereumProvider";
 export * from "./constants/rpc";
 export default Provider;


JonathanConn:
The export can be found here
https://github.com/WalletConnect/walletconnect-monorepo/blob/df685268af77be3ce30e7becbcf2c32546acd827/providers/ethereum-provider/src/EthereumProvider.ts#L205

Jurshsmith:
> The export can be found here

@JonathanConn It is indeed exposed from `EthereumProvider` but not from the entry point `index.ts`. 

There's currently no way to import this into my project. Maybe there is something I am missing?

JonathanConn:
My mistake. What are you trying to use the EthereumProviderOptions for?

Jurshsmith:
@JonathanConn A hybrid operation when initializing any web3 provider.

JonathanConn:
> @JonathanConn A hybrid operation when initializing any web3 provider.

What do you mean by hybrid operation?

JonathanConn (approved):
My mistake, good to go 👍

Jurshsmith:
> What do you mean by hybrid operation?

It is generally good to allow extensibility with exposed interfaces. It opens up a lot of use cases.

A simple use case would be an `EthereumProvider` factory that takes a partial subset of `EthereumProviderOptions` used across different in-house dApps.

Issue #3084, opened 2023-07-10
Security improvement: Replace stablelib with audited noble libraries


paulmillr:
noble cryptography (https://paulmillr.com/noble/) is a high-security, auditable set of contained cryptographic libraries with following features:

- No dependencies, protection against supply chain attacks
- Auditable TypeScript / JS code
- Supported on all major platforms
- Releases are signed with PGP keys and built transparently with NPM provenance

Furthermore, noble-hashes were created with a grant by Ethereum Foundation. hashes and curves have been independently audited by 3rd party entity.

You have following deps:

https://github.com/WalletConnect/walletconnect-monorepo/blob/4ee518cf58b7aa0c59289999e7443918b9d14706/packages/utils/package.json#L33-L38

They can be replaced with:

```json
"@noble/ciphers": "0.1.3",
"@noble/curves": "1.1.0",
"@noble/hashes": "1.3.1",
```

The usage is as following:

```js
import { chacha20poly1305 } from "@noble/ciphers/chacha";
import { hkdf } from "@noble/hashes/hkdf";
import { sha256 } from "@noble/hashes/sha256";
import { x25519 } from "@noble/curves/ed25519";
import { randomBytes as random } from "@noble/hashes/utils";
```

paulmillr:
I assume this repo is unmaintained and you don't really care, since there was no response for 2.5 months (#2272)


@pedrouid @ganchoradkov @0xAsimetriq @bkrem

hazae41:
Sad to see no response to this

I will build a custom WalletConnect lib, it will include adapters for Noble, WebAssembly and Stablelib so you will be able to choose which one you prefer (and you will be able to do it at runtime, e.g. depending on the user connection speed, tradeoff performances against bundle size)

By the way, it will also support arbitrary transport protocols like WebSocket-over-Tor instead of only native WebSockets

EDIT: this might explain why this issue is ignored
<img width="500" src="https://github.com/WalletConnect/walletconnect-monorepo/assets/4405263/54c427e3-9482-42a8-85af-b5cc9f75c15a" />

pedrouid:
Hi @paulmillr @hazae41, can you guys chill out?

I have no connection to StableLib whatsoever 😅 so no need to start conspiracies just because I have one-line commit correcting a typo inside a comment: https://github.com/StableLib/stablelib/commit/cad7e89f9b4df831babc0342f17d1508c31a604e

[...]

PR #3621, opened 2023-09-09
feat: core replay protection

pedrouid:
## Description

This introduces replay protection for each typed enveloped based on the topic

## Type of change

## How has this been tested?

THIS IS STILL MISSING TESTING!

## Checklist

### DIFF
--- a/packages/core/src/controllers/crypto.ts
+++ b/packages/core/src/controllers/crypto.ts
@@ -106,13 +106,13 @@ export class Crypto implements ICrypto {
   public encode: ICrypto["encode"] = async (topic, payload, opts) => {
     this.isInitialized();
     const params = validateEncoding(opts);
-    const protectedPayload = { topic, ...payload };
-    const message = safeJsonStringify(protectedPayload);
     if (isTypeOneEnvelope(params)) {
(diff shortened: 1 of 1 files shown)


packages/core/src/controllers/crypto.ts
```diff
@@ -130,6 +131,12 @@ export class Crypto implements ICrypto {
       const symKey = this.getSymKey(topic);
       const message = decrypt({ symKey, encoded });
       const payload = safeJsonParse(message);
+      if (typeof payload.topic !== "undefined" && payload.topic !== topic) {
```
ganchoradkov:
```suggestion
      if (typeof payload.topic !== "undefined") {
```
I'm not sure whether `payload.topic !== topic` should be in the first `if`. It makes the second `if` redundant as both conditions are the same.

paulmillr:
@pedrouid can you check out security improvement issue instead of ignoring it? #3084

pedrouid:
> @pedrouid can you check out security improvement issue instead of ignoring it? #3084

@paulmillr please do not comment on unrelated PRs to elevate the awareness of another issue

This type of commentary is unproductive so please see my response on #3084

PR #3688, opened 2023-09-21
feat: adds `customStoragePrefix` option to Core

bkrem:
## Description

* Allows for a client instantiating core to set a custom storagePrefix, which allows for Core controllers to be namespaced under different storage keys than the default.
* This is useful for scenarios where two Core instances exist side by side, because the higher level abstractions running `Core` are unable to share a single instance.

## Type of change

## How has this been tested?

- Unit tests
- Canary: `2.10.1-c362fdc6`
	- Tested against web3inbox app

## Checklist

### DIFF
--- a/packages/types/src/core/core.ts
+++ b/packages/types/src/core/core.ts
@@ -19,6 +19,7 @@ export declare namespace CoreTypes {
     keychain?: IKeyChain;
     storage?: IKeyValueStorage;
     storageOptions?: KeyValueStorageOptions;
+    customStoragePrefix?: string;
   }
 
   interface Metadata {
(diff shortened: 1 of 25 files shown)


packages/types/src/core/core.ts
```diff
@@ -19,6 +19,7 @@ export declare namespace CoreTypes {
     keychain?: IKeyChain;
     storage?: IKeyValueStorage;
     storageOptions?: KeyValueStorageOptions;
+    customStoragePrefix?: string;
```
ganchoradkov, line 22: This would've been nice to be included `storageOptions?` but might be difficult to achieve

ganchoradkov (approved):
💯 💯

devceline (approved):
LGTM

packages/types/src/core/core.ts
bkrem: Yeah I see that in terms of opts here. But that would create other aspects that don't really make sense, i.e. the `KeyValueStorageOptions` are specifically consumed by the `KeyValueStorage` class, but these keys are actually set by the controllers, not `KeyValueStorage`.

OkponkoXchange (commented):
[**image**](https://m.facebook.com/checkpoint/?wtsid=rdr_0lnSvBNLF0crwlDad&locale2=en_GB&_rdr)

PR #3830, opened 2023-10-26
feat(experimental): use shared core reference via globals

bkrem:
## Description

- This is an experimental PoC of sharing a `Core` instance by:
	- identifying the current global scope (`window`, `global`, `globalThis`)
	- Binding the instantiated `Core` instance to `globalScope.__walletconnect_core__`
	- If the binding already exists, instruct the current core controller to consume this global reference instead and bind its controllers from it, instead of instantiating its own.
	- Will only bind to an existing core reference if the `CoreOpts` match, else instantiate a separate one.

- I've made some accommodations due to technicalities of this current approach:
	- Using `console.log` instead of `this.logger` since it was erroring at runtime regarding `this.logger` not yet being defined at this stage of initialisation
	- Creating a separate `this.self = this` reference: This is primarily to work around Typescript complaining about direct assignments to `this`.

## Type of change

## How has this been tested?

- Release on `experimental` dist for all monorepo pkgs: `2.10.4-184f2234`
- I've done some manual validation that two cores can run in parallel and share a relayer, WS connection etc successfully this way, but it's by no means validated as stable.
- Some tests are currently still failing due to the attempt to share cores, which I've overridden where possible in test setups/teardowns for now.

## Checklist

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -4,5 +4,5 @@
     "packages/*",
     "providers/*"
   ],
-  "version": "2.10.4"
+  "version": "2.10.4-184f2234"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -6283,6 +6283,40 @@
         "node": ">=16"
       }
     },
+    "node_modules/@walletconnect/auth-client/node_modules/@walletconnect/types": {
+      "version": "2.10.4",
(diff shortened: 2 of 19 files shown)


glitch-txs:
can we close this one? @bkrem

bkrem:
> can we close this one? @bkrem

Yes but please don't delete the branch 🙏 Will close.

PR #3847, opened 2023-10-30
Feat/send bundleid

ignaciosantise:
## Description

Send application's bundle id as a new param in relay url string in react native apps

## Type of change

## How has this been tested?

- [x] Tested in Android projects with old and new architecture
- [x] Tested in iOS projects with old and new architecture
- [x] Tested in a Expo project

## Checklist

   - [x] https://github.com/WalletConnect/walletconnect-utils/pull/146

* Built code based on https://github.com/callstack/react-native-builder-bob


llbartekll (approved):

jakubuid (approved):

packages/utils/src/misc.ts
```diff
@@ -58,6 +58,21 @@ export function getEnvironment(): string {
...
 }
 
+export function getBundleId(): string | undefined {
+  try {
+    if (isReactNative() && typeof global !== "undefined" && (global as any)?.Application) {
+      if (typeof (global as any).Application?.getConstants === "function") {
+        return (global as any).Application?.getConstants().applicationId;
+      } else if (typeof (global as any).Application?.applicationId === "string") {
+        return (global as any).Application?.applicationId;
+      }
```
ganchoradkov:
```suggestion
return (global as any).applicationId
```
isn't it cleaner to just assign the applicationId to the `global` and move these conditions in the `react-native-compat`? This way all of the RN logic/conditions would be isolated to `react-native-compat`

ganchoradkov (changes requested):

packages/utils/src/misc.ts
ignaciosantise: @ganchoradkov Yes it is cleaner, but i thought that maybe is also easier for malicious devs to override the appId in react-native-compat code? Maybe im overthinking it.

ganchoradkov (approved):
💯

PR #3939, opened 2023-11-16
feat: deeplink fallback

ganchoradkov:
## Description
Added an explicit `localStorage` fallback when getting a deeplink main storage

## Type of change

## How has this been tested?
canary `2.10.5-canary-98641be`

## Checklist

### DIFF
--- a/packages/utils/src/misc.ts
+++ b/packages/utils/src/misc.ts
@@ -406,6 +406,7 @@ export async function getDeepLink(store: IKeyValueStorage, key: string) {
     if (!isBrowser()) return;
     return localStorage.getItem(key) as string;
   } catch (err) {
+    // eslint-disable-next-line no-console
     console.error(err);
   }
 }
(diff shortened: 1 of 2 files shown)


packages/utils/src/misc.ts
```diff
@@ -395,3 +396,17 @@ export async function handleDeeplinkRedirect({
     console.error(err);
   }
 }
+
+export async function getDeepLink(store: IKeyValueStorage, key: string) {
+  try {
+    const deepLink = await store.getItem(key);
+    if (deepLink) return deepLink;
+
+    // check localStorage as fallback
```
bkrem, line 405: LGTM code-wise, but not 100% clear to me why we need this behaviour? Assumption being that `store` now refers to IndexedDB by default and we want to ensure we still check localStorage too in a browser env?

bkrem (approved):

packages/utils/src/misc.ts
ganchoradkov: yes, some modal versions set the deeplink in `localStorage` directly and are not using the `Core's storage` so this fallback covers them

Issue #3906, opened 2023-11-10
Double session proposal event


aditbisa:
**Describe the bug**
We use `SignClient` from backend and store the session proposal to our database.
We rely on `session_proposal` event for this.

But the `session_proposal` event called twice on new pairing.

Fortunately, our Dapp add `sessionProperties` into the proposal.
And we are able to identify the culprit of this double `session_proposal` event.
Its this [`onPairingCreated` function](https://github.com/WalletConnect/walletconnect-monorepo/blob/124a174d2dd905560b80e8a2c21bf27421482d4b/packages/sign-client/src/controllers/engine.ts#L1099C32-L1099C32) where it call `onSessionProposeRequest` with imitation data from the relay.

Our `session_proposal` handler receive 2 data:
1. From the relay with custom `sessionProperties`
2. From `onPairingCreated` without it

**SDK Version (if relevant)**
 - Client: JS
 - Version: 2.10

**To Reproduce**
1. From Wallet SignClient, listen `session_proposal` and dump the data.
2. Connect new Dapp
3. See double data from `session_proposal` handler

**Expected behavior**
`onPairingCreated` should not compose the proposal by itself.

**Screenshots**
![image](https://github.com/WalletConnect/walletconnect-monorepo/assets/16781093/f6699d5c-5fd1-4def-8ddc-d5b35f4c1e21)

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser: NodeJs

**Smartphone (please complete the following information):**

**Additional context**

ganchoradkov:
resolved with https://github.com/WalletConnect/walletconnect-monorepo/pull/3944

PR #3944, opened 2023-11-17
fix: `pair` flow

ganchoradkov:
## Description
Moved storing `symKey` & subscribing to the pairing topic after the pairing itself has been stored to avoid cases where the topic can start to receive payloads before the pairing has been saved & `pair()` method is resolved   

## Type of change

## How has this been tested?
tests

## Checklist

### DIFF
--- a/packages/core/test/pairing.spec.ts
+++ b/packages/core/test/pairing.spec.ts
@@ -293,21 +293,20 @@ describe("Pairing", () => {
       const { uri } = await coreA.pairing.create();
       const { topic } = parseUri(uri);
       coreB.pairing.events.on(PAIRING_EVENTS.create, () => {
-        expect(coreB.pairing.pairings.keys.length).toBe(1);
-        expect(coreB.pairing.pairings.values[0].topic).toEqual(topic);
         pairingCreatedEventTime = performance.now();
         pairingCreatedEvent = true;
       });
 
-      coreB.relayer.subscriber.events.on(SUBSCRIBER_EVENTS.created, (subscription) => {
+      coreB.relayer.subscriber.events.on(SUBSCRIBER_EVENTS.created, () => {
         subscriptionCreatedEventTime = performance.now();
         subscriptionCreatedEvent = true;
-        expect(subscription.topic).toEqual(topic);
       });
 
       coreB.pairing.pair({ uri });
       await waitForEvent(() => pairingCreatedEvent);
       await waitForEvent(() => subscriptionCreatedEvent);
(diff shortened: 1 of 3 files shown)


ganchoradkov:
indirectly resolves https://github.com/WalletConnect/walletconnect-monorepo/issues/3906, where payloads are received before `pairing_create` event is emitted

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1114,6 +1114,7 @@ export class Engine extends IEngine {
           optionalNamespaces: proposal.optionalNamespaces,
           relays: proposal.relays,
           proposer: proposal.proposer,
+          sessionProperties: proposal.sessionProperties,
```
devceline, line 1117: Why are we adding this here?  Is it to allow to mimic how #3906  was caught?

packages/core/test/pairing.spec.ts
```diff
@@ -277,4 +277,38 @@ describe("Pairing", () => {
...
+    });
+    it("should store pairing before subscribing to its topic", async () => {
+      let pairingCreatedEvent = false;
+      let pairingCreatedEventTime = 0;
+      let subscriptionCreatedEvent = false;
+      let subscriptionCreatedEventTime = 0;
+      const { uri } = await coreA.pairing.create();
+      const { topic } = parseUri(uri);
+      coreB.pairing.events.on(PAIRING_EVENTS.create, () => {
+        expect(coreB.pairing.pairings.keys.length).toBe(1);
```
devceline: In `vitest`, an `expect` in an event listener throws but you never get to see what went wrong. You just get a timeout here because `waitForEvent` never resolves below. 

I see this is used everywhere in tests so I wonder if this just happens for me :sweat: .

devceline (approved):

packages/sign-client/src/controllers/engine.ts
ganchoradkov: yes

packages/core/test/pairing.spec.ts
ganchoradkov: you're correct. It does happen to me as well. Moved them out

packages/sign-client/src/controllers/engine.ts
devceline: Ah okay - then we should update [this](https://docs.walletconnect.com/api/sign/wallet-usage#setting-up-event-listeners) in the docs I think

Issue #3813, opened 2023-10-23
An uncaught exception when pairing with a WC v1 URI


katspaugh:
When calling `web3Wallet.core.pairing.pair({ uri })` with a v1 URI, it throws an uncaught excpetion even if the `pair` call itself is caught.

<img width="1250" alt="Screenshot 2023-10-23 at 11 10 32" src="https://github.com/WalletConnect/walletconnect-monorepo/assets/381895/04f5b909-10a9-44e4-891c-fe90accff56a">

This can be reproduced when connecting for example to 1inch with their v1 link.

I realize v1 isn't supported but this error should be probably handled properly. Please let me know if I'm missing something, maybe this exception can be caught on our side.

katspaugh:
My bad, the error is propagated by the SDK correctly.

amiryadid-fb:
Hey @katspaugh ,
I'm not from the WC team, but I'm pretty sure V1 is deprecated and can no longer be used using WC bridge

katspaugh:
I know that but it shouldn't throw an uncaught exception. The promise returned by `web3Wallet.core.pairing.pair` should be rejected instead.

PR #3960, opened 2023-11-21
fix: URI validation

ganchoradkov:
## Description
Fixed a bug where when invalid URI is provided such as `v1` uri, or any other that is missing `relay-protocol` was resulting in unhandled exception instead of returning the rejection 
## Type of change

## How has this been tested?
tests 

## Fixes/Resolves (Optional)
https://github.com/WalletConnect/walletconnect-monorepo/issues/3813

## Checklist

### DIFF
--- a/packages/core/src/controllers/pairing.ts
+++ b/packages/core/src/controllers/pairing.ts
@@ -387,6 +387,10 @@ export class Pairing implements IPairing {
       const { message } = getInternalError("MISSING_OR_INVALID", `pair() uri: ${params.uri}`);
       throw new Error(message);
     }
+    if (!parseUri(params.uri)?.relay?.protocol) {
+      const { message } = getInternalError("MISSING_OR_INVALID", `pair() uri#relay-protocol`);
+      throw new Error(message);
+    }
   };
 
   private isValidPing = async (params: { topic: string }) => {
--- a/packages/core/src/controllers/pairing.ts
+++ b/packages/core/src/controllers/pairing.ts
@@ -387,8 +387,13 @@ export class Pairing implements IPairing {
       const { message } = getInternalError("MISSING_OR_INVALID", `pair() uri: ${params.uri}`);
       throw new Error(message);
     }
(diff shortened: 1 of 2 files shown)


packages/core/src/controllers/pairing.ts
```diff
@@ -388,6 +388,10 @@ export class Pairing implements IPairing {
       const { message } = getInternalError("MISSING_OR_INVALID", `pair() uri: ${params.uri}`);
       throw new Error(message);
     }
+    if (!parseUri(params.uri)?.relay?.protocol) {
+      const { message } = getInternalError("MISSING_OR_INVALID", `pair() uri#relay-protocol`);
+      throw new Error(message);
+    }
```
devceline: Non blocking: I know `symKey` exists in both `v1` and `v2` but if we're handling this specific param, would be nice and maybe more future proof of we also handled `symKey` too?

devceline (approved):

packages/core/test/pairing.spec.ts
```diff
@@ -205,6 +205,13 @@ describe("Pairing", () => {
           "Missing or invalid. pair() uri: undefined",
         );
       });
+      it("throws when uri missing relay protocol is provided", async () => {
+        await expect(
+          coreA.pairing.pair({
+            uri: "wc:e9d6ef98-6b65-490b-8726-a21e1afb181d@1?bridge=https%3A%2F%2Fwalletconnect.com&key=73f096cb97aaee97b3d9871ced35fdce1668e652db3d39423ea6cd22e14528bf",
```
devceline:
```suggestion
        // Using v1 pairing URI as it is unsupported
        const v1PairingUri = "wc:e9d6ef98-6b65-490b-8726-a21e1afb181d@1?bridge=https%3A%2F%2Fwalletconnect.com&key=73f096cb97aaee97b3d9871ced35fdce1668e652db3d39423ea6cd22e14528bf"
        await expect(
          coreA.pairing.pair({
            uri: v1PairingUri,
```

PR #3996, opened 2023-11-29
sendAsync method result response formatted

### LINKED ISSUE #3984: JSON RPC response is in invalid format
In provider.ts file Result should be in a specific type which is 

**formatJsonRpcResult**
```
return {
    id,
    jsonrpc: "2.0",
    result,
  };
```
@walletconnect/ethereum-provider giving response.result directly to user while any method calling like *eth_call* , *eth_sendTransaction*, etc.

example code

```
const sendTransactionAsync = async (txData) => {
        try {
            const txHash = await sendRawPayloadAsync({ method: 'eth_sendTransaction', params: [txData] });
            return txHash;
        } catch (error) {
            console.error("Error in sendTransactionAsync:", error);
            throw error;
        }
    }
```

image

```
 protected async requestStrict<Result = any, Params = any>(     <----- ( Result should be specific type formatJsonRpcResult )
    request: JsonRpcRequest<Params>,
    context?: any,
  ): Promise<Result> {
    return new Promise(async (resolve, reject) => {
      if (!this.connection.connected) {
        try {
          await this.open();
        } catch (e) {
          reject(e);
        }
      }
      this.events.on(`${request.id}`, (response) => {
        if (isJsonRpcError(response)) {
          reject(response.error);
        } else {
          resolve(response.result);  <----- ( It should be in specific type formatJsonRpcResult )
        }
      });
      try {
        await this.connection.send(request, context);
      } catch (e) {
        reject(e);
      }
    });
  }

```

versions: 

Web3: 1.2.11
@walletconnect/Ethereum-provider: 2.8.5
system: ubuntu

kr-nikunj:

## Description

fixed JSON RPC response in specific format, issue: #3984

```
return {
    id,
    jsonrpc: "2.0",
    result,
  };
```

## Type of change

## How has this been tested?

Using manually testing of logs and and running the test file

Use this table template to show examples of your changes:

| Before       | After        |
| ------------ | ------------ |
| image | image |

## Checklist

### DIFF
--- a/providers/universal-provider/src/UniversalProvider.ts
+++ b/providers/universal-provider/src/UniversalProvider.ts
@@ -1,5 +1,6 @@
 import SignClient, { PROPOSAL_EXPIRY_MESSAGE } from "@walletconnect/sign-client";
 import { SessionTypes } from "@walletconnect/types";
+import { JsonRpcResult } from "@walletconnect/jsonrpc-types";
 import { getSdkError, isValidArray, parseNamespaceKey } from "@walletconnect/utils";
 import { getDefaultLoggerOptions, Logger, pino } from "@walletconnect/logger";
 import {
@@ -34,6 +35,7 @@ import {
 
 import { RELAY_URL, LOGGER, STORAGE, PROVIDER_EVENTS } from "./constants";
 import EventEmitter from "events";
+import { formatJsonRpcResult } from "@walletconnect/jsonrpc-utils";
 
 export class UniversalProvider implements IUniversalProvider {
   public client!: SignClient;
@@ -87,12 +89,13 @@ export class UniversalProvider implements IUniversalProvider {
 
   public sendAsync(
     args: RequestArguments,
-    callback: (error: Error | null, response: any) => void,
+    callback: (error: Error | null, response: JsonRpcResult) => void,
     chain?: string | undefined,
   ): void {
+    const id = new Date().getTime();
     this.request(args, chain)
-      .then((response) => callback(null, response))
-      .catch((error) => callback(error, undefined));
+      .then((response) => callback(null, formatJsonRpcResult(id, response)))
+      .catch((error) => callback(error, undefined as any));
   }
 
   public async enable(): Promise<ProviderAccounts> {
--- a/providers/ethereum-provider/src/EthereumProvider.ts
+++ b/providers/ethereum-provider/src/EthereumProvider.ts
@@ -10,6 +10,7 @@ import {
 } from "./types";
 import { Metadata, Namespace, UniversalProvider } from "@walletconnect/universal-provider";
 import { SessionTypes, SignClientTypes } from "@walletconnect/types";
+import { JsonRpcResult } from "@walletconnect/jsonrpc-types";
 import {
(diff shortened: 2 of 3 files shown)


providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -87,12 +89,13 @@ export class UniversalProvider implements IUniversalProvider {
...
     args: RequestArguments,
-    callback: (error: Error | null, response: any) => void,
+    callback: (error: Error | null, response: JsonRpcResult) => void,
     chain?: string | undefined,
   ): void {
+    const id = new Date().getTime();
     this.request(args, chain)
-      .then((response) => callback(null, response))
-      .catch((error) => callback(error, undefined));
+      .then((response) => callback(null, formatJsonRpcResult(id, response)))
```
ganchoradkov, line 97: >.then((response) => callback(null, formatJsonRpcResult(id, response)))

Do we need to format the response here? Afaik all responses are already formatted as `jsonRpcResult` by the responding client

ganchoradkov (changes requested):

providers/universal-provider/src/UniversalProvider.ts
kr-nikunj: Yes, the client is not responding in a formatted manner

kr-nikunj: image


@ganchoradkov, Facing the same issue while running these test cases on the main branch, please check it

ganchoradkov: please make sure you run

- `npm i` in root to fetch new packages
- `npm run reset` to rebuild the monorepo
for some reason the relay is rejecting your jwt and/or projectId

also your system's datetime as jwt is valid for 1 day so if you datetime is forward/back in time might cause it

kr-nikunj: Hey @ganchoradkov, I tried that also and ran it in a local file but it gave the **same error**. For passing the test case there might be a change in the workflow file of the repo where it is not getting environment variables correctly. 

If possible can I DM you on Discord or any other platform to resolve this issue?

ganchoradkov: `@GanchoRadkov` on telegram

kr-nikunj: @ganchoradkov, can you please tell me where you set environment variables? 

image
in workflow file: 
https://github.com/WalletConnect/walletconnect-monorepo/actions/runs/7084385293/workflow?pr=3996

Because I run test cases locally and it's running **properly**. and facing issue here only.

kr-nikunj:
In the universal provider's test file ( walletconnect-monorepo/providers/universal provider/test/shared/constant.ts), If I put my project ID and relay id directly, it passes all test cases. for getting the environment variable it failed test cases.

image

image

This is run while I define my project ID directly to a constant file

image

please check it @ganchoradkov

ganchoradkov (approved):
thanks for fixing @kr-nikunj 💯 🥇

PR #3997, opened 2023-11-29
feat: decrypted notifications

ganchoradkov:
## Description

This includes
- new `Echo` controller in `Core` that allows devices to be registered with Echo server
- new static property in `Web3Wallet` called `notifications` that allows for decrypting push payloads & getting the peer's metadata without initializing the whole client

### Usage

- Registering device
`core.echo.registerDeviceToken(params)` or shorthand `web3wallet.registerDeviceToken(params)`

- Decrypting notification payload
```typescript
 // handling firebase remote notificaiton
  messaging().setBackgroundMessageHandler(async notification => {
    const { topic, message } = notification.data;
    // decrypt the payload
    const decrypted = await Web3Wallet.notifications.decryptMessage({ topic, encryptedMessage: message })
    // get metadata (name, description, icons, etc) of the sender
    const metadata = await Web3Wallet.notifications.getMetadata({ topic })
    
    // emit local notification
    ...
  });
```

## Type of change

## How has this been tested?
tests
dogfooding

## Checklist

specs https://github.com/WalletConnect/walletconnect-specs/pull/147

### DIFF
--- a/packages/core/src/controllers/echo.ts
+++ b/packages/core/src/controllers/echo.ts
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import { generateChildLogger, Logger } from "@walletconnect/logger";
 import { IEcho } from "@walletconnect/types";
 import { ECHO_CONTEXT, ECHO_URL } from "../constants";
@@ -8,7 +7,6 @@ export class Echo extends IEcho {
   constructor(public projectId: string, public logger: Logger) {
     super(projectId, logger);
     this.logger = generateChildLogger(logger, this.context);
-    console.log("Echo constructor called", projectId);
   }
 
   public registerDeviceToken: IEcho["registerDeviceToken"] = async (params) => {
@@ -16,8 +14,7 @@ export class Echo extends IEcho {
 
     const echoUrl = `${ECHO_URL}/${this.projectId}/clients`;
 
-    console.log("register called", params, echoUrl);
-    const echoResponse = await fetch(echoUrl, {
+    await fetch(echoUrl, {
       method: "POST",
       headers: {
         "Content-Type": "application/json",
@@ -29,13 +26,5 @@ export class Echo extends IEcho {
         always_raw: enableAlwaysDecrypted,
       }),
     });
-    console.log("echo body", {
-      client_id: clientId,
-      type: notificationType,
-      token,
-      always_raw: enableAlwaysDecrypted,
-    });
-
-    console.log("echoResponse", echoResponse);
   };
 }
--- a/packages/core/src/controllers/echo.ts
+++ b/packages/core/src/controllers/echo.ts
@@ -1,7 +1,6 @@
 import { generateChildLogger, Logger } from "@walletconnect/logger";
 import { IEchoClient } from "@walletconnect/types";
 import { ECHO_CONTEXT, ECHO_URL } from "../constants";
-import "isomorphic-unfetch";
 
 export class EchoClient extends IEchoClient {
   public readonly context = ECHO_CONTEXT;
--- a/packages/core/src/controllers/echo.ts
+++ b/packages/core/src/controllers/echo.ts
@@ -1,0 +1,41 @@
+/* eslint-disable no-console */
+import { generateChildLogger, Logger } from "@walletconnect/logger";
+import { IEcho } from "@walletconnect/types";
+import { ECHO_CONTEXT, ECHO_URL } from "../constants";
+
+export class Echo extends IEcho {
+  public readonly context = ECHO_CONTEXT;
+  constructor(public projectId: string, public logger: Logger) {
+    super(projectId, logger);
+    this.logger = generateChildLogger(logger, this.context);
+    console.log("Echo constructor called", projectId);
+  }
+
+  public registerDeviceToken: IEcho["registerDeviceToken"] = async (params) => {
+    const { clientId, token, notificationType, enableAlwaysDecrypted = false } = params;
+
+    const echoUrl = `${ECHO_URL}/${this.projectId}/clients`;
+
+    console.log("register called", params, echoUrl);
+    const echoResponse = await fetch(echoUrl, {
+      method: "POST",
+      headers: {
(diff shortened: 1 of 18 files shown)


packages/core/src/controllers/echo.ts
```diff
@@ -0,0 +1,30 @@
+import { generateChildLogger, Logger } from "@walletconnect/logger";
+import { IEcho } from "@walletconnect/types";
+import { ECHO_CONTEXT, ECHO_URL } from "../constants";
+
+export class Echo extends IEcho {
```
devceline: nit: `EchoClient`, `IEchoClient`

devceline: nit: Why did we make an entire class for this? Couldn't we have just had a function defined somewhere to register the token? I feel like it's just extra complexity and more state to store needlessly

packages/web3wallet/src/controllers/engine.ts
```diff
@@ -98,6 +98,11 @@ export class Engine extends IWeb3WalletEngine {
     return this.authClient.formatMessage(params, iss);
   };
 
+  // Push //
```
devceline, line 101:
```suggestion
  // ---------- Push -------------------------------------------------- //
```

packages/web3wallet/src/utils/notifications.ts
```diff
@@ -0,0 +1,24 @@
...
+export const Notifications: Web3WalletTypes.INotifications = {
+  decryptMessage: async (params) => {
+    const core = new Core({
+      storageOptions: params.storageOptions,
+      storage: params.storage,
+    });
+    await core.crypto.init();
+    return core.crypto.decode(params.topic, params.encryptedMessage);
+  },
+  getMetadata: async (params) => {
```
devceline, line 18:
```suggestion
  getDappMetadata: async (params) => {
```

packages/core/src/core.ts
```diff
@@ -77,6 +78,7 @@ export class Core extends ICore {
     });
     this.pairing = new Pairing(this, this.logger);
     this.verify = new Verify(this.projectId || "", this.logger);
+    this.echo = new Echo(this.projectId || "", this.logger);
```
devceline: Why are we putting this in core? `Sign/Auth/Chat` won't make a use of this, no?

packages/core/src/controllers/echo.ts
```diff
@@ -0,0 +1,30 @@
...
+    super(projectId, logger);
+    this.logger = generateChildLogger(logger, this.context);
+  }
+
+  public registerDeviceToken: IEcho["registerDeviceToken"] = async (params) => {
+    const { clientId, token, notificationType, enableEncrypted = false } = params;
+
+    const echoUrl = `${ECHO_URL}/${this.projectId}/clients`;
+
+    await fetch(echoUrl, {
```
devceline, line 18: Don't we need to use `isomorphic-unfetch` or something here? Or is it already polyfilled

devceline (approved):
LGTM - left questions and nits. All non blocking

packages/core/src/controllers/echo.ts
ganchoradkov: the idea is to have dedicated handling allowing functionality to be reused and expanded as needed in the future

packages/web3wallet/src/utils/notifications.ts
ganchoradkov: naming is according to agreed spec

packages/core/src/controllers/echo.ts
bkrem: I'm actually kind of surprised this hasn't come up before here where there's already a `fetch` call: https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/core/src/controllers/verify.ts#L103

Is the `Verify` controller called only in browsers?

packages/sign-client/src/index.ts
```diff
@@ -1,5 +1,6 @@
 import { SignClient as Client } from "./client";
-
+import { Session } from "./controllers/session";
+export const SessionStore = Session;
```
bkrem: Nit: would be cleaner to have this side by side with the other explicit `export const SignClient` assignment below

packages/web3wallet/src/types/client.ts
```diff
@@ -41,7 +42,21 @@ export declare namespace Web3WalletTypes {
     name?: string;
   }
 
-  type Metadata = AuthClientTypes.Metadata;
+  type Metadata = CoreTypes.Metadata;
```
bkrem, line 45: Is the underlying type identical before/after?

packages/web3wallet/src/utils/notifications.ts
```diff
@@ -0,0 +1,24 @@
+import { Core } from "@walletconnect/core";
+import { Web3WalletTypes } from "../types";
+import { SessionStore } from "@walletconnect/sign-client";
+
+export const Notifications: Web3WalletTypes.INotifications = {
+  decryptMessage: async (params) => {
+    const core = new Core({
```
bkrem: Bit concerned regarding memory and creating a new Core reference on every `decryptMessage` call. Is there any way to avoid this and init the Core once? Guessing this may have to be stateless right.

bkrem (approved):
🔥 🔥

packages/web3wallet/src/types/client.ts
ganchoradkov: yes, in both instances it's coming from `CoreTypes`

packages/web3wallet/src/utils/notifications.ts
devceline: I don't think we have a way around this because our modules are not stateless/pure - crypto depends on `core` and having a store AFAIK

ganchoradkov: Good point 💯  I will thinker a solution as we do need to reinitialize to update the data

bkrem (approved):

PR #4072, opened 2023-12-21
fix: responding to request on disconnected session

### LINKED ISSUE #4063: Invalid session requests remain in the queue and new requests cannot be processed
**Describe the bug**
If the user disconnects while there are still pending requests and my wallet responds to them, an exception is thrown [in this line](https://github.com/WalletConnect/walletconnect-monorepo/blob/51c3a97db5895835e9fc997bee14a1de87202116/packages/sign-client/src/controllers/engine.ts#L384) which results in the [cleanup function](https://github.com/WalletConnect/walletconnect-monorepo/blob/51c3a97db5895835e9fc997bee14a1de87202116/packages/sign-client/src/controllers/engine.ts#L392) not being called. This means that those requests are not marked as completed and new requests cannot be processed. Only after 5 minutes will the queue get cleaned and new requests can be processed again.

**SDK Version (if relevant)**
 - Client: JS
 - Version 2.10.6

**Expected behavior**
The queue needs to be cleaned from invalid requests when such are detected, instead of waiting 5 minutes.

ganchoradkov:
## Description
Fixed a bug where trying to respond  to a session request on already disconnected session was causing the request queue to hang for the remaining expiry of that request

resolves https://github.com/WalletConnect/walletconnect-monorepo/issues/4063

## Type of change

## How has this been tested?
tests
dogfooding in example apps

## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -381,6 +381,13 @@ export class Engine extends IEngine {
 
   public respond: IEngine["respond"] = async (params) => {
     await this.isInitialized();
+
+    // if the session is already disconnected, we can't respond to the request so we need to delete it
+    await this.isValidSessionTopic(params.topic).catch((error) => {
+      this.cleanupAfterResponse(params);
+      throw error;
+    });
+
     await this.isValidRespond(params);
     const { topic, response } = params;
     const { id } = response;
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -381,13 +381,6 @@ export class Engine extends IEngine {
 
   public respond: IEngine["respond"] = async (params) => {
     await this.isInitialized();
-
-    // if the session is already disconnected, we can't respond to the request so we need to delete it
-    await this.isValidSessionTopic(params.topic).catch((error) => {
-      this.cleanupAfterResponse(params);
-      throw error;
-    });
-
     await this.isValidRespond(params);
     const { topic, response } = params;
     const { id } = response;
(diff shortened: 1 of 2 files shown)


packages/sign-client/src/controllers/engine.ts
```diff
@@ -1039,7 +1044,7 @@ export class Engine extends IEngine {
   };
 
   private cleanupAfterResponse = (params: EngineTypes.RespondParams) => {
-    this.deletePendingSessionRequest(params.response.id, { message: "fulfilled", code: 0 });
+    this.deletePendingSessionRequest(params.response?.id, { message: "fulfilled", code: 0 });
```
bkrem: Why the change to optional chaining here? Can `deletePendingSessionRequest` handle being passed `undefined` if the optional chain operator defaults? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1388,7 +1393,11 @@ export class Engine extends IEngine {
       throw new Error(message);
     }
     const { topic, response } = params;
-    await this.isValidSessionTopic(topic);
+    // if the session is already disconnected, we can't respond to the request so we need to delete it
+    await this.isValidSessionTopic(topic).catch((error) => {
```
bkrem: Style nit: would prefer if we stick with imperative `try/catch + await` like in other places in the engine rather than mixing in Promise `then/catch`.

I get this is more convenient syntax wise but feels a bit out of place.

packages/sign-client/test/sdk/client.spec.ts
```diff
@@ -346,6 +346,88 @@ describe("Sign Client Integration", () => {
...
+                    });
+                    _resolve();
+                  });
+                }
+                await clients.B.respond({
+                  topic,
+                  response: formatJsonRpcResult(id, "ok"),
+                }).catch((err) => {
+                  // eslint-disable-next-line no-console
+                  console.log("respond error", err);
```
bkrem, line 409: Still needed/wanted?

bkrem (approved):
💯 💯

packages/sign-client/src/controllers/engine.ts
ganchoradkov: we have validation tests where params are forced empty and was resulting in an exception `.id of undefined`. Will think to handle it differently

packages/sign-client/test/sdk/client.spec.ts
ganchoradkov: the log not, no but the catch is required to handle the rejection

PR #4105, opened 2024-01-05
fix: saves `session` object before sending `sessionSettle` request

ganchoradkov:
## Description
Flipped the execution order of `sessionSettle` & saving the `session` object to avoid potential race condition where `sessionSettle response` was received before the session object was saved resulting in an exception that the session is missing.

## Type of change

## How has this been tested?
dogfooding 

## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -291,6 +291,7 @@ export class Engine extends IEngine {
       this.client.logger.error(error);
       // if the publish fails, delete the session and throw an error
       this.client.session.delete(sessionTopic, getSdkError("USER_DISCONNECTED"));
+      await this.client.core.relayer.unsubscribe(sessionTopic);
       throw error;
     }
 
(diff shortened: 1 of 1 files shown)


packages/sign-client/src/controllers/engine.ts
```diff
@@ -286,6 +280,20 @@ export class Engine extends IEngine {
...
+      await this.sendRequest({
+        topic: sessionTopic,
+        method: "wc_sessionSettle",
+        params: sessionSettle,
+        throwOnFailedPublish: true,
+      });
+    } catch (error) {
+      this.client.logger.error(error);
+      // if the publish fails, delete the session and throw an error
+      this.client.session.delete(sessionTopic, getSdkError("USER_DISCONNECTED"));
```
bkrem, line 293: Do we also need to unsubscribe from `sessionTopic` in this case to undo https://github.com/WalletConnect/walletconnect-monorepo/pull/4105/files#diff-04e05e19b9270d1267b68d324e16997e4465d6879324fe4427a6194c00410eabR269 or is that undesirable?

bkrem (approved):

packages/sign-client/src/controllers/engine.ts
ganchoradkov: good catch! Adding..

PR #4106, opened 2024-01-05
fix: awaits sending session proposal before returning connection URI

ganchoradkov:
## Description
Added an await when sending session proposal so we can avoid cases where publishing is delayed or fails thus rendeing the connection URI unusable 
canary release - `2.11.0-canary-f92166d`

## Type of change

## How has this been tested?
dogfooding

## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -206,6 +206,7 @@ export class Engine extends IEngine {
       topic,
       method: "wc_sessionPropose",
       params: proposal,
+      throwOnFailedPublish: true,
     });
 
     const expiry = calcExpiry(FIVE_MINUTES);


bkrem (approved):

arein (commented):
not really clear why `throwOnFailedPublish` results in `awaits sending session proposal`?

It suggests this throws an error?

ganchoradkov:
> not really clear why `throwOnFailedPublish` results in `awaits sending session proposal`?
> 
> It suggests this throws an error?

Yes, `throwOnFailedPublish` rejects when the published payload doesn't receive an ACK from the relay meaning when it resolves, the relay definitely did receive the payload, and can be considered sent @arein

arein:
> > not really clear why `throwOnFailedPublish` results in `awaits sending session proposal`?
> 
> > 
> 
> > It suggests this throws an error?
> 
> 
> 
> Yes, `throwOnFailedPublish` rejects when the published payload doesn't receive an ACK from the relay meaning when it resolves, the relay definitely did receive the payload, and can be considered sent @arein 

Challenging this a bit more - shouldn't this be the default? When should we not wait for the ACK @ganchoradkov ?

ganchoradkov:
> > > not really clear why `throwOnFailedPublish` results in `awaits sending session proposal`?
> > 
> > 
> > > 
> > 
> > 
> > > It suggests this throws an error?
> > 
> > 
> > Yes, `throwOnFailedPublish` rejects when the published payload doesn't receive an ACK from the relay meaning when it resolves, the relay definitely did receive the payload, and can be considered sent @arein
> 
> Challenging this a bit more - shouldn't this be the default? When should we not wait for the ACK @ganchoradkov ?

internal requests/ack's between the SDKs are done async thus and failed publishes are handled by the `heartbeat`/auto retry. For requests requiring immediate user feedback for success & failure, the auto retry is disabled and the request is rejected to notify users something went wrong @arein

PR #4114, opened 2024-01-09
refactor: removed required & optional namespaces from `sessionSettle`

ganchoradkov:
## Description

...request to match the spec https://specs.walletconnect.com/2.0/specs/clients/sign/rpc-methods#wc_sessionsettle as its redundant to transmit these given they're readily available from the `session proposal` on the dapp side

## Type of change

## How has this been tested?
tests

## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -38,7 +38,6 @@ import {
   createDelayedPromise,
   engineEvent,
   getInternalError,
-  getRequiredNamespacesFromNamespaces,
   getSdkError,
   isConformingNamespaces,
   isExpired,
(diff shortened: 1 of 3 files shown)


packages/sign-client/src/controllers/engine.ts
```diff
@@ -826,25 +822,17 @@ export class Engine extends IEngine {
...
       const session = {
         topic,
         relay,
         expiry,
         namespaces,
         acknowledged: true,
         pairingTopic,
-        requiredNamespaces,
-        optionalNamespaces,
+        requiredNamespaces: {},
```
bkrem, line 834: Why set to empty object here? Guessing simply to satisfy types since it looks like these keys are overwritten with `proposal.[key]` anyways on the engineEvent handler above?

bkrem (approved):

packages/sign-client/src/controllers/engine.ts
ganchoradkov: yes, thought its redundant to create new `connect` event type just for this

PR #4159, opened 2024-01-22
fix: synced expiries

ganchoradkov:
## Description
Implemented synced expiries for session proposals and session requests. Synced expiries are defined by the sending peer and used by both sending & receiving peers. If no `expiryTimestamp` is received, default values are used.

Additionally
- patched a few flaky tests to always await async requests
- increased publish<->ack timeout to 20s
- implemented mechanism to wait for all requests in flight (that wait for a relay ack) to complete before closing transport 
- implemented cleanup on sent pending requests (history store) when the session is disconnected

## Type of change

## How has this been tested?
tests
dogfooding
https://github.com/WalletConnect/web-examples/pull/432

## Checklist

### DIFF
--- a/packages/core/src/controllers/pairing.ts
+++ b/packages/core/src/controllers/pairing.ts
@@ -34,7 +34,7 @@ import {
   isJsonRpcResult,
   isJsonRpcError,
 } from "@walletconnect/jsonrpc-utils";
-import { FIVE_MINUTES, THIRTY_DAYS } from "@walletconnect/time";
+import { FIVE_MINUTES, THIRTY_DAYS, toMiliseconds } from "@walletconnect/time";
 import EventEmitter from "events";
 import {
   PAIRING_CONTEXT,
@@ -99,6 +99,7 @@ export class Pairing implements IPairing {
       topic,
       symKey,
       relay,
+      expiryTimestamp: expiry,
     });
     await this.pairings.set(topic, pairing);
     await this.core.relayer.subscribe(topic);
@@ -110,7 +111,7 @@ export class Pairing implements IPairing {
   public pair: IPairing["pair"] = async (params) => {
     this.isInitialized();
     this.isValidPair(params);
-    const { topic, symKey, relay } = parseUri(params.uri);
+    const { topic, symKey, relay, expiryTimestamp } = parseUri(params.uri);
     let existingPairing;
     if (this.pairings.keys.includes(topic)) {
       existingPairing = this.pairings.get(topic);
@@ -121,7 +122,7 @@ export class Pairing implements IPairing {
       }
     }
 
-    const expiry = calcExpiry(FIVE_MINUTES);
+    const expiry = expiryTimestamp || calcExpiry(FIVE_MINUTES);
     const pairing = { topic, relay, expiry, active: false };
     await this.pairings.set(topic, pairing);
     this.core.expirer.set(topic, expiry);
@@ -396,6 +397,16 @@ export class Pairing implements IPairing {
     if (!uri?.symKey) {
       const { message } = getInternalError("MISSING_OR_INVALID", `pair() uri#symKey`);
       throw new Error(message);
+    }
+    if (uri?.expiryTimestamp) {
+      const expiration = toMiliseconds(uri?.expiryTimestamp);
+      if (expiration < Date.now()) {
+        const { message } = getInternalError(
+          "EXPIRED",
+          `pair() URI has expired. Please try again with a new connection URI.`,
+        );
+        throw new Error(message);
+      }
     }
   };
 
--- a/packages/core/src/controllers/publisher.ts
+++ b/packages/core/src/controllers/publisher.ts
@@ -50,7 +50,7 @@ export class Publisher extends IPublisher {
         const publish = await createExpiringPromise(
           this.rpcPublish(topic, message, ttl, relay, prompt, tag, id),
           this.publishTimeout,
-          "Failed to publish payload, please try again.",
+          `Failed to publish payload, please try again. id:${id} tag:${tag}`,
         );
         await publish;
         this.removeRequestFromQueue(id);
--- a/packages/core/src/controllers/publisher.ts
+++ b/packages/core/src/controllers/publisher.ts
@@ -20,7 +20,7 @@ export class Publisher extends IPublisher {
   public name = PUBLISHER_CONTEXT;
   public queue = new Map<string, PublisherTypes.Params>();
 
-  private publishTimeout = toMiliseconds(TEN_SECONDS + 5);
+  private publishTimeout = toMiliseconds(TEN_SECONDS + TEN_SECONDS);
   private needsTransportRestart = false;
 
   constructor(public relayer: IRelayer, public logger: Logger) {
--- a/packages/core/src/controllers/publisher.ts
+++ b/packages/core/src/controllers/publisher.ts
@@ -20,7 +20,7 @@ export class Publisher extends IPublisher {
   public name = PUBLISHER_CONTEXT;
   public queue = new Map<string, PublisherTypes.Params>();
 
-  private publishTimeout = toMiliseconds(TEN_SECONDS + TEN_SECONDS);
+  private publishTimeout = toMiliseconds(TEN_SECONDS * 2);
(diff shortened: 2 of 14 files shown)


bkrem:
> If no expiryTimestamp is received, default values are used.

This is for backwards compat I'm assuming?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -156,6 +157,8 @@ export class Engine extends IEngine {
 
     const publicKey = await this.client.core.crypto.generateKeyPair();
 
+    const expiry = ENGINE_RPC_OPTS.wc_sessionPropose.req.ttl || FIVE_MINUTES;
```
bkrem, line 160: `ENGINE_RPC_OPTS.wc_sessionPropose.req.ttl` is already a constant right, why do we need the `||` defaulting?

packages/core/src/controllers/pairing.ts
```diff
@@ -397,6 +398,16 @@ export class Pairing implements IPairing {
       const { message } = getInternalError("MISSING_OR_INVALID", `pair() uri#symKey`);
       throw new Error(message);
     }
+    if (uri?.expiryTimestamp) {
+      const expiration = toMiliseconds(uri?.expiryTimestamp);
+      if (expiration < Date.now()) {
```
bkrem, line 403: Probably strictly theoretical rather than realistic, but I think there's an edge case here where:
* `toMiliseconds` simply adds zeroes to the end of seconds-based `expiration`
* `Date.now()` has non-zero values in the last 3 digits

Either way I think we can ignore because that only applies within the same second :)

packages/core/src/controllers/publisher.ts
```diff
@@ -20,7 +20,7 @@ export class Publisher extends IPublisher {
   public name = PUBLISHER_CONTEXT;
   public queue = new Map<string, PublisherTypes.Params>();
 
-  private publishTimeout = toMiliseconds(TEN_SECONDS);
+  private publishTimeout = toMiliseconds(TEN_SECONDS + TEN_SECONDS);
```
bkrem:
```suggestion
  private publishTimeout = toMiliseconds(TEN_SECONDS * 2);
```

ganchoradkov:
> This is for backwards compat I'm assuming?

correct

packages/sign-client/src/controllers/engine.ts
```diff
@@ -349,7 +352,7 @@ export class Engine extends IEngine {
   public request: IEngine["request"] = async <T>(params: EngineTypes.RequestParams) => {
     await this.isInitialized();
     await this.isValidRequest(params);
-    const { chainId, request, topic, expiry } = params;
+    const { chainId, request, topic, expiry = FIVE_MINUTES } = params;
```
bkrem: Should be using mapped constant rather than a plain duration constant?

```suggestion
    const { chainId, request, topic, expiry = ENGINE_RPC_OPTS.wc_sessionRequest.req.ttl } = params;
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -541,21 +550,21 @@ export class Engine extends IEngine {
 
   private setProposal: EnginePrivate["setProposal"] = async (id, proposal) => {
     await this.client.proposal.set(id, proposal);
-    this.client.core.expirer.set(id, proposal.expiry);
+    this.client.core.expirer.set(id, calcExpiry(FIVE_MINUTES));
```
```suggestion
    this.client.core.expirer.set(id, calcExpiry(ENGINE_RPC_OPTS.wc_sessionPropose.req.ttl));
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -541,21 +550,21 @@ export class Engine extends IEngine {
...
   };
 
   private setPendingSessionRequest: EnginePrivate["setPendingSessionRequest"] = async (
     pendingRequest: PendingRequestTypes.Struct,
   ) => {
-    const expiry = ENGINE_RPC_OPTS.wc_sessionRequest.req.ttl;
     const { id, topic, params, verifyContext } = pendingRequest;
+    const expiry = params.request.expiryTimestamp || calcExpiry(FIVE_MINUTES);
```
```suggestion
    const expiry = params.request.expiryTimestamp || calcExpiry(ENGINE_RPC_OPTS.wc_sessionRequest.req.ttl);
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -768,8 +777,8 @@ export class Engine extends IEngine {
     const { params, id } = payload;
     try {
       this.isValidConnect({ ...payload.params });
-      const expiry = calcExpiry(FIVE_MINUTES);
-      const proposal = { id, pairingTopic: topic, expiry, ...params };
+      const expiryTimestamp = params.expiryTimestamp || calcExpiry(FIVE_MINUTES);
```
```suggestion
      const expiryTimestamp = params.expiryTimestamp || calcExpiry(ENGINE_RPC_OPTS.wc_sessionPropose.req.ttl);
```

packages/types/src/sign-client/proposal.ts
```diff
@@ -17,7 +17,8 @@ export declare namespace ProposalTypes {
 
   export interface Struct {
     id: number;
-    expiry: number;
+    expiry?: number; // deprecated in favour of expiryTimespamp
```
bkrem:
```suggestion
    expiry?: number; // deprecated in favour of expiryTimestamp
```

bkrem: Might be more useful to mark this as deprecated in a way TS/editors will pick up:
https://stackoverflow.com/questions/60755711/is-it-possible-to-mark-something-as-deprecated-in-typescript

packages/core/src/controllers/relayer.ts
```diff
@@ -204,6 +220,13 @@ export class Relayer extends IRelayer {
   }
 
   public async transportClose() {
+    // wait for all requests to finish before closing the transport
+    if (this.requestsInFlight.size > 0) {
+      this.requestsInFlight.forEach(async (value) => {
```
bkrem, line 226:
```suggestion
      this.logger.debug("Waiting for all in-flight requests to finish before closing transport...")
      this.requestsInFlight.forEach(async (value) => {
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1058,6 +1073,30 @@ export class Engine extends IEngine {
...
+  }: {
+    topic: string;
+    error: ErrorResponse;
+  }) => {
+    const pendingRequests = this.client.core.history.pending;
+    if (pendingRequests.length > 0) {
+      const forSession = pendingRequests.filter(
+        (r) => r.topic === topic && r.request.method === "wc_sessionRequest",
+      );
+      if (forSession.length > 0) {
```
bkrem: Not really needed, if `filter` returned an empty array the `.forEach` is effectively a noop anyways.

bkrem (approved):
Nice 💯 some minor tweaks suggested but LGTM overall

packages/sign-client/src/controllers/engine.ts
ganchoradkov: it's type is actually `number | undefined` causing ts to complain

ganchoradkov: might be better just updating it to explicit type `number` and not optional

kieranbop1990:
@ganchoradkov it's not entirely clear how to apply this fix to override those five minutes. Any documentation on it? Thanks

UTkzhang:
yea, how would this translate to folks using wallet connect through providers like rainbowkit/wagmi?

microHoffman:
Personally I did not find a way how to pass `expiryTimestamp` param from wagmi, so I just overwritten in my `main.ts` the default values from https://github.com/WalletConnect/walletconnect-monorepo/blob/6d7164e8c4de52803ad6cfa7fb5510e559c6c64d/packages/sign-client/src/constants/engine.ts#L65 like this:

```ts
import { ENGINE_RPC_OPTS } from '@walletconnect/sign-client'

ENGINE_RPC_OPTS.wc_sessionRequest.req.ttl = 604800 // 7 days
ENGINE_RPC_OPTS.wc_sessionRequest.res.ttl = 604800 // 7 days
```

PR #4194, opened 2024-02-02
feat: granular error messages

ganchoradkov:
## Description
- added more granular error messages for easier debugging
- updated expirer missing item to warning
- added `recentlyDeleted` items to `Store` to return different error message when item is attempted to be accessed after delete
- rework publish & connection flow to be more robust and handle disconnects mid publish

## Type of change

## How has this been tested?
tests
canary - `2.11.1-canary-a72001fd1`
wallet -> https://react-wallet-v2-git-chore-test-2111-canar-f1e11a-walletconnect1.vercel.app/walletconnect
dapp -> https://react-dapp-v2-git-chore-test-2111-canary-2fec2c-walletconnect1.vercel.app/
## Checklist


packages/core/src/controllers/store.ts
```diff
@@ -23,6 +23,9 @@ export class Store<Key, Data extends Record<string, any>> extends IStore<Key, Da
 
   private storagePrefix = CORE_STORAGE_PREFIX;
 
+  // stores recently deleted key to return different rejection message when key is not found
+  private recentlyDeleted: unknown[] = [];
```
bkrem: From what I'm understanding the lifecycle here is that this is in-memory only and cleanup is implicitly happening via reset on client restart?

What about edge cases of this accumulating e.g. 100s/1000s of entries in long running clients (e.g. Fireblocks)?

packages/core/src/controllers/store.ts
```diff
@@ -147,6 +151,15 @@ export class Store<Key, Data extends Record<string, any>> extends IStore<Key, Da
   private getData(key: Key) {
     const value = this.map.get(key);
     if (!value) {
+      if (this.recentlyDeleted.includes(key)) {
```
bkrem, line 163: Again curious here about the upper bound we can get to in edge cases with this.

Arguably would only really make a difference from like 1 million entries upwards, but might be wiser to use `Set` for this? https://dev.to/danielbayerlein/array-prototype-includes-can-slow-down-your-code-43c7

Also avoids pushing duplicate keys into the array for any reason, thus bloating it.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -108,6 +108,12 @@ export class Engine extends IEngine {
 
   private requestQueueDelay = ONE_SECOND;
 
+  // temporary map to store recently deleted items
```
```suggestion
  // Ephemeral (in-memory) map to store recently deleted items
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -155,6 +166,11 @@ export class Engine extends IEngine {
       uri = newUri;
     }
 
+    if (!topic) {
+      const { message } = getInternalError("NO_MATCHING_KEY", `connect() pairing topic: ${topic}`);
```
bkrem, line 171: What is the actual root cause of this error state/can we state it in the error message?

From my interpretation: seems like this happens if there is no valid `pairingTopic` param provided AND generating a topic via `pairing.create()` failed?

bkrem: I see this simply got moved up, but still keen to make this clearer if we're already touching the code.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -576,7 +639,14 @@ export class Engine extends IEngine {
       const hash = hashMessage(JSON.stringify(payload));
       this.client.core.verify.register({ attestationId: hash });
     }
-    const message = await this.client.core.crypto.encode(topic, payload);
+    let message;
+    try {
+      message = await this.client.core.crypto.encode(topic, payload);
+    } catch (error) {
+      await this.cleanup();
+      this.client.logger.error(`sendRequest() -> encode(${topic}) failed`);
```
bkrem: This feels misleading because it seems to show a single function parameter when there are two. Think there's less potential for confusion with e.g.:
```suggestion
      this.client.logger.error(`sendRequest() -> core.crypto.encode() for topic ${topic} failed`);
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -576,7 +639,14 @@ export class Engine extends IEngine {
       const hash = hashMessage(JSON.stringify(payload));
       this.client.core.verify.register({ attestationId: hash });
     }
-    const message = await this.client.core.crypto.encode(topic, payload);
+    let message;
+    try {
+      message = await this.client.core.crypto.encode(topic, payload);
+    } catch (error) {
+      await this.cleanup();
```
bkrem, line 695: Context on the added `this.cleanup` call?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -598,8 +668,21 @@ export class Engine extends IEngine {
...
     const { id, topic, result, throwOnFailedPublish } = args;
     const payload = formatJsonRpcResult(id, result);
-    const message = await this.client.core.crypto.encode(topic, payload);
-    const record = await this.client.core.history.get(topic, id);
+    let message;
+    try {
+      message = await this.client.core.crypto.encode(topic, payload);
+    } catch (error) {
+      await this.cleanup();
+      this.client.logger.error(`sendResult() -> encode(${topic}) failed`);
```
```suggestion
      this.client.logger.error(`sendResult() -> core.crypto.encode() for topic ${topic} failed`);
```

packages/sign-client/src/controllers/engine.ts
bkrem: Also, we can pass objects to `logger`, so might be more useful to move to an approach like this to get the maximum amount of context:

```
logger.error({message: "sendRequest() -> core.crypto.encode...", topic, payload})
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1203,6 +1301,9 @@ export class Engine extends IEngine {
       );
       throw new Error(message);
     }
+
+    this.checkRecentlyDeleted(topic);
+    // Store will throw custom message if topic was recently deleted
```
bkrem, line 1421: Should this comment be above `this.checkRecentlyDeleted`?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1343,6 +1447,7 @@ export class Engine extends IEngine {
       throw new Error(message);
     }
     const { id, reason } = params;
+    this.checkRecentlyDeleted(id);
```
bkrem, line 1566: I don't think this is covered in tests atm, only string-based topic `set/get` for the `recentlyDeleted` map. 

Are we sure this will behave as assumed here when using the proposal id (i.e. `number`) as a key?

bkrem:
Nice clarifications and additionals guards here 💪  have some concerns around error formatting, upper bounds on `recentlyDeleted`, and seems like proposal + recentlyDeleted behaviour is a blind spot right now

packages/core/src/controllers/pairing.ts
```diff
@@ -136,9 +136,8 @@ export class Pairing implements IPairing {
     // avoid overwriting keychain pairing already exists
     if (!this.core.crypto.keychain.has(topic)) {
       await this.core.crypto.setSymKey(symKey, topic);
-      await this.core.relayer.subscribe(topic, { relay });
     }
-
+    await this.core.relayer.subscribe(topic, { relay });
```
devceline, line 140: What's the motivation/implication of this?

packages/core/src/controllers/publisher.ts
```diff
@@ -37,42 +38,57 @@ export class Publisher extends IPublisher {
...
-      try {
+      /**
+       * Loop until the publish is successful or the timeout is reached
+       * The loop allows to retry to retry the publish in case of disconnect
+       */
+      while (result === undefined) {
+        if (Date.now() - startPublish > this.publishTimeout) {
+          throw new Error(failedPublishMessage);
+        }
+        this.logger.trace({ id, attempts }, "publisher.publish - attempt");
```
devceline:
```suggestion
        this.logger.trace({ id, attempts }, `publisher.publish - attempt ${attempts}`);
```

packages/core/src/controllers/relayer.ts
```diff
@@ -84,6 +88,9 @@ export class Relayer extends IRelayer {
     }
   >();
 
+  private pingTimeout: NodeJS.Timeout | undefined;
+  private heartBeatTimeout = toMiliseconds(THIRTY_SECONDS + ONE_SECOND);
```
devceline, line 96: Where do these numbers come from? Can we add a comment here

packages/core/src/controllers/relayer.ts
```diff
@@ -131,11 +139,13 @@ export class Relayer extends IRelayer {
   }
 
   get connected() {
-    return this.provider.connection.connected;
+    // @ts-ignore
+    return this.provider?.connection?.socket?.readyState === 1;
```
devceline: can we use `expect-error` and explain why?

packages/core/src/controllers/relayer.ts
```diff
@@ -131,11 +139,13 @@ export class Relayer extends IRelayer {
...
   }
 
   get connecting() {
-    return this.provider.connection.connecting;
+    // @ts-ignore
+    return this.provider?.connection?.socket?.readyState === 0;
```
devceline: ditto here

packages/core/src/controllers/relayer.ts
```diff
@@ -152,9 +162,6 @@ export class Relayer extends IRelayer {
   public async subscribe(topic: string, opts?: RelayerTypes.SubscribeOptions) {
     this.isInitialized();
     let id = this.subscriber.topicMap.get(topic)?.[0] || "";
-
-    if (id) return id;
-
```
devceline, line 157: Why are we removing this logic?

packages/core/src/controllers/relayer.ts
```diff
@@ -429,43 +492,31 @@ export class Relayer extends IRelayer {
...
-    this.events.on(RELAYER_EVENTS.connection_stalled, () => {
-      this.restartTransport().catch((error) => this.logger.error(error));
-    });
-
     let lastConnectedState = await isOnline();
     subscribeToNetworkChange(async (connected: boolean) => {
-      if (!this.initialized) return;
-
-      // sometimes the network change event is triggered multiple times so avoid reacting to the same value
+      // sometimes the network change event is triggered multiple times so avoid reacting to the samFe value
```
devceline, line 501:
```suggestion
      // sometimes the network change event is triggered multiple times so avoid reacting to the same value
```

packages/core/src/controllers/subscriber.ts
```diff
@@ -217,15 +225,16 @@ export class Subscriber extends ISubscriber {
...
       );
-      await subscribe;
+      const result = await subscribe;
+      return result ? hashMessage(topic + this.clientId) : "";
     } catch (err) {
       this.logger.debug(`Outgoing Relay Subscribe Payload stalled`);
       this.relayer.events.emit(RELAYER_EVENTS.connection_stalled);
     }
-    return hashMessage(topic + this.clientId);
+    return "";
```
devceline: Why return empty string here?

packages/core/src/controllers/subscriber.ts
```diff
@@ -382,13 +390,14 @@ export class Subscriber extends ISubscriber {
...
+
     const result = (await this.rpcBatchSubscribe(subscriptions)) as string[];
     if (!isValidArray(result)) return;
+
     this.onBatchSubscribe(result.map((id, i) => ({ ...subscriptions[i], id })));
   }
 
   private async onConnect() {
-    if (this.restartInProgress) return;
     await this.restart();
```
devceline, line 403: What happens if there is there is already one in progress now?

packages/core/src/controllers/pairing.ts
ganchoradkov: Subscribing on every pairing attempt allows for any messages that are not yet received to be delivered. Extra safeguard

packages/core/src/controllers/relayer.ts
ganchoradkov: Discussion with relay team. The relay pings 30s after no messages were transmimitted so 31s is safe to assume connection is dead

packages/core/src/controllers/relayer.ts
ganchoradkov: same reason as above, if `subscribe` is attempted, we should subscribe and not rely on past subscriptions

packages/core/src/controllers/subscriber.ts
ganchoradkov: to differenciate between success and failed subscriptions. Empty result = failed subscribe

packages/core/src/controllers/subscriber.ts
ganchoradkov: It will restart the transport. Logic is, and confirmed that it happens in the wild
- open socket
- attempt to subscribe
- disconnect received while subscribe is still in progress
- retry all again

packages/core/src/controllers/pairing.ts
bkrem: I think it is on the `relayer` controller, but want to confirm: if the topic is already subscribed is this effectively a no-op?

packages/core/src/controllers/store.ts
```diff
@@ -23,6 +23,10 @@ export class Store<Key, Data extends Record<string, any>> extends IStore<Key, Da
 
   private storagePrefix = CORE_STORAGE_PREFIX;
 
+  // stores recently deleted key to return different rejection message when key is not found
+  private recentlyDeleted: unknown[] = [];
```
bkrem: Seems like the type is known here?

```suggestion
  private recentlyDeleted: Key[] = [];
```

packages/core/src/controllers/store.ts
```diff
@@ -130,11 +134,20 @@ export class Store<Key, Data extends Record<string, any>> extends IStore<Key, Da
...
     this.map.delete(key);
+    this.addToRecentlyDeleted(key);
     await this.persist();
   };
 
   // ---------- Private ----------------------------------------------- //
 
+  private addToRecentlyDeleted(key: Key) {
+    this.recentlyDeleted.push(key);
+    // limit the size of the recentlyDeleted array
```
bkrem:
```suggestion
    // limit the size of the recentlyDeleted array, truncate the 100 oldest entries.
```

packages/core/src/controllers/publisher.ts
```diff
@@ -37,42 +38,57 @@ export class Publisher extends IPublisher {
...
-        this.needsTransportRestart = true;
-        if (opts?.internal?.throwOnFailedPublish) {
-          // remove the request from the queue so it's not retried automatically
-          this.removeRequestFromQueue(id);
-          throw err;
+        result = await publish;
+        attempts++;
+
+        if (!result) {
+          // small buffer to allow transport to beging reconnecting
```
bkrem: Explains the "what?" but not the "why?", i.e. why is this needed?

```suggestion
          // small buffer to allow transport to begin reconnecting
```

bkrem: Also: is it sufficient under all circumstances? Can this create more edge cases if this assumption fails?

packages/core/src/controllers/publisher.ts
```diff
@@ -37,42 +38,57 @@ export class Publisher extends IPublisher {
...
-      const params = { topic, message, opts: { ttl, relay, prompt, tag, id } };
-      // delay adding to queue to avoid cases where heartbeat might pulse right after publish resulting in duplicate publish
-      const queueTimeout = setTimeout(() => this.queue.set(id, params), this.publishTimeout);
-      try {
+      /**
+       * Loop until the publish is successful or the timeout is reached
+       * The loop allows to retry to retry the publish in case of disconnect
+       */
+      while (result === undefined) {
+        if (Date.now() - startPublish > this.publishTimeout) {
```
bkrem, line 60:
```suggestion
	// Terminate the publishing attempts if publisTimeout has been exceeded
        if (Date.now() - startPublish > this.publishTimeout) {
```

packages/core/src/controllers/subscriber.ts
bkrem: Can we add that as a comment for context pls 🙏 non-intuitive that empty string is being used as a flag.

Couldn't we return `null` here which would represent an empty/missing result better?

packages/core/src/controllers/relayer.ts
bkrem: Ack on context, can we add as comment pls as suggested 🙏 They happen to be constants here but it's still effectively a magic number like this :)

packages/core/src/controllers/relayer.ts
```diff
@@ -118,7 +125,8 @@ export class Relayer extends IRelayer {
     }
     this.initialized = true;
     setTimeout(async () => {
-      if (this.subscriber.topics.length === 0) {
+      // @ts-ignore
```
```suggestion
      // @ts-expect-error - reason...
```

bkrem: Why does this `if` need a type override?

packages/core/src/controllers/relayer.ts
bkrem: If the TS issue is over mixed return types, can we use logic that will always return a boolean? E.g. `?? false`

packages/core/src/controllers/relayer.ts
```diff
@@ -219,80 +252,74 @@ export class Relayer extends IRelayer {
...
-        await value.promise;
-      });
+  public async transportDisconnect() {
+    if (!this.hasExperiencedNetworkDisruption && this.connected && this.requestsInFlight.size > 0) {
+      try {
+        await Promise.all(
+          Array.from(this.requestsInFlight.values()).map((request) => request.promise),
+        );
+      } catch (e) {
+        this.logger.warn(e);
```
bkrem, line 266: Do we need an explicit error message here or do the promises return the wanted context if they throw?

bkrem (commented):
First batch of comments, will come back to this :)

packages/core/src/controllers/relayer.ts
ganchoradkov: we don't expect errors but its just in case. Also ts complains on empty try/catch

packages/core/src/controllers/publisher.ts
ganchoradkov: I've updated the comment, but also its role is to avoid spam call `rpcPublish` if network is down for a few seconds but limit publish to max 1 try per second

packages/core/src/controllers/relayer.ts
ganchoradkov: subscriber.pending wasn't included in the types initially even though was public property

packages/core/src/controllers/publisher.ts
```diff
@@ -71,7 +73,8 @@ export class Publisher extends IPublisher {
         attempts++;
 
         if (!result) {
-          // small buffer to allow transport to beging reconnecting
+          // small delay before retrying so we can limit retry to max 1 time per second
+          // if the network is `rpcPublish` will throw immediately
```
bkrem: if the network is... :)

packages/core/src/controllers/relayer.ts
```diff
@@ -219,80 +256,74 @@ export class Relayer extends IRelayer {
...
+      await new Promise<void>(async (resolve, reject) => {
+        const onDisconnect = () => {
+          this.provider.off(RELAYER_PROVIDER_EVENTS.disconnect, onDisconnect);
+          reject(new Error(`Connection interrupted while trying to subscribe`));
+        };
+        this.provider.on(RELAYER_PROVIDER_EVENTS.disconnect, onDisconnect);
+
+        await createExpiringPromise(
+          this.provider.connect(),
+          toMiliseconds(ONE_MINUTE),
```
bkrem, line 303: Sanity check: this seems to have gone from `10_000` milliseconds (10 seconds) to one minute. Was that intentional?

packages/core/src/controllers/relayer.ts
```diff
@@ -304,6 +335,41 @@ export class Relayer extends IRelayer {
...
+   * In Node, we must detect when the connection is stalled and terminate it.
+   * The logic is, if we don't receive ping from the relay within a certain time, we terminate the connection.
+   * The timer is refreshed on every message received from the relay.
+   *
+   * In the browser, ping/pong events are not exposed, so the above behaviour is handled by `subscribeToNetworkChange` and `isOnline` functions.
+   */
+  private startPingTimeout() {
+    if (!isNode()) return;
+    try {
+      //@ts-ignore
```
bkrem: `expect-error` generally the better one to use if really needed, since TS will flag if there is no longer a type error here.

(guessing this is why, please adjust as needed)
```suggestion
      //@ts-expect-error - Types are divergent between the node and browser WS API
```

packages/core/src/controllers/relayer.ts
```diff
@@ -304,6 +335,41 @@ export class Relayer extends IRelayer {
...
+          this.heartbeat();
+        });
+      }
+      this.heartbeat();
+    } catch (e) {
+      this.logger.warn(e);
+    }
+  }
+
+  private heartbeat = () => {
```
bkrem: Feels like we're overloading the term `heartbeat` bc we already have other use cases with this name in our core, would rather it just says what it does:

```suggestion
  private resetPingTimeout = () => {
```

packages/core/test/store.spec.ts
```diff
@@ -148,6 +148,43 @@ describe("Store", () => {
...
+      try {
+        await store.get(key);
+      } catch (e) {
+        expect(e.message).to.equal(
+          `Missing or invalid. Record was recently deleted - mock-entity: ${key}`,
+        );
+      }
+    });
+    it("should cleanup recentlyDeleted when size limit is reached", async () => {
+      //@ts-ignore
```
bkrem: Why so many `@ts-ignore` needed here?

packages/sign-client/package.json
```diff
@@ -27,7 +27,7 @@
     "test:run": "vitest run --dir test/sdk",
     "test:concurrency": "vitest run --dir test/concurrency",
     "test:xregion": "vitest run --dir test/xregion -- --dangerouslyIgnoreUnhandledErrors --segfault-retry=3",
-    "test": "npm run test:pre; npm run test:run",
+    "test": "npm run test:pre; npm run test:run  --threads=false",
```
bkrem, line 30: Still wanted or experimental?

packages/sign-client/src/constants/engine.ts
```diff
@@ -91,12 +91,12 @@ export const ENGINE_RPC_OPTS: EngineTypes.RpcOptsMap = {
   },
   wc_sessionPing: {
     req: {
-      ttl: THIRTY_SECONDS,
+      ttl: ONE_DAY,
```
bkrem, line 94: Seems to be going off-spec? https://specs.walletconnect.com/2.0/specs/clients/sign/rpc-methods#wc_sessionping

packages/sign-client/src/controllers/engine.ts
```diff
@@ -108,6 +109,14 @@ export class Engine extends IEngine {
 
   private requestQueueDelay = ONE_SECOND;
 
+  // Ephemeral (in-memory) map to store recently deleted items
+  private recentlyDeletedMap = new Map<
+    string | number,
+    "pairing" | "session" | "proposal" | "request"
+  >();
+
+  private recentlyDeletedLimit = 200;
```
bkrem, line 118: Defining this limit in 2 separate places, can we use a shared constant pls 🙏

packages/sign-client/src/controllers/engine.ts
```diff
@@ -832,6 +978,11 @@ export class Engine extends IEngine {
       await this.client.core.pairing.activate({ topic });
     } else if (isJsonRpcError(payload)) {
       await this.client.proposal.delete(id, getSdkError("USER_DISCONNECTED"));
+      const target = engineEvent("session_connect");
+      const listeners = this.events.listenerCount(target);
+      if (listeners === 0) {
+        throw new Error(`emitting ${target} without any listeners, 954`);
```
bkrem, line 984: What are these numbers?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -865,7 +1016,17 @@ export class Engine extends IEngine {
...
+      await this.sendResult<"wc_sessionSettle">({
+        id: payload.id,
+        topic,
+        result: true,
+        throwOnFailedPublish: true,
+      });
+      const target = engineEvent("session_connect");
+      const listeners = this.events.listenerCount(target);
+      if (listeners === 0) {
+        throw new Error(`emitting ${target} without any listeners 997`);
```
bkrem, line 1028: Here too?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1549,4 +1775,36 @@ export class Engine extends IEngine {
...
+          break;
+        }
+        this.recentlyDeletedMap.delete(k);
+      }
+    }
+  };
+
+  private checkRecentlyDeleted = (id: string | number) => {
+    const deletedRecord = this.recentlyDeletedMap.get(id);
+    if (this.recentlyDeletedMap.size > 200) {
```
bkrem: Using magic number instead of already defined const

packages/sign-client/test/sdk/integration/events.spec.ts
```diff
@@ -12,7 +12,8 @@ import {
 } from "../../shared";
 import { EngineTypes, PairingTypes, SessionTypes } from "@walletconnect/types";
 
-describe("Sign Client Events Validation", () => {
+// skip tests as they are already testeed in integration.spec.ts
```
bkrem:
```suggestion
// skip tests as they are already tested in integration.spec.ts
```

packages/sign-client/test/shared/connect.ts
```diff
@@ -162,8 +162,8 @@ export async function testConnectMethod(clients: Clients, params?: TestConnectPa
   expect(sessionA.namespaces).to.eql(approveParams.namespaces);
   expect(sessionA.namespaces).to.eql(sessionB.namespaces);
   expect(sessionA.sessionProperties).to.eql(TEST_SESSION_PROPERTIES_APPROVE);
-  // expiry
-  expect(Math.abs(sessionA.expiry - sessionB.expiry)).to.be.lessThan(5);
+  // testing expiry is not reliable as on slow networks take longer to settle
```
bkrem:
```suggestion
  // testing expiry is not reliable as on slow networks we take longer to settle
```

packages/web3wallet/test/sign.spec.ts
```diff
@@ -384,7 +384,7 @@ describe("Sign Integration", () => {
     ]);
   });
 
-  it("receive proposal_expire event", async () => {
+  it.skip("receive proposal_expire event", async () => {
```
bkrem, line 387: Also being skipped for reliability reasons as noted above or what is the reasoning?

providers/universal-provider/test/shared/connect.ts
```diff
@@ -141,8 +141,8 @@ export async function testConnectMethod(
   // namespaces
   expect(sessionA.namespaces).to.eql(approveParams.namespaces);
   expect(sessionA.namespaces).to.eql(sessionB.namespaces);
-  // expiry
-  expect(Math.abs(sessionA.expiry - sessionB.expiry)).to.be.lessThan(5);
+  // testing expiry is not reliable as on slow networks take longer to settle
```
bkrem:
```suggestion
  // testing expiry is not reliable as on slow networks we take longer to settle
```

bkrem (approved):
Mammoth effort here, powerful Gancho at it again 💪 

LGTM code-wise but still need to do some dogfooding/testing

packages/core/test/store.spec.ts
ganchoradkov: Updated to `//@ts-expect-error`. Reason is that its not public property but we're still acceesing it for tests

packages/web3wallet/test/sign.spec.ts
ganchoradkov: Yes, I will rework them as faking timers seems to impact other tests so may be there is an issue with test isolation

packages/sign-client/src/controllers/engine.ts
ganchoradkov: actually the whole check is redundant, as its performed during every add so I removed it

packages/core/src/controllers/relayer.ts
ganchoradkov: Yes, we want the SDK to function in all network conditions. For normal/fast networks everything stays the same while users on very slower networks will actually be able to use us

packages/core/src/controllers/relayer.ts
```diff
@@ -305,6 +335,41 @@ export class Relayer extends IRelayer {
...
+   *
+   * In the browser, ping/pong events are not exposed, so the above behaviour is handled by `subscribeToNetworkChange` and `isOnline` functions.
+   */
+  private startPingTimeout() {
+    if (!isNode()) return;
+    try {
+      //@ts-expect-error - Types are divergent between the node and browser WS API
+      if (this.provider?.connection?.socket) {
+        //@ts-expect-error
+        this.provider?.connection?.socket?.once("ping", () => {
```
devceline, line 351:
```suggestion
        this.provider.connection.socket.once("ping", () => {
```
We don't need the question marks if we're null checking above

packages/sign-client/test/shared/init.ts
```diff
@@ -56,16 +56,14 @@ export async function initTwoPairedClients(
         initTwoClients(clientOptsA, clientOptsB, sharedClientOpts),
         TESTS_CONNECT_TIMEOUT,
       )) as Clients;
-
       const settled: any = await createExpiringPromise(
         testConnectMethod(clients),
-        TESTS_CONNECT_TIMEOUT,
+        TESTS_CONNECT_TIMEOUT * 2,
```
devceline, line 61: Why `* 2`? Sorry if I missed something here, is something special with the tests?

devceline (approved):
LGTM

packages/sign-client/test/shared/init.ts
ganchoradkov: allows for several additional retries in case of flaky ci network

PR #4331, opened 2024-03-14
feat: auth aka sign 2.5

ganchoradkov:
## Description
Full Sign 2.5 implementation according to specs https://github.com/WalletConnect/walletconnect-specs/pull/171/

- added new `authenticate` method where dapps can request One Client Authentication as per https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-222.md
- added the corresponding wallet side approve/reject methods & handling
- tons of utils for recap handling & validations

## Type of change

## How has this been tested?
tests
## Checklist


packages/types/src/sign-client/auth.ts
```diff
@@ -0,0 +1,174 @@
...
+    nonce: string;
+    version?: string;
+    iat?: string;
+    nbf?: string;
+    exp?: string;
+    chainId?: string;
+    statement?: string;
+    requestId?: string;
+    resources?: string[];
+    expiry?: number;
```
Cali93, line 87: Not sure if a decision was taken on this but the [specs discussion](https://github.com/WalletConnect/walletconnect-specs/pull/171/files#r1432667513) suggests to rename `expiry` to `ttl`
```suggestion
    ttl?: number;
```

packages/utils/src/misc.ts
```diff
@@ -410,3 +410,7 @@ export async function getDeepLink(store: IKeyValueStorage, key: string) {
     console.error(err);
   }
 }
+
+export function getCommonValuesInArrays<T>(arr1: T[], arr2: T[]): unknown[] {
+  return arr1.filter((value) => arr2.includes(value));
+}
```
Cali93: Just want to flag that this will only work for primitive types.

packages/utils/src/namespaces.ts
```diff
@@ -228,3 +228,19 @@ export function normalizeNamespaces(
   }
   return normalizedNamespaces;
 }
+
+export function buildNamespacesFromAuth(methods: string[], accounts: string[]) {
+  accounts = accounts.map((account) => account.replace("did:pkh:", ""));
+
+  const namespace = accounts[0].split(":")[0];
```
Cali93: Aren't we assuming that all accounts are on the same namespace ?
What about wallets like Phantom that could have accounts on both `Solana` and `EVM` chains
Also could we use utils here ?
```suggestion
  const namespaces = getNamespacesFromAccounts(accounts);
```

packages/utils/src/namespaces.ts
```diff
@@ -228,3 +228,19 @@ export function normalizeNamespaces(
...
+
+  const namespaceObj = {
+    [namespace]: {
+      accounts,
+      chains: accounts.map((account) => `${account.split(":")[0]}:${account.split(":")[1]}`),
```
Cali93: Can use a util for formatting here ?

packages/utils/src/signatures.ts
```diff
@@ -0,0 +1,92 @@
...
+    );
+    const { result } = await response.json();
+    if (!result) return false;
+
+    // Remove right-padded zeros from result to get only the concrete recovered value.
+    const recoveredValue = result.slice(0, eip1271MagicValue.length);
+    return recoveredValue.toLowerCase() === eip1271MagicValue.toLowerCase();
+  } catch (error: any) {
+    // eslint-disable-next-line no-console
+    console.error("isValidEip1271Signature: ", error);
```
Cali93, line 85: Should we pass the logger as a param and use the client's logger ?

packages/utils/src/uri.ts
```diff
@@ -28,12 +28,14 @@ export function parseUri(str: string): EngineTypes.UriParameters {
   const requiredValues = path.split("@");
   const queryString: string = typeof pathEnd !== "undefined" ? str.substring(pathEnd) : "";
   const queryParams = qs.parse(queryString);
+  const methods = queryParams.methods ? (queryParams.methods as string).split(",") : undefined;
```
Cali93: Consider using a constant for both splitting and joining.
You can also avoid casting the methods to a string if you change your initial condition like below.

```suggestion
  const METHODS_SEPARATOR = ","
  const methods = typeof queryParams.methods === 'string' ? queryParams.methods.split(METHODS_SEPARATOR) : undefined;
```

providers/ethereum-provider/src/EthereumProvider.ts
```diff
@@ -320,6 +324,60 @@ export class EthereumProvider implements IEthereumProvider {
...
+            })
+            .then((result: AuthTypes.AuthenticateResponseResult) => {
+              resolve(result);
+            })
+            .catch((error: Error) => {
+              reject(new Error(error.message));
+            });
+        },
+      );
+      if (!result) return;
```
Cali93: Should we add a log here if `result` is `undefined` to catch potential edge cases ?

providers/ethereum-provider/src/EthereumProvider.ts
```diff
@@ -320,6 +324,60 @@ export class EthereumProvider implements IEthereumProvider {
...
+      const session = result.session;
+      if (session) {
+        const accounts = getAccountsFromNamespaces(session.namespaces, [this.namespace]);
+        // if no required chains are set, use the approved accounts to fetch chainIds
+        this.setChainIds(this.rpc.chains.length ? this.rpc.chains : accounts);
```
Cali93, line 368: This is confusing as `setChainIds` accepts both `chains` and `accounts`

providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -138,6 +139,32 @@ export class UniversalProvider implements IUniversalProvider {
...
+  public async authenticate(opts: AuthenticateParams) {
+    if (!this.client) {
+      throw new Error("Sign Client not initialized");
+    }
+    this.setNamespaces(opts);
+    await this.cleanupPendingPairings();
+
+    const { uri, response } = await this.client.authenticate({
+      ...opts,
+    });
```
Cali93: No need to spread here 
```suggestion
    const { uri, response } = await this.client.authenticate(opts);
```

Cali93 (commented):
Super cool to see this coming to life! 💪 
Added a few comments & suggestions.

packages/types/src/sign-client/auth.ts
ganchoradkov: `expiry` the public API, `ttl` is derived from the `expiry` devs are setting

packages/utils/src/namespaces.ts
ganchoradkov: this first iteration supports only EVM. Later on, potentially more will be added but good suggestion. Will add it

packages/utils/src/signatures.ts
ganchoradkov: imo restricts the usage. This can be applied to any of our projects that validate signature even where logger is not available

providers/ethereum-provider/src/EthereumProvider.ts
ganchoradkov: actually the condition was redundant as if result is not available, the promise will throw anyway

providers/ethereum-provider/src/EthereumProvider.ts
ganchoradkov: added additional clarification - the thing is both chains & accounts contain ` <namespace>:<chainId>` so chains can be taken from both

packages/sign-client/src/client.ts
```diff
@@ -200,6 +230,42 @@ export class SignClient extends ISignClient {
...
+  };
+
+  public rejectSessionAuthenticate: ISignClient["rejectSessionAuthenticate"] = (params) => {
+    try {
+      return this.engine.rejectSessionAuthenticate(params);
+    } catch (error: any) {
+      this.logger.error(error.message);
+      throw error;
+    }
+  };
```
Cali93: It's missing `await` to catch the error.
```suggestion
  public approveSessionAuthenticate: ISignClient["approveSessionAuthenticate"] = (params) => {
    try {
      return await this.engine.approveSessionAuthenticate(params);
    } catch (error: any) {
      this.logger.error(error.message);
      throw error;
    }
  };

  public rejectSessionAuthenticate: ISignClient["rejectSessionAuthenticate"] = (params) => {
    try {
      return await this.engine.rejectSessionAuthenticate(params);
    } catch (error: any) {
      this.logger.error(error.message);
      throw error;
    }
  };
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +591,402 @@ export class Engine extends IEngine {
...
 
+  // ---------- Auth ------------------------------------------------ //
+
+  public authenticate: IEngine["authenticate"] = async (params) => {
+    this.isInitialized();
+    this.isValidAuthenticate(params);
+
+    const { chains, statement = "", uri, domain, nonce, type, exp, nbf, methods = [] } = params;
+    // reassign resources to remove reference as the array is modified and might cause side effects
+    const resources = [...(params.resources || [])];
```
Cali93, line 609: This can be simplified and avoid spreading over the array.
```suggestion
    const resources = params.resources ?? [];
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +591,402 @@ export class Engine extends IEngine {
...
+
+    const { topic: pairingTopic, uri: connectionUri } = await this.client.core.pairing.create({
+      methods: ["wc_sessionAuthenticate"],
```
Cali93, line 612: We're referencing this hardcoded `wc_sessionAuthenticate` string in many places. Can we use a constant and replace any hardcoded occurrence ?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +591,402 @@ export class Engine extends IEngine {
...
+    this.client.logger.info({
+      message: "Generated new pairing",
+      pairing: { topic: pairingTopic, uri: connectionUri },
+    });
+
+    const publicKey = await this.client.core.crypto.generateKeyPair();
+    const responseTopic = hashKey(publicKey);
+
+    await this.client.auth.authKeys.set(AUTH_PUBLIC_KEY_NAME, { responseTopic, publicKey });
+    await this.client.auth.pairingTopics.set(responseTopic, { topic: responseTopic, pairingTopic });
```
Cali93: Is there a reason to do this sequentially or can we use `Promise.all` ?
```suggestion
    await Promise.all([
      this.client.auth.authKeys.set(AUTH_PUBLIC_KEY_NAME, { responseTopic, publicKey }),
      this.client.auth.pairingTopics.set(responseTopic, { topic: responseTopic, pairingTopic })
    ]);
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +591,402 @@ export class Engine extends IEngine {
...
+
+    // Subscribe to response topic
+    await this.client.core.relayer.subscribe(responseTopic);
+
+    this.client.logger.info(`sending request to new pairing topic: ${pairingTopic}`);
+
+    if (methods.length > 0) {
+      const namespace = chains[0].split(":")[0];
```
Cali93: Can we use a util here ?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +591,402 @@ export class Engine extends IEngine {
...
+      },
+      requester: { publicKey, metadata: this.client.metadata },
+      expiryTimestamp,
+    };
+
+    // ----- build namespaces for fallback session proposal ----- //
+    const namespaces = {
+      eip155: {
+        chains,
+        // request `personal_sign` method by default to allow for fallback siwe
```
Cali93, line 668:
```suggestion
        // request `personal_sign` method by default to allow for fallback siwe and deduplicate methods
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +591,402 @@ export class Engine extends IEngine {
...
+      expiryTimestamp,
+    };
+
+    const { done, resolve, reject } = createDelayedPromise(
+      ENGINE_RPC_OPTS.wc_sessionAuthenticate.req.ttl,
+      "Request expired",
+    );
+
+    // handle fallback session proposal response
+    const onSessionConnect = async ({ error, session }: any) => {
```
Cali93, line 691: Can we add types here ?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +591,402 @@ export class Engine extends IEngine {
...
+          });
+        }
+        const sessionObject = this.client.session.get(session.topic);
+        resolve({
+          session: sessionObject,
+        });
+      }
+    };
+    // handle session authenticate response
+    const onAuthenticate = async (payload: any) => {
```
Cali93, line 712: Same here, can we add types please ?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +591,402 @@ export class Engine extends IEngine {
...
+
+      const pendingRequest = this.client.auth.requests.get(id);
+      if (!pendingRequest) {
+        throw new Error(`Could not find pending auth request with id ${id}`);
+      }
+
+      const approvedMethods: string[] = [];
+      const approvedAccounts: string[] = [];
+      for (const cacao of cacaos) {
+        const isValid = await validateSignedCacao({ cacao, projectId: this.client.core.projectId });
```
Cali93, line 733: Any reason to do this sequentially ?
We could speed things up if we use `Promise.all` or `Promise.allSettled` depending on your use case.
eg: 
```js
const cacaoPromises = cacaos.map(async => {
 const isValid = await validateSignedCacao({ cacao, projectId: this.client.core.projectId });
 // keep going with the rest of the logic
})
await Promise.all(cacaoPromises)
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +591,402 @@ export class Engine extends IEngine {
...
+          peer: responder,
+          controller: responder.publicKey,
+          expiry: calcExpiry(SESSION_EXPIRY),
+          requiredNamespaces: {},
+          optionalNamespaces: {},
+          relay: { protocol: "irn" },
+          pairingTopic,
+          namespaces: buildNamespacesFromAuth(
+            [...new Set(approvedMethods)],
+            [...new Set(approvedAccounts)],
```
Cali93, line 781: I assume you're doing this to dedup the arrays but I'm curious what are the cases we'd have duplicates ?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +591,402 @@ export class Engine extends IEngine {
...
+        }),
+        this.sendRequest({
+          topic: pairingTopic,
+          method: "wc_sessionPropose",
+          params: proposal,
+          expiry: ENGINE_RPC_OPTS.wc_sessionPropose.req.ttl,
+          throwOnFailedPublish: true,
+          clientRpcId: fallbackId,
+        }),
+      ]);
```
Cali93, line 823: If one fails before the other it'll throw an error. Is this the desired behaviour ?
Based on your comment it sounds like `Promise.allSettled` is more appropriate here.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +591,402 @@ export class Engine extends IEngine {
...
+    const encodeOpts = {
+      type: TYPE_1,
+      receiverPublicKey,
+      senderPublicKey,
+    };
+
+    const approvedMethods: string[] = [];
+    const approvedAccounts: string[] = [];
+    for (const cacao of auths) {
+      const isValid = await validateSignedCacao({ cacao, projectId: this.client.core.projectId });
```
Cali93, line 874: Since we're running the validation on `authenticate`, do we still need to validate `approveSessionAuthenticate` or is it safe to assume the proposal has valid cacaos ?
If we still need the validation, same suggestion to run this in parallel rather than sequentially.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +591,402 @@ export class Engine extends IEngine {
...
+          encodeOpts,
+        });
+
+        throw new Error(invalidErr.message);
+      }
+
+      const { p: payload } = cacao;
+      const recap = getRecapFromResources(payload.resources);
+
+      const approvedChains: string[] = [getNamespacedDidChainId(payload.iss) as string];
```
Cali93, line 894: `getNamespacedDidChainId` could potentially return `undefined`
```suggestion
      const namespacedDidChainId =  getNamespacedDidChainId(payload.iss)
      const approvedChains: string[] = namespacedDidChainId ? [namespacedDidChainId] : [];
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +591,402 @@ export class Engine extends IEngine {
...
+
+      if (recap) {
+        const methodsfromRecap = getMethodsFromRecap(recap);
+        const chainsFromRecap = getChainsFromRecap(recap);
+        approvedMethods.push(...methodsfromRecap);
+        approvedChains.push(...chainsFromRecap);
+      }
+
+      approvedAccounts.push(...approvedChains.map((chain) => `${chain}:${parsedAddress}`));
+    }
```
Cali93: You can use a `for of` to improve performance and avoid spreading + mapping here.
```suggestion
for (const chain of approvedChains) {
  approvedAccounts.push(`${chain}:${parsedAddress}`);
}
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -748,11 +1163,12 @@ export class Engine extends IEngine {
     await this.client.core.history.resolve(payload);
   };
 
-  private sendError: EnginePrivate["sendError"] = async (id, topic, error) => {
+  private sendError: EnginePrivate["sendError"] = async (params) => {
+    const { id, topic, error, encodeOpts } = params;
```
Cali93, line 1175: `encodeOpts` sounds like a function.
```suggestion
    const { id, topic, error, encodedOpts } = params;
```

Cali93 (commented):
Very impressive stuff going on here 💪 
I'm almost half way through the PR, I'll try to come back at it later this week.

packages/sign-client/src/controllers/engine.ts
ganchoradkov: >If one fails before the other it'll throw an error. Is this the desired behaviour ?

Yes, we need both to succeed publishing in order to continue with the flows

packages/sign-client/src/controllers/engine.ts
ganchoradkov: Decided to intentially omit this check as `payload.iss` is already validated during signature validations

packages/sign-client/src/controllers/engine.ts
ganchoradkov: given that we need to respond if the validation fails, running in parallel with multiple invalid `cacaos` will result multiple error responces that can cause other potential issues. Lets keep it as is, and if we deem that performance is suffering will look into optimizing

packages/sign-client/src/controllers/engine.ts
ganchoradkov: I would leave it as is because this is param for `encode` fx and imo seems intuitive when you go over the code

packages/sign-client/src/controllers/engine.ts
ganchoradkov: It was done as you suggest but as described in the comment, we modify the `resources` array which might not be expected by the implementing party leading to side effects

packages/sign-client/src/controllers/engine.ts
ganchoradkov: Lets tackle this during tech debt as all protocol methods are hardcoded thoughout Sign client and would look weird to have just this one as constant but rest not

packages/web3wallet/src/controllers/engine.ts
```diff
@@ -25,7 +25,7 @@ export class Engine extends IWeb3WalletEngine {
       metadata: this.client.metadata,
     });
 
-    this.initializeEventListeners();
+    // this.initializeEventListeners();
```
bkrem: Can be removed?

packages/web3wallet/src/client.ts
```diff
@@ -32,19 +32,19 @@ export class Web3Wallet extends IWeb3Wallet {
...
   };
 
   public off: IWeb3Wallet["off"] = (name, listener) => {
-    return this.events.off(name, listener);
+    return this.engine.off(name, listener);
   };
 
   public removeListener: IWeb3Wallet["removeListener"] = (name, listener) => {
-    return this.events.removeListener(name, listener);
+    return this.engine.removeListener(name, listener);
```
bkrem, line 47: Context? Event listeners are being merged into the core to avoid `web3wallet` client having its own set?

packages/sign-client/src/client.ts
```diff
@@ -49,6 +62,23 @@ export class SignClient extends ISignClient {
     this.proposal = new Proposal(this.core, this.logger);
     this.pendingRequest = new PendingRequest(this.core, this.logger);
     this.engine = new Engine(this);
+    this.auth = {
+      authKeys: new Store(
```
bkrem: Nit: would be nicer to abstract these like with `Session`, `Proposal` stores etc but not pressing if this is the only occurrence of these stores atm

packages/sign-client/src/constants/engine.ts
```diff
@@ -101,6 +101,18 @@ export const ENGINE_RPC_OPTS: EngineTypes.RpcOptsMap = {
...
   },
+  wc_sessionAuthenticate: {
+    req: {
+      ttl: ONE_HOUR,
+      prompt: true,
+      tag: 1116,
+    },
+    res: {
+      ttl: ONE_HOUR,
+      prompt: false,
```
bkrem, line 112: `prompt` was removed from the spec a while back: https://specs.walletconnect.com/2.0/specs/clients/sign/rpc-methods

Do we still depend on it somehow in the JS implementation?

packages/sign-client/src/controllers/engine.ts
bkrem: +1

packages/sign-client/src/controllers/engine.ts
bkrem: +1

bkrem (commented):
Posting initial batch of notes, will come back to this after interview calls/throughout the day

packages/web3wallet/src/client.ts
ganchoradkov: This is done in order to make sure we're handling `session_authenticate` requests only when wallet is subscribed for it, otherwise fallback to `session_propose`

packages/sign-client/src/constants/engine.ts
ganchoradkov: `prompt` is still present in all other methods + types so lets keep it for consistency

packages/sign-client/src/client.ts
```diff
@@ -210,6 +248,9 @@ export class SignClient extends ISignClient {
       await this.proposal.init();
       await this.pendingRequest.init();
       await this.engine.init();
+      await this.auth.authKeys.init();
+      await this.auth.pairingTopics.init();
+      await this.auth.requests.init();
```
devceline: (optional) (pedant): Would be much cleaner to match the rest and just have a 
```ts
this.auth.init()
```

packages/sign-client/src/controllers/authStore.ts
```diff
@@ -0,0 +1,16 @@
...
+import { IAuth, ICore } from "@walletconnect/types";
+import { AuthPairingTopic } from "./authPairingTopic";
+import { AuthRequest } from "./authRequest";
+import { AuthKey } from "./authKey";
+
+export class AuthStore {
+  public authKeys: IAuth["authKeys"];
+  public pairingTopics: IAuth["pairingTopics"];
+  public requests: IAuth["requests"];
+  constructor(public core: ICore, public logger: Logger) {
```
devceline, line 12:
```suggestion
  
  constructor(public core: ICore, public logger: Logger) {
```

packages/sign-client/src/controllers/authStore.ts
```diff
@@ -0,0 +1,16 @@
...
+    this.authKeys = new AuthKey(this.core, this.logger);
+    this.pairingTopics = new AuthPairingTopic(this.core, this.logger);
+    this.requests = new AuthRequest(this.core, this.logger);
+  }
```
devceline, line 16:
```suggestion
  }
  
  public async init() {
    await this.authKeys.init();
    await this.pairingTopics.init();
    await this.requests.init();
  }
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +592,408 @@ export class Engine extends IEngine {
...
 
+  // ---------- Auth ------------------------------------------------ //
+
+  public authenticate: IEngine["authenticate"] = async (params) => {
+    this.isInitialized();
+    this.isValidAuthenticate(params);
+
+    const { chains, statement = "", uri, domain, nonce, type, exp, nbf, methods = [] } = params;
+    // reassign resources to remove reference as the array is modified and might cause side effects
+    const resources = [...(params.resources || [])];
```
devceline, line 609: image

Potentially **Important**: This does not create a deep copy and we might face unexepected side effects since the array is modified in place.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +592,408 @@ export class Engine extends IEngine {
...
+        const parsedAddress = getDidAddress(payload.iss) as string;
+
+        if (recap) {
+          const methodsfromRecap = getMethodsFromRecap(recap);
+          const chainsFromRecap = getChainsFromRecap(recap);
+          approvedMethods.push(...methodsfromRecap);
+          approvedChains.push(...chainsFromRecap);
+        }
+
+        // approvedAccounts.push(...approvedChains.map((chain) => `${chain}:${parsedAddress}`));
```
devceline:
```suggestion
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1776,6 +2337,17 @@ export class Engine extends IEngine {
     });
   };
 
+  // private getPendingAuthRequests = () => {
+  //   return this.client.auth.requests
+  //     .getAll()
+  //     .filter((value) => typeof value === "object") as unknown as AuthTypes.PendingRequest[];
+  // };
```
devceline:
```suggestion
```

packages/sign-client/src/controllers/authPairingTopic.ts
```diff
@@ -0,0 +1,11 @@
+import { Store } from "@walletconnect/core";
+import { Logger } from "@walletconnect/logger";
+import { ICore } from "@walletconnect/types";
+
+import { AUTH_KEYS_CONTEXT, AUTH_PAIRING_TOPIC_CONTEXT } from "../constants";
+
+export class AuthPairingTopic extends Store<string, { topic: string; pairingTopic: string }> {
+  constructor(public core: ICore, public logger: Logger) {
+    super(core, logger, AUTH_KEYS_CONTEXT, AUTH_PAIRING_TOPIC_CONTEXT);
```
bkrem: Why is this consuming two context consts, one of them being `AUTH_KEYS_CONTEXT`? Is this missing `AUTH_STORAGE_PREFIX`?

packages/sign-client/src/controllers/engine.ts
bkrem: +1, spread operator works for primitive types like strings and numbers at the root level of the object-to-be-cloned, but any nested object types (array, object etc) are shallow reference copies.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +592,408 @@ export class Engine extends IEngine {
...
+          session: sessionObject,
+        });
+      }
+    };
+    // handle session authenticate response
+    const onAuthenticate = async (payload: any) => {
+      // remove cleanup for fallback response
+      this.events.off(engineEvent("session_connect"), onSessionConnect);
+      if (payload.error) {
+        // ignore unsupported method error
```
bkrem: Explains the "what" but not the "why", context pls 🙏

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +592,408 @@ export class Engine extends IEngine {
...
+      }
+      const {
+        cacaos,
+        responder,
+      }: {
+        cacaos: AuthTypes.SessionAuthenticateResponseParams["cacaos"];
+        responder: AuthTypes.SessionAuthenticateResponseParams["responder"];
+      } = payload.result;
+
+      const pendingRequest = this.client.auth.requests.get(id);
```
bkrem: Will this throw by default?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -572,6 +592,408 @@ export class Engine extends IEngine {
...
+      if (!pendingRequest) {
+        throw new Error(`Could not find pending auth request with id ${id}`);
+      }
+
+      const approvedMethods: string[] = [];
+      const approvedAccounts: string[] = [];
+      for (const cacao of cacaos) {
+        const isValid = await validateSignedCacao({ cacao, projectId: this.client.core.projectId });
+        if (!isValid) {
+          this.client.logger.error(cacao);
```
bkrem: Do we need more context on the error rather than forwarding just the failed cacao?

packages/sign-client/src/controllers/engine.ts
bkrem: +1

packages/sign-client/src/controllers/engine.ts
```diff
@@ -916,6 +1351,25 @@ export class Engine extends IEngine {
     throw new Error(message);
   };
 
+  private shouldIgnorePairingRequest = (params: { topic: string; requestMethod: string }) => {
```
bkrem: Can we add this fn signature type to `EnginePrivate` pls for consistency and define `boolean` as return type explicitly so it can't return `void` accidentally if someone later touches the early or default returns here 🙏

bkrem (commented):
Still going 🏃

packages/sign-client/src/controllers/engine.ts
ganchoradkov: Current implementation should be fine as `resources` by spec are an array of text values aka `strings`

packages/sign-client/src/controllers/authPairingTopic.ts
ganchoradkov: good catch!

packages/sign-client/src/controllers/engine.ts
ganchoradkov: actually this whole check was redundant as the  `pendingRequest` value is not really used elsewhere

packages/utils/src/signatures.ts
```diff
@@ -0,0 +1,92 @@
+import { AuthTypes } from "@walletconnect/types";
+import { hashMessage } from "@ethersproject/hash";
+import { recoverAddress } from "@ethersproject/transactions";
+export const DEFAULT_RPC_URL = "https://rpc.walletconnect.com/v1";
```
bkrem: Belongs in `constants.ts` if being exported/is non-local to this util

packages/web3wallet/src/controllers/engine.ts
```diff
@@ -129,12 +165,38 @@ export class Engine extends IWeb3WalletEngine {
...
 
-  private initializeEventListeners = () => {
-    this.signClient.events.on("session_proposal", this.onSessionProposal);
-    this.signClient.events.on("session_request", this.onSessionRequest);
-    this.signClient.events.on("session_delete", this.onSessionDelete);
-    this.authClient.on("auth_request", this.onAuthRequest);
-    this.signClient.events.on("proposal_expire", this.onProposalExpire);
-    this.signClient.events.on("session_request_expire", this.onSessionRequestExpire);
+  private onSessionRequestAuthenticate = (event: Web3WalletTypes.SessionAuthenticate) => {
+    // console.log("onSessionRequestAuthenticate", event);
```
bkrem:
```suggestion
```

packages/web3wallet/src/controllers/engine.ts
```diff
@@ -129,12 +165,38 @@ export class Engine extends IWeb3WalletEngine {
...
+        this.authClient[action]("auth_request", this.onAuthRequest);
+        break;
+      case "proposal_expire":
+        this.signClient.events[action]("proposal_expire", this.onProposalExpire);
+        break;
+      case "session_request_expire":
+        this.signClient.events[action]("session_request_expire", this.onSessionRequestExpire);
+        break;
+      case "session_authenticate":
+        // console.log("session_authenticate", action);
```
bkrem:
```suggestion
```

bkrem (approved):
💯 💯 

Huge seeing this reach the finish line. Thank you for your continued efforts on this Gancho 🙏

packages/utils/src/signatures.ts
ganchoradkov: removed the `export` so it stays local

devceline (approved):
LGTM

PR #4346, opened 2024-03-20
feat: updates session without waiting for an ack from the peer

ganchoradkov:
## Description
Added improvement to session update that allows wallets to update session withut waiting for the dapp to acknowledge. There are several benefits to this approach such as
- optimized for mobile where almost always only the foreground app is online thus acknowledgement is impossible from app that is in background
- better handles updates on multiple sessions where all/almost all peers are offline
- allows client to immediately utilize the updated session such as emit events with the updated valus

## Type of change

## How has this been tested?
dogfooding, tests

## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -373,7 +373,7 @@ export class Engine extends IEngine {
         resolve();
       }
     });
-    // update the session with the new namespaces, if the publish fails, revert to the old
+    // Update the session with the new namespaces, if the publish fails, revert to the old.
     // This allows the client to use the updated session like emitting events
     // without waiting for the peer to acknowledge
     await this.client.session.update(topic, { namespaces });
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -366,7 +366,7 @@ export class Engine extends IEngine {
     const relayRpcId = getBigIntRpcId().toString() as any;
 
     const oldNamespaces = this.client.session.get(topic).namespaces;
-    this.events.once(engineEvent("session_update", clientRpcId), async ({ error }: any) => {
+    this.events.once(engineEvent("session_update", clientRpcId), ({ error }: any) => {
       if (error) reject(error);
       else {
         resolve();
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -369,11 +369,13 @@ export class Engine extends IEngine {
     this.events.once(engineEvent("session_update", clientRpcId), async ({ error }: any) => {
       if (error) reject(error);
       else {
-        await this.client.session.update(topic, { namespaces });
         resolve();
       }
     });
-
+    // update the session with the new namespaces, if the publish fails, revert to the old
+    // this allows the client to use the updated session like emitting events
+    // without waiting for the peer to acknowledge
+    await this.client.session.update(topic, { namespaces });
     this.sendRequest({
       topic,
       method: "wc_sessionUpdate",
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -374,7 +374,7 @@ export class Engine extends IEngine {
       }
     });
     // update the session with the new namespaces, if the publish fails, revert to the old
-    // this allows the client to use the updated session like emitting events
+    // This allows the client to use the updated session like emitting events
     // without waiting for the peer to acknowledge
     await this.client.session.update(topic, { namespaces });
     this.sendRequest({
--- a/packages/web3wallet/src/controllers/engine.ts
+++ b/packages/web3wallet/src/controllers/engine.ts
@@ -47,11 +47,11 @@ export class Engine extends IWeb3WalletEngine {
   };
 
   public updateSession: IWeb3WalletEngine["updateSession"] = async (params) => {
-    return await (await this.signClient.update(params)).acknowledged();
+    return await this.signClient.update(params);
   };
 
   public extendSession: IWeb3WalletEngine["extendSession"] = async (params) => {
-    return await (await this.signClient.extend(params)).acknowledged();
+    return await this.signClient.extend(params);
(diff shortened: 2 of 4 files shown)


packages/sign-client/src/controllers/engine.ts
```diff
@@ -369,11 +369,13 @@ export class Engine extends IEngine {
...
       if (error) reject(error);
       else {
-        await this.client.session.update(topic, { namespaces });
         resolve();
       }
     });
-
+    // update the session with the new namespaces, if the publish fails, revert to the old
+    // this allows the client to use the updated session like emitting events
+    // without waiting for the peer to acknowledge
```
bkrem, line 378: What's the maximum lag here between the client applying the update optimistically and then reverting it due to failed publish? Could see this looking like a weird "time travel" phenomenon, where the client acts on the optimistic update via side effects (like emitting events) and then reverts back to the pre-sessionUpdate state.

packages/web3wallet/src/controllers/engine.ts
```diff
@@ -47,11 +47,11 @@ export class Engine extends IWeb3WalletEngine {
   };
 
   public updateSession: IWeb3WalletEngine["updateSession"] = async (params) => {
-    return await (await this.signClient.update(params)).acknowledged();
+    return await this.signClient.update(params);
   };
 
   public extendSession: IWeb3WalletEngine["extendSession"] = async (params) => {
-    return await (await this.signClient.extend(params)).acknowledged();
```
bkrem, line 57: While obviously nicer in terms of performance, doesn't this weaken guarantees about clients' sessions being in sync? E.g. clientA (calling) ends up with an extended session for it then to expire unexpectedly on clientB (responding) bc it didn't explicitly handle the extension.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -369,11 +369,13 @@ export class Engine extends IEngine {
     this.events.once(engineEvent("session_update", clientRpcId), async ({ error }: any) => {
```
bkrem: GHA code lint: `Async arrow function has no 'await' expression`

packages/web3wallet/src/types/engine.ts
```diff
@@ -38,10 +38,12 @@ export abstract class IWeb3WalletEngine {
...
     topic: string;
     namespaces: SessionTypes.Namespaces;
-  }): Promise<void>;
+  }): Promise<{ acknowledged: () => Promise<void> }>;
 
   // update session expiry (SIGN)
-  public abstract extendSession(params: { topic: string }): Promise<void>;
+  public abstract extendSession(params: {
+    topic: string;
+  }): Promise<{ acknowledged: () => Promise<void> }>;
```
bkrem, line 48: Will require a docs update right? 

Guess this can be seen as non-breaking since we're going from `void` to non-void return type

bkrem (commented):
Would love to see this for responsiveness and performance but have some concerns about breaking base assumptions of the protocol

packages/sign-client/src/controllers/engine.ts
ganchoradkov: reverting happens in maximum a minute if the relay doesn't acknowledge the publish. Once the publish is acked we can safely assume the dapp will receive and process the request as intended.
Additionally, with recent updates to [onSessionUpdateRequest](https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/sign-client/src/controllers/engine.ts#L1063) we can be sure the last sent update will be applied even when there are multiple received when the client is started.

>Could see this looking like a weird "time travel" phenomenon, where the client acts on the optimistic update via side effects (like emitting events) and then reverts back to the pre-sessionUpdate state.

technically yes, but imo it shouldn't be concern as clients are built to handle that case. For example this scenario
- wallet optimistically updates session with chain `x`
- wallet emits event with chain `x`
- publish timeouts and fails
- wallet reverts to session without chain 'x'

then when the dapp comes online it would receive event with chain 'x' but will discard it as it would fail validation https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/sign-client/src/controllers/engine.ts#L1264

packages/web3wallet/src/controllers/engine.ts
ganchoradkov: >doesn't this weaken guarantees about clients' sessions being in sync

Kind of, in cases when the dapp is offline BUT It has the big benefit of actually publishing update to all sessions regardless if the dapp is online. This solves an issue like 

- wallet has two session that wants to update
- one with active dapp (thats online), one without
- wallet tries to update both
- session with online dapp succeeds, session with offline dapp is stuck

With users having tons of active sessions on their wallet, chances are that most will have their peer offline. 
This just clogs the SDK with tasks that will never resolve.

packages/web3wallet/src/types/engine.ts
ganchoradkov: https://github.com/WalletConnect/walletconnect-docs/pull/1511

packages/sign-client/src/controllers/engine.ts
```diff
@@ -367,14 +367,16 @@ export class Engine extends IEngine {
...
-    this.events.once(engineEvent("session_update", clientRpcId), async ({ error }: any) => {
+    this.events.once(engineEvent("session_update", clientRpcId), ({ error }: any) => {
       if (error) reject(error);
       else {
-        await this.client.session.update(topic, { namespaces });
         resolve();
       }
     });
-
+    // update the session with the new namespaces, if the publish fails, revert to the old
```
```suggestion
    // Update the session with the new namespaces, if the publish fails, revert to the old.
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -367,14 +367,16 @@ export class Engine extends IEngine {
...
+    // this allows the client to use the updated session like emitting events
```
```suggestion
    // This allows the client to use the updated session like emitting events
```

bkrem:
Can we link this PR to the docs one pls: https://github.com/WalletConnect/walletconnect-docs/pull/1511

bkrem (approved):

bkrem:
🏎️ 🚢

PR #4373, opened 2024-03-28
feat: optionally disable deeplinks

ganchoradkov:
## Description
Added `sessionConfig` to sessions with a prop `disableDeepLink` that allows wallets to disable deeplink behaviour on session by session basis. Use case is when a wallet is launched in a browser tab, to avoid opening in new tab on every request
context: https://walletconnect.slack.com/archives/C03RVH94K5K/p1711033401102829
## Type of change

## How has this been tested?
dogfooding, tests

## Checklist

https://linear.app/walletconnect/issue/W3M-90/fireblocks-issue-with-opening-new-tab

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -268,7 +268,7 @@ export class Engine extends IEngine {
       controller: { publicKey: selfPublicKey, metadata: this.client.metadata },
       expiry: calcExpiry(SESSION_EXPIRY),
       ...(sessionProperties && { sessionProperties }),
-      sessionConfig,
+      ...(sessionConfig && { sessionConfig }),
     };
     await this.client.core.relayer.subscribe(sessionTopic);
     const session = {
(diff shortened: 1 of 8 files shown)


bkrem:
> Use case is when a wallet is launched in a browser tab, to avoid opening in new tab on every request

What's the context for this? Is there a related GH issue?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -994,9 +999,17 @@ export class Engine extends IEngine {
...
+        relay,
+        controller,
+        expiry,
+        namespaces,
+        sessionProperties,
+        pairingTopic,
+        sessionConfig,
+      } = payload.params;
       const session = {
+        ...payload.params,
```
bkrem: Why is the spread of all `payload.params` being added here?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1015,6 +1028,7 @@ export class Engine extends IEngine {
           metadata: controller.metadata,
         },
         ...(sessionProperties && { sessionProperties }),
+        sessionConfig,
```
bkrem: Do we need to take into account this being possibly `null`? Why are we treating it differently to `sessionProperties` here?

bkrem (approved):

PR #4420, opened 2024-04-08
fix: adds `session_authenticate` to expirer

ganchoradkov:
## Description
- Added `session_authenticate` requests to expirer
- Updated execution order to first set items in expirer and then saving to store
- Updated tests

## Type of change

## How has this been tested?
tests

## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -653,7 +653,7 @@ export class Engine extends IEngine {
       resources.push(recap);
     }
 
-    // ensure the expiry is at least the minimum required for the request - currently 1h
+    // Ensure the expiry is greater than the minimum required for the request - currently 1h
     const authRequestExpiry =
       expiry && expiry > ENGINE_RPC_OPTS.wc_sessionAuthenticate.req.ttl
         ? expiry
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -604,7 +604,18 @@ export class Engine extends IEngine {
     this.isInitialized();
     this.isValidAuthenticate(params);
 
-    const { chains, statement = "", uri, domain, nonce, type, exp, nbf, methods = [] } = params;
+    const {
+      chains,
+      statement = "",
+      uri,
+      domain,
+      nonce,
+      type,
+      exp,
+      nbf,
+      methods = [],
+      expiry,
+    } = params;
     // reassign resources to remove reference as the array is modified and might cause side effects
     const resources = [...(params.resources || [])];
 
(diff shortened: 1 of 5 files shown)


packages/sign-client/src/controllers/engine.ts
```diff
@@ -642,7 +653,12 @@ export class Engine extends IEngine {
       resources.push(recap);
     }
 
-    const expiryTimestamp = calcExpiry(ENGINE_RPC_OPTS.wc_sessionPropose.req.ttl);
+    // ensure the expiry is at least the minimum required for the request - currently 1h
```
```suggestion
    // Ensure the expiry is greater than the minimum required for the request - currently 1h
```

Alternatively we can leave the comment as-is if condition uses `>=` but seems pointless if it can just default to the internal `TTL` anyways if equal.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -720,7 +738,8 @@ export class Engine extends IEngine {
         this.events.off(engineEvent("session_connect"), onSessionConnect);
         return reject(payload.error.message);
       }
-
+      // delete fallback proposal on successful authenticate as the proposal will not be responded to
+      await this.deleteProposal(fallbackId);
```
bkrem, line 742: Nice catch on these authRequest/proposal lifecycle cleanups 💪

packages/sign-client/src/controllers/engine.ts
```diff
@@ -834,14 +853,9 @@ export class Engine extends IEngine {
...
 
     await this.setProposal(fallbackId, { id: fallbackId, ...proposal });
-
-    await this.client.auth.requests.set(id, {
-      authPayload: request.authPayload,
-      requester: request.requester,
-      expiryTimestamp,
-      id,
+    await this.setAuthRequest(id, {
+      request: { ...request, verifyContext: {} as any },
```
bkrem, line 857: I know it's not new but why do we need an `any` cast here? Can we avoid it?

packages/types/src/sign-client/engine.ts
```diff
@@ -214,6 +214,19 @@ export interface EnginePrivate {
 
   setProposal(id: number, proposal: ProposalTypes.Struct): Promise<void>;
 
+  setAuthRequest(
+    id: number,
+    params: {
+      request: {
```
bkrem: Seems like this is basically a straight extension of the `SessionAuthenticateRequestParams` interface (https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/types/src/sign-client/auth.ts#L158) with `verifyContext` added in.

Might benefit from declaring a new interface e.g. `SessionAuthenticateRequest extends SessionAuthenticateRequestParams` so these two types are linked properly.

Otherwise if we ever change `SessionAuthenticateRequestParams ` this inline typing needs to be updated separately.

bkrem (approved):
Thanks for pinning the expected behaviour with the tests 💯

packages/sign-client/src/controllers/engine.ts
ganchoradkov: It's cast to any since there is no verify attestation on the dapp side. I can fill it with dummy data instead of empty {}

packages/types/src/sign-client/engine.ts
ganchoradkov: good call 👍

PR #4447, opened 2024-04-15
feat: disable request queue

ganchoradkov:
## Description
Added new init param `signConfig` with initial `disableRequestQueue` parameter. This option allows wallets to disable the session request queue if they wish.
Additionally, added a few improvements to connection handling because if several days long `socket hang up` issues in CI such as
- always creating new ws instance when attempting to connect
- proper cleanup of `ping` timers when connection is closed intentionally 

## Type of change

## How has this been tested?
tests
canary - `2.12.2-req-q.0`

## Checklist

### DIFF
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -117,7 +117,7 @@ export class Relayer extends IRelayer {
   public async init() {
     this.logger.trace(`Initialized`);
     this.registerEventListeners();
-    await this.createProvider();
+    // await this.createProvider();
     await Promise.all([this.messages.init(), this.subscriber.init()]);
     try {
       await this.transportOpen();
@@ -291,8 +291,8 @@ export class Relayer extends IRelayer {
     if (relayUrl && relayUrl !== this.relayUrl) {
       this.relayUrl = relayUrl;
       await this.transportDisconnect();
-      await this.createProvider();
     }
+    await this.createProvider();
     this.connectionAttemptInProgress = true;
     this.transportExplicitlyClosed = false;
     try {
@@ -321,7 +321,6 @@ export class Relayer extends IRelayer {
       if (!this.isConnectionStalled(error.message)) {
         throw e;
       }
-      this.onProviderDisconnect();
       console.log("reconnecting...");
     } finally {
       this.connectionAttemptInProgress = false;
@@ -334,7 +333,7 @@ export class Relayer extends IRelayer {
     this.relayUrl = relayUrl || this.relayUrl;
     await this.confirmOnlineStateOrThrow();
     await this.transportClose();
-    await this.createProvider();
+    // await this.createProvider();
     await this.transportOpen();
   }
 
@@ -535,7 +534,6 @@ export class Relayer extends IRelayer {
     if (this.transportExplicitlyClosed) return;
 
     setTimeout(async () => {
-      await this.createProvider();
       await this.transportOpen().catch((error) => this.logger.error(error));
     }, toMiliseconds(RELAYER_RECONNECT_TIMEOUT));
   }
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -330,7 +330,7 @@ export class Relayer extends IRelayer {
         `ID: ${this.relayerId}`,
         `elapsed: ${performance.now() - this.start}ms`,
       );
-      this.logger.error(e);
+      this.logger.error(e, `ID: ${this.relayerId}`);
       const error = e as Error;
       if (!this.isConnectionStalled(error.message)) {
         throw e;
@@ -393,6 +393,7 @@ export class Relayer extends IRelayer {
           "pingTimeout called",
           `ID: ${this.relayerId}, elapsed: ${performance.now() - this.start}ms`,
         );
+        console.log("terminating connection", `ID: ${this.relayerId}`);
         //@ts-expect-error
         this.provider?.connection?.socket?.terminate();
       }, this.heartBeatTimeout);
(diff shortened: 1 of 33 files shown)


arein:
seems like there's also a bug fix with `requestsInFlight.clear()`?

ganchoradkov:
>seems like there's also a bug fix with requestsInFlight.clear()?

I'm testing what might be causing the network drops

packages/core/src/controllers/relayer.ts
```diff
@@ -286,8 +286,10 @@ export class Relayer extends IRelayer {
     if (relayUrl && relayUrl !== this.relayUrl) {
       this.relayUrl = relayUrl;
       await this.transportDisconnect();
-      await this.createProvider();
     }
+    // Always create new socket instance when trying to connect because if the socket is dropped due to `socket hang up` exception
+    // It wont be able to reconnect
+    await this.createProvider();
```
devceline: Does this have any risk of unnecessarily resetting the socket ?

packages/core/src/controllers/relayer.ts
```diff
@@ -286,8 +286,10 @@ export class Relayer extends IRelayer {
     if (relayUrl && relayUrl !== this.relayUrl) {
       this.relayUrl = relayUrl;
       await this.transportDisconnect();
-      await this.createProvider();
     }
+    // Always create new socket instance when trying to connect because if the socket is dropped due to `socket hang up` exception
```
```suggestion
    // Always create new socket instance when trying to connect because if the socket was dropped due to `socket hang up` exception
```

packages/core/src/controllers/relayer.ts
```diff
@@ -395,6 +397,7 @@ export class Relayer extends IRelayer {
   private async createProvider() {
     if (this.provider.connection) {
       this.unregisterProviderListeners();
+      this.provider = undefined as any;
```
bkrem: Why do we need this step if it's being reassigned either way below?

packages/core/src/controllers/relayer.ts
```diff
@@ -315,6 +317,7 @@ export class Relayer extends IRelayer {
       if (!this.isConnectionStalled(error.message)) {
         throw e;
       }
+      this.hasExperiencedNetworkDisruption = true;
```
devceline: Shouldn't it be this because with the way it's written if it's an error other than stalled connection, we'd never flag `hasExperiencedNetworkDisruption` as true
```suggestion
      this.hasExperiencedNetworkDisruption = true;
      if (!this.isConnectionStalled(error.message)) {
        throw e;
      }
```

packages/core/src/controllers/relayer.ts
devceline: This breaks the API of `provider`. 

This might cause bugs we'd need to track down later since everything dealing with provider expects `IJsonRpcProvider` and not something potentially `undefined`.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -396,7 +396,7 @@ export class Engine extends IEngine {
     // This allows the client to use the updated session like emitting events
     // without waiting for the peer to acknowledge
     await this.client.session.update(topic, { namespaces });
-    this.sendRequest({
+    await this.sendRequest({
```
devceline, line 399: This updates the behavior of `update` to not be as "Fire and forget" - Are there any implications of this we need to consider?

Or is this fine because of the reasoning of
"If it's fired by the queue, no big deal since the queue itself is not awaited. and if someone disabled the queue they'd want to wait"

Just want to be on the same page

packages/types/src/sign-client/client.ts
```diff
@@ -60,9 +60,14 @@ export declare namespace SignClientTypes {
 
   type Metadata = CoreTypes.Metadata;
 
+  type SignConfig = {
+    disableRequestQueue?: boolean;
+  };
```
devceline, line 65: nitpick, optional: shouldn't this be in types?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1137,7 +1137,7 @@ export class Engine extends IEngine {
     const { id, topic, params, verifyContext } = pendingRequest;
     const expiry =
       params.request.expiryTimestamp || calcExpiry(ENGINE_RPC_OPTS.wc_sessionRequest.req.ttl);
-    this.client.core.expirer.set(id, expiry);
+    if (expiry) this.client.core.expirer.set(id, expiry);
```
bkrem: Why do we need to guard the value as truthy if it should always have a default?

Is this what is wanted?
```suggestion
    if (expiry > 0) this.client.core.expirer.set(id, expiry);
```

packages/sign-client/test/shared/helpers.ts
```diff
@@ -17,7 +17,9 @@ export async function deleteClients(clients: {
     client.core.relayer.provider.connection.events.removeAllListeners();
     client.events.removeAllListeners();
   }
+  console.log("Client A connected:", clients.A?.core.relayer.connected);
```
bkrem: Still wanted?

bkrem (commented):
Thanks for the solid test coverage on the new option 🙏

packages/core/src/controllers/relayer.ts
ganchoradkov: nope, it will create new instance instead of trying to restart the existing socket which isn't possible if the socket is stalled anyway

packages/core/src/controllers/relayer.ts
ganchoradkov: `this.hasExperiencedNetworkDisruption` flag is used to determine how to handle `disconnect` logic. When socket stalls on connect, there is no disconnect flow involved so it isn't really necessary

packages/sign-client/src/controllers/engine.ts
ganchoradkov: this resolves flaky tests but also make sending request(update) consistent and allows using the updated session immediately

packages/types/src/sign-client/client.ts
ganchoradkov: what do you mean? It is in `packages/types/src/sign-client/client.ts`

packages/core/src/controllers/relayer.ts
ganchoradkov: removed

packages/sign-client/src/controllers/engine.ts
ganchoradkov: the condition was copied along the `expirer.set` from other flows where expiry didn't have default value, removed it

bkrem (approved):

PR #4459, opened 2024-04-18
feat: implements `irn_batchFetchMessages`

ganchoradkov:
## Description
Implemented `irn_batchFetchMessages` that is called before subscribing to topics so we can sort all pending messages from the mailbox and avoid out of sync messages and improve consistency in handling when a peer comes online and receives bunch of mailbox messages

## Type of change

## How has this been tested?
this test covers exactly this scenario https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/web3wallet/test/sign.spec.ts#L142
## Checklist

### DIFF
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -334,6 +334,23 @@ export class Relayer extends IRelayer {
     throw new Error("No internet connection detected. Please restart your network and try again.");
   }
 
+  public async handleBatchMessageEvents(messages: RelayerTypes.MessageEvent[]) {
+    if (messages?.length === 0) {
+      this.logger.trace("Batch message events is empty. Ignoring...");
+      return;
+    }
+    try {
+      const sortedMessages = messages.sort((a, b) => a.publishedAt - b.publishedAt);
+      this.logger.trace(`Batch message events sorted`);
+      for (const message of sortedMessages) {
+        await this.onMessageEvent(message);
+      }
+      this.logger.trace(`Batch message events processed`);
+    } catch (e) {
+      this.logger.warn(e);
+    }
+  }
+
   // ---------- Private ----------------------------------------------- //
   /*
    * In Node, we must detect when the connection is stalled and terminate it.
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -339,16 +339,16 @@ export class Relayer extends IRelayer {
       this.logger.trace("Batch message events is empty. Ignoring...");
       return;
     }
-    try {
-      const sortedMessages = messages.sort((a, b) => a.publishedAt - b.publishedAt);
-      this.logger.trace(`Batch message events sorted`);
-      for (const message of sortedMessages) {
+    const sortedMessages = messages.sort((a, b) => a.publishedAt - b.publishedAt);
+    this.logger.trace(`Batch of ${sortedMessages.length} message events sorted`);
+    for (const message of sortedMessages) {
+      try {
         await this.onMessageEvent(message);
+      } catch (e) {
+        this.logger.warn(e);
       }
-      this.logger.trace(`Batch message events processed`);
-    } catch (e) {
-      this.logger.warn(e);
     }
+    this.logger.trace(`Batch of ${sortedMessages.length} message events processed`);
   }
 
   // ---------- Private ----------------------------------------------- //
--- a/packages/core/src/controllers/subscriber.ts
+++ b/packages/core/src/controllers/subscriber.ts
@@ -277,15 +277,19 @@ export class Subscriber extends ISubscriber {
     };
     this.logger.debug(`Outgoing Relay Payload`);
     this.logger.trace({ type: "payload", direction: "outgoing", request });
+    let result;
     try {
       const fetchMessages = await createExpiringPromise(
         this.relayer.request(request).catch((e) => this.logger.warn(e)),
         this.subscribeTimeout,
       );
-      return await fetchMessages;
+      result = (await fetchMessages) as {
+        messages: RelayerTypes.MessageEvent[];
+      };
     } catch (err) {
       this.relayer.events.emit(RELAYER_EVENTS.connection_stalled);
     }
+    return result;
   }
 
   private rpcUnsubscribe(topic: string, id: string, relay: RelayerTypes.ProtocolOptions) {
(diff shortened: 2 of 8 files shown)


packages/core/src/controllers/subscriber.ts
```diff
@@ -263,6 +265,29 @@ export class Subscriber extends ISubscriber {
     }
   }
 
+  private async rpcBatchFetchMessages(subscriptions: SubscriberTypes.Params[]) {
+    if (!subscriptions.length) return;
+    // const relay = subscriptions[0].relay;
```
bkrem: Needed?

ganchoradkov:
waiting on https://github.com/WalletConnect/walletconnect-utils/pull/170 to be merged/release so I can bump versions in this PR

packages/core/src/controllers/subscriber.ts
```diff
@@ -263,6 +265,29 @@ export class Subscriber extends ISubscriber {
...
+    this.logger.debug(`Outgoing Relay Payload`);
+    this.logger.trace({ type: "payload", direction: "outgoing", request });
+    try {
+      const fetchMessages = await createExpiringPromise(
+        this.relayer.request(request).catch((e) => this.logger.warn(e)),
+        this.subscribeTimeout,
+      );
+      return await fetchMessages;
+    } catch (err) {
+      this.relayer.events.emit(RELAYER_EVENTS.connection_stalled);
```
devceline, line 290: Is this the only possible cause for the error?

ganchoradkov: pretty much yes, other failures e.g. in `.relayer.request` are logged

bkrem:
> waiting on [WalletConnect/walletconnect-utils#170](https://github.com/WalletConnect/walletconnect-utils/pull/170) to be merged/release so I can bump versions in this PR

Approved ✅

packages/core/src/controllers/relayer.ts
```diff
@@ -334,6 +334,23 @@ export class Relayer extends IRelayer {
...
   }
 
+  public async handleBatchMessageEvents(messages: RelayerTypes.MessageEvent[]) {
+    if (messages?.length === 0) {
+      this.logger.trace("Batch message events is empty. Ignoring...");
+      return;
+    }
+    try {
+      const sortedMessages = messages.sort((a, b) => a.publishedAt - b.publishedAt);
+      this.logger.trace(`Batch message events sorted`);
```
bkrem: Would be good to trace the batch size too:

```suggestion
      this.logger.trace(`Batch of ${sortedMessages.length} message events sorted`);
```

packages/core/src/controllers/relayer.ts
```diff
@@ -334,6 +334,23 @@ export class Relayer extends IRelayer {
...
+      for (const message of sortedMessages) {
+        await this.onMessageEvent(message);
+      }
+      this.logger.trace(`Batch message events processed`);
```
```suggestion
      this.logger.trace(`Batch of ${sortedMessages.length} message events processed`);
```

packages/core/src/controllers/subscriber.ts
```diff
@@ -364,6 +389,7 @@ export class Subscriber extends ISubscriber {
       const batches = Math.ceil(this.cached.length / this.batchSubscribeTopicsLimit);
```
bkrem: Nit: `numOfBatches` would be clearer here to indicate that this is a number and not related the `batch` assignment below
```suggestion
      const numOfBatches = Math.ceil(this.cached.length / this.batchSubscribeTopicsLimit);
```

packages/core/src/controllers/subscriber.ts
```diff
@@ -399,6 +425,17 @@ export class Subscriber extends ISubscriber {
     this.onBatchSubscribe(result.map((id, i) => ({ ...subscriptions[i], id })));
   }
 
+  private async batchFetchMessages(subscriptions: SubscriberTypes.Params[]) {
+    if (!subscriptions.length) return;
```
bkrem, line 433: Is this `subscriptions` guaranteed to always be an array?

packages/core/src/controllers/subscriber.ts
```diff
@@ -399,6 +425,17 @@ export class Subscriber extends ISubscriber {
...
+    this.logger.trace(`Fetching batch messages for ${subscriptions.length} subscriptions`);
+    const response = (await this.rpcBatchFetchMessages(subscriptions)) as {
```
bkrem: Why are we casting? Can we attach the expected return type to `rpcBatchFetchMessages` directly?

packages/core/src/controllers/relayer.ts
```diff
@@ -334,6 +334,23 @@ export class Relayer extends IRelayer {
...
+  public async handleBatchMessageEvents(messages: RelayerTypes.MessageEvent[]) {
+    if (messages?.length === 0) {
+      this.logger.trace("Batch message events is empty. Ignoring...");
+      return;
+    }
+    try {
+      const sortedMessages = messages.sort((a, b) => a.publishedAt - b.publishedAt);
+      this.logger.trace(`Batch message events sorted`);
+      for (const message of sortedMessages) {
+        await this.onMessageEvent(message);
```
bkrem, line 346: Thought: if there's an invalid message event half way through, the throw here will then discard all other (possibly) valid messages afterwards.

Could be safer to try/catch on the individual message level instead?

packages/core/src/controllers/subscriber.ts
ganchoradkov: yes, when there are no messages, response is `[]`

ganchoradkov: I can still check that value exist as a precaution

ganchoradkov: Yes, it uses same values/types as `batchSubscribe` which has been in prod for many months

devceline (approved):
LGTM

packages/core/src/controllers/subscriber.ts
```diff
@@ -277,15 +277,19 @@ export class Subscriber extends ISubscriber {
     };
     this.logger.debug(`Outgoing Relay Payload`);
     this.logger.trace({ type: "payload", direction: "outgoing", request });
+    let result;
     try {
       const fetchMessages = await createExpiringPromise(
```
bkrem:
```suggestion
      const fetchMessagesPromise = await createExpiringPromise(
```

bkrem (approved):

PR #4486, opened 2024-05-02
Feat/wallet get capabilities cache

ganchoradkov:
## Description
Implemented `wallet_getCapabilities` caching in `universal-provider`

## Type of change

## How has this been tested?
tests 

## Checklist


providers/universal-provider/src/providers/eip155.ts
```diff
@@ -168,6 +170,34 @@ class Eip155Provider implements IProvider {
...
+    const capabilities = await this.client.request(args as EngineTypes.RequestParams);
+    try {
+      // update the session with the capabilities so they can be retrieved later
+      await this.client.session.update(args.topic, {
+        sessionProperties: {
+          ...(session.sessionProperties || {}),
+          capabilities: {
+            ...(sessionCapabilities || {}),
+            [address]: capabilities,
+          } as any, // by spec sessionProperties should be <string, string> but here are used as objects?
```
arein, line 193: what do we do about this? The `spec` is the EIP-5792 spec. I think we do something else maybe in our CAIP-25 spec? So maybe we just cover the "right" think in our wallet docs / wallet SDK?

We probably need to expand our namespace builder for this usecase

providers/universal-provider/test/index.spec.ts
```diff
@@ -1001,6 +1002,78 @@ describe("UniversalProvider", function () {
           expectedChainId,
         });
       });
+      it("should cache `wallet_getCapabilities` request", async () => {
```
arein, line 1005: do you need a negative test here if not cached?

arein (approved):

providers/universal-provider/src/providers/eip155.ts
ganchoradkov: according to `caip-25` spec the sessionProperties (scopeProperties) is `<string, string>`

providers/universal-provider/test/index.spec.ts
ganchoradkov: the first part of the test where the wallet receives the request covers the not cached part. If it was cached already the request wouldn't have been sent to the wallet

tomiir (approved):

PR #4499, opened 2024-05-07
feat:add eip5792 methods as optional methods

KannuSingh:
Adding EIP-5792 methods on the list of optional RPC methods. These method will become part of session proposal request to wallet.
This will enable wallet supporting EIP-5792 allowing dapps to use it after session is successfully approved by the user.


tomiir (commented):
LGTM but would wait for an approval from @ganchoradkov

arein (commented):
triple-checking: are these all just the optional methods?

KannuSingh:
IMO Yes, it all optional methods. 
And if made required_methods then wallet need to support all those, else session can't be established.

ganchoradkov (approved):
Awesome 🔥

PR #4527, opened 2024-05-10
feat: rejection tags

ganchoradkov:
## Description

- Added new rejection tags according to spec https://github.com/WalletConnect/walletconnect-specs/pull/230
- Added try/catch around `onSessionAuthenticateRequest` that auto responds with an error if there is an exception

## Type of change

## How has this been tested?
tests 

## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -1851,6 +1851,8 @@ export class Engine extends IEngine {
         id: payload.id,
       });
     } catch (err: any) {
+      this.client.logger.error(err);
+
       const receiverPublicKey = payload.params.requester.publicKey;
       const senderPublicKey = await this.client.core.crypto.generateKeyPair();
 
@@ -1866,7 +1868,6 @@ export class Engine extends IEngine {
         encodeOpts,
         rpcOpts: ENGINE_RPC_OPTS.wc_sessionAuthenticate.autoReject,
       });
-      this.client.logger.error(err);
     }
   };
 
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -366,6 +366,7 @@ export class Engine extends IEngine {
         id,
         topic: pairingTopic,
         error: reason,
+        rpcOpts: ENGINE_RPC_OPTS.wc_sessionPropose.reject,
       });
       await this.client.proposal.delete(id, getSdkError("USER_DISCONNECTED"));
     }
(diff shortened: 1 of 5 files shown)


packages/sign-client/src/controllers/engine.ts
```diff
@@ -1434,6 +1436,7 @@ export class Engine extends IEngine {
         id,
         topic,
         error: err,
+        rpcOpts: ENGINE_RPC_OPTS.wc_sessionPropose.autoReject,
```
bkrem, line 1439: Is there any potential for breakage with pre-rejection-tags clients when receiving this new tag instead of the default  (https://github.com/WalletConnect/walletconnect-monorepo/pull/4527/files#diff-04e05e19b9270d1267b68d324e16997e4465d6879324fe4427a6194c00410eabL1234) that was used until now?

bkrem (approved):

packages/sign-client/src/controllers/engine.ts
ganchoradkov: nope, the tags are used internally and have no impact over `client <-> client` behaviour

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1829,23 +1832,42 @@ export class Engine extends IEngine {
...
+    } catch (err: any) {
+      const receiverPublicKey = payload.params.requester.publicKey;
+      const senderPublicKey = await this.client.core.crypto.generateKeyPair();
+
+      const encodeOpts = {
+        type: TYPE_1,
+        receiverPublicKey,
+        senderPublicKey,
+      };
+      await this.sendError({
```
Cali93, line 1864: Would it be worth to not await `sendError` ?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1829,23 +1832,42 @@ export class Engine extends IEngine {
...
+        id: payload.id,
+        topic,
+        error: err,
+        encodeOpts,
+        rpcOpts: ENGINE_RPC_OPTS.wc_sessionAuthenticate.autoReject,
+      });
+      this.client.logger.error(err);
```
Cali93: I'd suggest to move this line to be the first one line in the catch block. It should be easier to debug when something goes wrong.

packages/sign-client/test/sdk/auth.spec.ts
```diff
@@ -141,6 +143,8 @@ describe("Authenticated Sessions", () => {
     expect(wallet.auth.requests.getAll().length).to.eq(0);
     expect(dapp.proposal.getAll().length).to.eq(0);
     expect(dapp.auth.requests.getAll().length).to.eq(0);
+
+    await deleteClients({ A: dapp, B: wallet });
```
Cali93, line 147: Small nit: I think you might be able to use `afterAll`

Cali93 (approved):
LGTM 👌 
I just have small suggestions

jakubuid (approved):

packages/sign-client/src/controllers/engine.ts
ganchoradkov: imo having the await will allow the publish of the error response to complete so better have it

packages/sign-client/test/sdk/auth.spec.ts
ganchoradkov: lets leave for another PR

PR #4553, opened 2024-05-23
chore: pairing topic setter on session settle

ganchoradkov:
## Description
Reworked dapp side to set `pairingTopic` on its own instead of relying on being provided by session settle request

## Type of change

## How has this been tested?
existing tests 

## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -271,7 +271,6 @@ export class Engine extends IEngine {
     }
 
     const { pairingTopic, proposer, requiredNamespaces, optionalNamespaces } = proposal;
-    // pairingTopic = pairingTopic || "";
 
     const selfPublicKey = await this.client.core.crypto.generateKeyPair();
     const peerPublicKey = proposer.publicKey;
@@ -1512,7 +1511,7 @@ export class Engine extends IEngine {
         expiry,
         namespaces,
         acknowledged: true,
-        pairingTopic: "", // pairingTopic will be set in the `session_connect` listener
+        pairingTopic: "", // pairingTopic will be set in the `session_connect` handler
         requiredNamespaces: {},
         optionalNamespaces: {},
         controller: controller.publicKey,
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -200,6 +200,7 @@ export class Engine extends IEngine {
         metadata: this.client.metadata,
       },
       expiryTimestamp,
+      pairingTopic: topic,
       ...(sessionProperties && { sessionProperties }),
     };
     const {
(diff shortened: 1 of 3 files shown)


packages/sign-client/src/controllers/engine.ts
```diff
@@ -267,8 +270,7 @@ export class Engine extends IEngine {
       throw error;
     }
 
-    let { pairingTopic, proposer, requiredNamespaces, optionalNamespaces } = proposal;
-    pairingTopic = pairingTopic || "";
```
bkrem, line 271: Is this safe to remove in terms of backwards compat? Seems like existing (dapp) clients that don't do [this](https://github.com/WalletConnect/walletconnect-monorepo/pull/4553/files#diff-04e05e19b9270d1267b68d324e16997e4465d6879324fe4427a6194c00410eabR219) will have `pairingTopic: undefined` in the proposal they send, i.e. not a string?

packages/types/src/sign-client/jsonrpc.ts
```diff
@@ -44,7 +44,6 @@ export declare namespace JsonRpcTypes {
       namespaces: SessionTypes.Namespaces;
       sessionProperties?: ProposalTypes.SessionProperties;
       sessionConfig?: SessionTypes.SessionConfig;
-      pairingTopic: string;
```
bkrem, line 47: Do we risk breaking any externally used public interfaces by removing this?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -200,6 +200,7 @@ export class Engine extends IEngine {
         metadata: this.client.metadata,
       },
       expiryTimestamp,
+      pairingTopic: topic,
```
Cali93, line 203: How does this plays out when a client has an existing `pairingTopic` ?
Is it also handled on the dapp side ?

Cali93, line 203: How does this plays out when a client has an existing `pairingTopic` ?
Is it also handled on the dapp side ?

ganchoradkov: the dapp sets the pairing topic itself and does not rely on the wallet providing it as both peers know it ahead of time from the pairing uri.

packages/sign-client/src/controllers/engine.ts
ganchoradkov: clients are handling the case even at the moment because native wallets do not provide the pairing topic anyway

packages/types/src/sign-client/jsonrpc.ts
ganchoradkov: `wc_sessionSettle` is internal request so technically shouldn't be used directly by devs in any way

bkrem (approved):

PR #4560, opened 2024-05-24
fix: handle multiple event handlers for the same event

schmanu:
## Description
Currently when calling the Wallet Connect wallet's `on` function it registers the `onSessionProposal` handler multiple times in the `signClient`. 
This causes each event to be called as many times as there are event handlers registered for the event.

This fix always calls `off` before calling `on` when setting the `onSessionProposal` handler.

## Type of change

## How has this been tested?

I added a unit test which caught the issue.

## Checklist

The bug came up while working on https://github.com/safe-global/safe-wallet-web/pull/3656 and the fix is required to unblock it.


ganchoradkov (approved):
Many thanks for the investigation & fixing 💯 💯 💯 💯

PR #4638, opened 2024-06-21
fix: process incoming requests in order

ganchoradkov:
## Description
Updated intake logic of incoming relay messages so they are processed by Sign Client one by one instead of all at the same time. The old approach was resulting in conflicts when messages were dependent on each other such as `session_update` adding new chain to the session and `session_event`instructing the client to switch to that new chain. In the above case, the `session_event` must be processed after the `session_update` has finished the update, else the client has no knowledge of that new chain.

Additionally,  implemented `relayMessageCache` that stores any messages received while the client is still initializing and the messages are processed after init has finished. This allows implementing apps to setup all event listeners and not miss  events 

## Type of change

## How has this been tested?
dogfooding

## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -131,7 +131,7 @@ export class Engine extends IEngine {
   >();
 
   private recentlyDeletedLimit = 200;
-  private relayMessagesCache: RelayerTypes.MessageEvent[] = [];
+  private relayMessageCache: RelayerTypes.MessageEvent[] = [];
 
   constructor(client: IEngine["client"]) {
     super(client);
@@ -1032,12 +1032,10 @@ export class Engine extends IEngine {
   public processRelayMessageCache: IEngine["processRelayMessageCache"] = () => {
     // process the relay messages cache in the next tick to allow event listeners to be registered by the implementing app
     setTimeout(async () => {
-      console.log("Processing relay messages cache", this.relayMessagesCache.length);
-      if (this.relayMessagesCache.length === 0) return;
-      while (this.relayMessagesCache.length > 0) {
+      if (this.relayMessageCache.length === 0) return;
+      while (this.relayMessageCache.length > 0) {
         try {
-          const message = this.relayMessagesCache.shift();
-          console.log("Processing relay message", message);
+          const message = this.relayMessageCache.shift();
           if (message) {
             await this.onRelayMessage(message);
           }
@@ -1304,11 +1302,9 @@ export class Engine extends IEngine {
 
   private registerRelayerEvents() {
     this.client.core.relayer.on(RELAYER_EVENTS.message, (event: RelayerTypes.MessageEvent) => {
-      console.log("Relayer message received", this.initialized, this.relayMessagesCache.length > 0);
       // capture any messages that arrive before the client is initialized so we can process them after initialization is complete
-      if (!this.initialized || this.relayMessagesCache.length > 0) {
-        this.relayMessagesCache.push(event);
-        console.log("Relayer message cached", this.relayMessagesCache.length);
+      if (!this.initialized || this.relayMessageCache.length > 0) {
+        this.relayMessageCache.push(event);
       } else {
         this.onRelayMessage(event);
       }
@@ -1364,11 +1360,9 @@ export class Engine extends IEngine {
       if (!request) continue;
 
       try {
-        console.log("Processing request", request.payload.id);
         await this.processRequest(request);
         // small delay to allow for any async tasks to complete
         await new Promise((resolve) => setTimeout(resolve, 300));
-        console.log("Request processed", request.payload.id, "✅");
       } catch (error) {
         this.client.logger.warn(error);
       }
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -1336,9 +1336,11 @@ export class Engine extends IEngine {
       if (!request) continue;
 
       try {
-        this.processRequest(request);
+        console.log("Processing request", request.payload.id);
+        await this.processRequest(request);
         // small delay to allow for any async tasks to complete
         await new Promise((resolve) => setTimeout(resolve, 300));
+        console.log("Request processed", request.payload.id, "✅");
       } catch (error) {
         this.client.logger.warn(error);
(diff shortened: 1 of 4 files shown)


glitch-txs (approved):

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1028,6 +1029,23 @@ export class Engine extends IEngine {
     return formatMessage(request, iss);
   };
 
+  public processRelayMessageCache: IEngine["processRelayMessageCache"] = () => {
+    // process the relay messages cache in the next tick to allow event listeners to be registered by the implementing app
+    setTimeout(async () => {
```
Cali93, line 1034: Not sure if it's a good idea to mix setTimeout with an async operation.
Unless I misunderstood, I'd rather isolate the `setTimeout` and do something along the lines of 
```suggestion
   await sleep(50)
   if (this.relayMessageCache.length === 0) return;
   while (this.relayMessageCache.length > 0) {
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1283,36 +1301,42 @@ export class Engine extends IEngine {
...
-        const { publicKey } = this.client.auth.authKeys.keys.includes(AUTH_PUBLIC_KEY_NAME)
-          ? this.client.auth.authKeys.get(AUTH_PUBLIC_KEY_NAME)
-          : ({ responseTopic: undefined, publicKey: undefined } as any);
-
-        const payload = await this.client.core.crypto.decode(topic, message, {
-          receiverPublicKey: publicKey,
-        });
+    this.client.core.relayer.on(RELAYER_EVENTS.message, (event: RelayerTypes.MessageEvent) => {
+      // capture any messages that arrive before the client is initialized so we can process them after initialization is complete
+      if (!this.initialized || this.relayMessageCache.length > 0) {
```
Cali93, line 1306: Is it a cache or a queue ? Why are we pushing the events in the cache instead of processing them?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1283,36 +1301,42 @@ export class Engine extends IEngine {
...
+        this.relayMessageCache.push(event);
+      } else {
+        this.onRelayMessage(event);
```
Cali93, line 1309: The event handler was an async operation but now it's a fire and forget, is it intentional ? What are the potential implications of this change ?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1283,36 +1301,42 @@ export class Engine extends IEngine {
...
-        }
-      },
-    );
+  private async onRelayMessage(event: RelayerTypes.MessageEvent) {
+    const { topic, message } = event;
+
+    // Retrieve the public key (if defined) to decrypt possible `auth_request` response
+    const { publicKey } = this.client.auth.authKeys.keys.includes(AUTH_PUBLIC_KEY_NAME)
+      ? this.client.auth.authKeys.get(AUTH_PUBLIC_KEY_NAME)
+      : ({ responseTopic: undefined, publicKey: undefined } as any);
```
Cali93, line 1320:
```suggestion
    const publicKey = this.client.auth.authKeys.keys.includes(AUTH_PUBLIC_KEY_NAME)
      ? this.client.auth.authKeys.get(AUTH_PUBLIC_KEY_NAME).publicKey
      : undefined;
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1336,7 +1360,7 @@ export class Engine extends IEngine {
       if (!request) continue;
 
       try {
-        this.processRequest(request);
+        await this.processRequest(request);
         // small delay to allow for any async tasks to complete
         await new Promise((resolve) => setTimeout(resolve, 300));
```
Cali93: You can abstract this in a `sleep` util function

```ts
const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms));
// await sleep(300)
```

Cali93: Is the timeout still needed ? You're awaiting `processRequest` so it shouldn't be ?
```suggestion
        await this.processRequest(request);
```

Cali93 (commented):
Are all tests passing ? Should we add unhappy path cases ?

packages/sign-client/src/controllers/engine.ts
ignaciosantise: why do we need that small timeout?

ganchoradkov: the goal is to process the message after the finish initializing. This allows to setup listeners after the `client.init()`

packages/sign-client/src/controllers/engine.ts
ganchoradkov: we put all messages received during init, and process them after init has finished. Otherwise it can lead to missed events by the implementing app

packages/sign-client/src/controllers/engine.ts
ganchoradkov: await here is not required because even if we await the `onRelayMessage` it doesn't prevent `this.client.core.relayer.on(RELAYER_EVENTS.message` to emit multiple times so the result is the same

packages/sign-client/src/controllers/engine.ts
ganchoradkov: it has been this way for a long time so would avoid changing it in this PR, it was just copy/paste to a new line

ganchoradkov:
>Are all tests passing ? Should we add unhappy path cases ?

Not sure how to slow down the client and force the behaviour thus has been manually dogfooded. Will tinker @Cali93

packages/sign-client/src/controllers/engine.ts
ignaciosantise: yes, but why 50 and not other number? Is 50 enough in a RN environment?

ganchoradkov: it's arbitrary number. We can change it to some other value if you have suggestion

Cali93: Shouldn't we use a flag like `client.initialized` then ?

Cali93 (approved):
LGTM

PR #5091, opened 2024-06-28
fix: always use fallback verify server

chris13524:
## Description

`verify.walletconnect.org` is routable in all regions, so just use it directly instead of having fallback logic which increases latency.

## Type of change

## How has this been tested?

Not tested

## Checklist


arein (approved):
was the fallback not covered by unit tests?

chris13524:
> was the fallback not covered by unit tests?

It was not. We need to add Playwright here to unity test Verify

ganchoradkov (approved):
💯

PR #5104, opened 2024-06-29
fix(utils): remove dependency on query-string and dedupe uint8arrays

v1rtl:
## Description

This PR removes usage of query-string in utils and replaces them with a widely available URLSearchParams. It also dedupes download of uint8arrays (currently it pulls 3.1.0 and 3.1.1 etc)

Related: https://github.com/43081j/ecosystem-cleanup/issues/10

## Type of change

- [x] Chore (non-breaking change that addresses non-functional tasks, maintenance, or code quality improvements)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Draft PR (breaking/non-breaking change which needs more work for having a proper functionality _[Mark this PR as ready to review only when completely ready]_)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How has this been tested?

Through unit tests

## Checklist

- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules

PR #5112, opened 2024-06-30
chore: partially dedupe fetch implementations and use native fetch instead

### LINKED ISSUE #5102: Duplicate fetch dependencies across WalletConnect SDK
**Describe the bug**

WalletConnect dependency tree includes 5 different, but identical fetch implementations instead of using just one:

- node-fetch-native@1.6.4
- node-fetch@2.7.0
- isomorphic-unfetch@3.1.0
- ofetch@1.3.4
- unfetch@4.2.0

This bloats up the bundle size and polyfills existing fetch implementation that all major platforms already support. The only platform without fetch API is IE11, in which case people who need to support IE11 can set up the polyfill themselves through webpack or any other bundler they're using. And from what I know, WalletConnect doesn't work on IE11.

Node.js supports fetch since v18, which is 2 years old. React Native has [native fetch support](https://t.co/OHWRfr1SJZ). Deno and Bun natively implement fetch. All browsers except IE implement fetch since very long ago.

**SDK Version (if relevant)**
 - Client: TypeScript
 - Version [e.g. 22] 2.13.3

**To Reproduce**
1. `npm i @walletconnect/ethereum-provider`
2. Get 5 different fetch implementations downloaded

**Expected behavior**

Either only one fetch implementation to be used, or preferably use native one.

**Additional context**

Dependency graph can be inspected here: https://npmgraph.js.org/?q=%40walletconnect%2Fethereum-provider%40latest

v1rtl:
## Description

This PR removes a few duplicate fetch implementations from WC and keeps only the ones by subdependencies (since downstream projects have to remove them). Instead packages now would use native fetch. It also puts a node 18.x requirement for fetch.

## Type of change

## How has this been tested?

Unit tests.

## Fixes/Resolves (Optional)

Fixes #5102

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10462,14 +10462,6 @@
         "node": ">=0.10"
       }
     },
-    "node_modules/data-uri-to-buffer": {
-      "version": "4.0.1",
-      "dev": true,
-      "license": "MIT",
-      "engines": {
-        "node": ">= 12"
-      }
-    },
     "node_modules/dateformat": {
       "version": "3.0.3",
       "resolved": "https://registry.npmjs.org/dateformat/-/dateformat-3.0.3.tgz",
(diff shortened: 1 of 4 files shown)


ganchoradkov:
hey, thanks for the PR 💯  I will test and get back as the SDK has to work on every JS env and want to confirm removing fetch doesn't cause regressions

v1rtl:
@ganchoradkov thanks for taking time to look. If you have time please take a look at https://github.com/WalletConnect/walletconnect-monorepo/pull/5104 as well, it's another PR to reduce dep count

v1rtl:
also it shouldn't break anything because WalletConnect uses fixed version tags from what I understand, instead of any kind of aliases such as "latest", and also because as long as you don't upgrade nothing's gonna break. If you upgrade, it'll only break if you're using node older than 18 and you don't have a global polyfill set up.

ganchoradkov (approved):
LGTM, thank you for the improvement 💯

v1rtl:
@ganchoradkov thanks for approving! there's some unrelated test failing, what should I do about it?

v1rtl:
@ganchoradkov is there anything blocking the PR?

ganchoradkov:
thank you for the PR again and sorry for the delay!

PR #5141, opened 2024-07-02
feat: link mode

ignaciosantise:
## Description

- Enable link-mode for mobile dapps/wallets
- Reworked relayer to start wss connection only when required i.e. topics present in storage

~TODO: cleanup logs before merging~

## Type of change

## How has this been tested?

Tested in RN sample dapp and wallet

## Checklist


packages/utils/src/crypto.ts
```diff
@@ -78,6 +79,14 @@ export function encrypt(params: CryptoTypes.EncryptParams): string {
   return serialize({ type, sealed, iv, senderPublicKey });
 }
 
+export function encodeTypeTwo(message: string): string {
```
ganchoradkov:
```suggestion
export function encodeTypeTwoEnvelope(message: string): string {
```

packages/utils/src/crypto.ts
```diff
@@ -86,7 +95,15 @@ export function decrypt(params: CryptoTypes.DecryptParams): string {
   return toString(message, UTF8);
 }
 
+export function decodeTypeTwo(encoded: string): string {
```
ganchoradkov:
```suggestion
export function decodeTypeTwoEnvelope(encoded: string): string {
```

packages/types/src/sign-client/engine.ts
```diff
@@ -390,6 +396,12 @@ export abstract class IEngine {
 
   public abstract authenticate: (
     params: AuthTypes.SessionAuthenticateParams,
+    walletUniversalLink?: string,
+  ) => Promise<EngineTypes.SessionAuthenticateResponsePromise>;
+
+  public abstract authenticateLinkMode: (
+    params: AuthTypes.SessionAuthenticateParams,
+    walletUniversalLink: string,
```
ganchoradkov: these two methods seem the same except the name, why do we need both?

packages/core/src/core.ts
```diff
@@ -166,4 +183,23 @@ export class Core extends ICore {
       throw error;
     }
   }
+
+  private dispatchEnvelope = ({ url }: { url: string }) => {
+    if (!url || !url.includes("wc_ev") || !url.includes("topic")) return;
+
+    const getSearchParamFromURL = (_url: string, param: any) => {
```
ignaciosantise: @ganchoradkov should we move this function to utils? where?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -868,11 +942,15 @@ export class Engine extends IEngine {
     });
 
     return {
-      uri: connectionUri,
+      uri: linkModeUri ?? connectionUri,
```
ignaciosantise: @ganchoradkov what do you think of this? in link-mode i return the URL to redirect.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -834,26 +899,35 @@ export class Engine extends IEngine {
...
-          throwOnFailedPublish: true,
-          clientRpcId: fallbackId,
-        }),
-      ]);
+      if (walletUniversalLink) {
+        const payload = formatJsonRpcRequest("wc_sessionAuthenticate", request, id);
+        this.client.core.history.set(pairingTopic, payload);
+        const message = await this.client.core.crypto.encode("", payload, { type: TYPE_2 });
+        const encodedMessage = encodeURIComponent(message);
+        linkModeUri = `${walletUniversalLink}?wc_ev=${encodedMessage}&topic=${pairingTopic}`;
```
ignaciosantise: should i create a function to get the link mode url?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -886,6 +964,13 @@ export class Engine extends IEngine {
       throw new Error(`Could not find pending auth request with id ${id}`);
     }
 
+    let transportType: RelayerTypes.TransportType = "relay";
+    let appLink;
+    if (this.isLinkModeEnabled(pendingRequest.requester.metadata)) {
+      transportType = "link_mode";
+      appLink = pendingRequest.requester.metadata.redirect?.universal;
+    }
+
```
ganchoradkov: can we move these inside the `sendError, sendRequest` instead of setting this on every usage e.g. if link mode is enabled etc etc to be used unless disabled by flag, for example on `disconnect` requests

packages/utils/src/crypto.ts
```diff
@@ -78,6 +79,14 @@ export function encrypt(params: CryptoTypes.EncryptParams): string {
   return serialize({ type, sealed, iv, senderPublicKey });
 }
 
+export function encodeTypeTwo(message: string): string {
+  const type = encodeTypeByte(TYPE_2);
+  // TODO: serialize iv should be optional
+  const iv = randomBytes(IV_LENGTH);
```
ignaciosantise, line 100: @ganchoradkov `serialize` expects `iv` in its params, so im generating one but its not used. Should i make the param optional?

ganchoradkov (commented):
Looking lit! writing down my initial thoughts, will dig more 💯

packages/types/src/sign-client/engine.ts
ignaciosantise: we dont. But i'll need to move the logic that checks in localstorage for the wallet list that support link mode, that nowadays is in `sign-client/src/client/authenticate`

packages/core/src/core.ts
ganchoradkov: `getSearchParamFromURL` for sure

packages/sign-client/src/controllers/engine.ts
ganchoradkov: the return of this fx is used for QR code generation so will have to test how it behaves

packages/utils/src/crypto.ts
ganchoradkov: imo better leave a comment that this param is not used as it's required for other envelop types

packages/sign-client/src/controllers/engine.ts
ignaciosantise: Okey, you mean to add a disable flag param to sendError, sendRequest?

`sendRequest({...args, enableLinkMode: true})`

packages/sign-client/src/controllers/engine.ts
ignaciosantise: The QR code should work, as in that case we don't send a wallet universal link as param so it generates an URI.

But it will break the `Copy Link` button as it will copy the URL

packages/sign-client/src/controllers/engine.ts
ignaciosantise: or define an array of methods that don't support linkMode, and add a condition inside `sendError`, `sendRequest` and `sendResult`

packages/types/src/core/relayer.ts
```diff
@@ -119,4 +123,5 @@ export abstract class IRelayer extends IEvents {
   public abstract restartTransport(relayUrl?: string): Promise<void>;
   public abstract confirmOnlineStateOrThrow(): Promise<void>;
   public abstract handleBatchMessageEvents(messages: RelayerTypes.MessageEvent[]): Promise<void>;
+  public abstract onLinkMessageEvent(messageEvent: RelayerTypes.MessageEvent): Promise<void>;
```
quetool, line 126: Should this be under relayer given the fact that LinkMode is a relay-less connection? Should we "separate concerns"?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1318,18 +1488,19 @@ export class Engine extends IEngine {
 
     const payload = await this.client.core.crypto.decode(topic, message, {
       receiverPublicKey: publicKey,
+      encoding: transportType === "link-mode" ? BASE64URL : BASE64,
```
quetool: Isn't `onRelayMessage` only called on relay events? And given that, is this distinction even needed? @ignaciosantise @ganchoradkov @jakubuid @llbartekll

packages/types/src/core/relayer.ts
quetool: Should this be under "relayer" given the fact that LinkMode is a relay-less connection? Should we "separate concerns"? Cc @ignaciosantise @ganchoradkov @jakubuid @llbartekll

packages/sign-client/src/controllers/engine.ts
ignaciosantise: not anymore. Now `onRelayMessage` its also being called for link-mode messages. That's why I added `transportType: "relay" | "link-mode"` in `MessageEvent` type

packages/types/src/core/relayer.ts
ignaciosantise: The problem is that `onLinkMessageEvent` internally calls `recordMessageEvent`, which is defined in the relayer. All message logic is tied to the relayer controller. Agree that we could make a refactor in the future if it's worth it

packages/sign-client/src/controllers/engine.ts
```diff
@@ -884,6 +988,15 @@ export class Engine extends IEngine {
...
 
+    const transportType =
+      pendingRequest.transportType || this.isLinkModeEnabled(pendingRequest.requester.metadata)
+        ? "link-mode"
+        : "relay";
+
+    if (transportType === "relay") {
+      await this.confirmOnlineStateOrThrow();
+    }
+
```
quetool, line 1087: Shouldn't we change `encodeOpts` at line 1005 to TYPE_1 or TYPE_2 depending on the transport type?

ignaciosantise: not sure TBH, are the other SDKs using type_2 in this case?

ignaciosantise: i think its type_1 but for link_mode it will be encoded in `BASE64URL`

packages/core/src/controllers/relayer.ts
```diff
@@ -465,6 +470,22 @@ export class Relayer extends IRelayer {
...
   }
 
+  public async onLinkMessageEvent(
+    messageEvent: RelayerTypes.MessageEvent,
+    opts: { sessionExists: boolean },
+  ) {
+    const { topic } = messageEvent;
+
+    if (!opts.sessionExists) {
+      const expiry = Math.floor(Date.now() / 1000) + 604800;
```
quetool: Shouldn't it be 30 days for the pairings?

ganchoradkov (approved):
💯

packages/core/src/controllers/relayer.ts
ganchoradkov: updated to 5min since we're deprecating pairings reuse anyway

packages/core/src/controllers/relayer.ts
```diff
@@ -278,9 +285,11 @@ export class Relayer extends IRelayer {
       this.relayUrl = relayUrl;
       await this.transportDisconnect();
     }
+    // if (this.initialized) {
```
ignaciosantise: @ganchoradkov can we remove this commented lines?

packages/core/src/controllers/subscriber.ts
```diff
@@ -217,7 +219,14 @@ export class Subscriber extends ISubscriber {
     }
   }
 
-  private async rpcSubscribe(topic: string, relay: RelayerTypes.ProtocolOptions) {
+  private async rpcSubscribe(
+    topic: string,
+    relay: RelayerTypes.ProtocolOptions,
+    transportType: RelayerTypes.TransportType = "relay",
```
ignaciosantise:
```suggestion
    transportType: RelayerTypes.TransportType = TRANSPORT_TYPES.relay,
```

packages/core/src/controllers/relayer.ts
ignaciosantise:
```suggestion
```

packages/core/src/controllers/relayer.ts
```diff
@@ -278,9 +285,11 @@ export class Relayer extends IRelayer {
...
     // Always create new socket instance when trying to connect because if the socket was dropped due to `socket hang up` exception
     // It wont be able to reconnect
     await this.createProvider();
+    // }
```
ignaciosantise:
```suggestion
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -929,12 +1042,17 @@ export class Engine extends IEngine {
...
-      request: { ...request, verifyContext: {} as any },
+      request: {
+        ...request,
+        verifyContext: {} as any,
+      },
       pairingTopic,
+      transportType,
     });
 
+    // TODO: check if it's ok to return linkmode url
```
ignaciosantise:
```suggestion
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1395,9 +1570,17 @@ export class Engine extends IEngine {
...
       throw error;
     }
-    const opts = rpcOpts || ENGINE_RPC_OPTS[record.request.method].res;
-    // await is intentionally omitted to speed up performance
-    this.client.core.relayer.publish(topic, message, opts);
+
+    if (isLinkMode) {
+      const redirectURL = getLinkModeURL(appLink, topic, message);
+      await (global as any).Linking.openURL(redirectURL, this.client.name);
+      redirectURL;
```
ignaciosantise:
```suggestion
```

packages/types/src/sign-client/engine.ts
```diff
@@ -259,6 +264,7 @@ export interface EnginePrivate {
   onSessionProposeResponse(
     topic: string,
     payload: JsonRpcResult<JsonRpcTypes.Results["wc_sessionPropose"]> | JsonRpcError,
+    transportType?: RelayerTypes.MessageEvent["transportType"],
```
ignaciosantise, line 267: @ganchoradkov this won't be used, unless we give support in the future. Should i remove it?

packages/types/src/sign-client/engine.ts
```diff
@@ -309,6 +315,7 @@ export interface EnginePrivate {
   onSessionRequest(params: {
     topic: string;
     payload: JsonRpcRequest<JsonRpcTypes.RequestParams["wc_sessionRequest"]>;
+    transportType?: RelayerTypes.MessageEvent["transportType"];
```
ignaciosantise, line 318: same here

packages/types/src/sign-client/engine.ts
ignaciosantise: nevermind, you are using transportTytpe for the subscribe function

packages/types/src/sign-client/engine.ts
ganchoradkov: session requests can be sent over link mode or i'm mistaken?

ignaciosantise: you are right, i misread the function name

packages/core/src/controllers/crypto.ts
```diff
@@ -108,28 +112,39 @@ export class Crypto implements ICrypto {
...
     const { type, senderPublicKey } = params;
-    const result = encrypt({ type, symKey, message, senderPublicKey });
+    const result = encrypt({ type, symKey, message, senderPublicKey, encoding: opts?.encoding });
     return result;
   };
 
   public decode: ICrypto["decode"] = async (topic, encoded, opts) => {
     this.isInitialized();
+    // eslint-disable-next-line no-console
+    console.log("decode", { encoded: encoded.substring(0, 10), encoding: opts?.encoding });
```
Cali93: Still needed ? Shouldn't we use the logger if it is ?

packages/core/src/controllers/relayer.ts
```diff
@@ -118,15 +125,11 @@ export class Relayer extends IRelayer {
...
-      if (this.subscriber.topics.length === 0 && this.subscriber.pending.size === 0) {
-        this.logger.info(`No topics subscribed to after init, closing transport`);
-        await this.transportClose();
-        this.transportExplicitlyClosed = false;
-      }
-    }, RELAYER_TRANSPORT_CUTOFF);
+    // @ts-expect-error - .cached is private
+    if (this.subscriber.cached.length > 0) {
+      await this.transportOpen().catch((error) => this.logger.warn(error));
+    }
```
Cali93: This seems weird, do you also get the error if you use `try/catch`?

packages/core/src/controllers/subscriber.ts
```diff
@@ -228,13 +237,25 @@ export class Subscriber extends ISubscriber {
...
+      const subId = hashMessage(topic + this.clientId);
+      // in link mode, allow the app to update its network state (i.e. active airplane mode) with small delay before attempting to subscribe
+      if (transportType === TRANSPORT_TYPES.link_mode) {
+        setTimeout(() => {
+          if (this.relayer.connected || this.relayer.connecting) {
+            this.relayer.request(request).catch((e) => this.logger.warn(e));
+          }
+        }, toMiliseconds(ONE_SECOND));
+        return subId;
+      }
```
Cali93, line 249: Is this kind of delay really needed ? Wouldn't this add a 1 second delay in every case (even if airplane mode is off).

On another note we should clear the timeout.

packages/utils/src/crypto.ts
```diff
@@ -86,33 +89,60 @@ export function encrypt(params: CryptoTypes.EncryptParams): string {
...
+): string {
+  const type = encodeTypeByte(TYPE_2);
+  // iv is not used in type 2 envelopes
+  const iv = randomBytes(IV_LENGTH);
+  const sealed = fromString(message, UTF8);
+  return serialize({ type, sealed, iv, encoding });
 }
 
 export function decrypt(params: CryptoTypes.DecryptParams): string {
+  console.log("decrypt", { encoding: params?.encoding });
```
Cali93: Still needed ?
```suggestion
```

Cali93 (commented):
Left some minor comments but very good job 💪

packages/core/src/controllers/crypto.ts
ignaciosantise: we still need to clear al console.logs

packages/core/src/controllers/subscriber.ts
ganchoradkov: >Wouldn't this add a 1 second delay in every case (even if airplane mode is off).

with link mode, yes but it doesn't cause any issues as we need sometime for the `react-native netinfo` to register network changes when the app is back in foreground 

>On another note we should clear the timeout.

why? there isn't a case where the timeout has to be stopped mid execution

Cali93: > with link mode, yes but it doesn't cause any issues as we need sometime for the react-native netinfo to register network changes when the app is back in foreground

Ideally we should detect when this operation is needed instead of having the delay in every situation but I understand it's probably not trivial.

> why? there isn't a case where the timeout has to be stopped mid execution

It's just a good practice I've been used to because of React development.

From Chat GPT
> Using clearTimeout is not strictly necessary every time you use setTimeout, but there are specific situations where it is beneficial or required:
> 
> Preventing Unwanted Execution: If there's a possibility that the action scheduled with setTimeout should not occur in certain conditions (for example, if the user navigates away from a page or a component is unmounted), then you should clear the timeout to prevent the action from executing when it's no longer needed or relevant.
> 
> Memory Leaks: In React components or other dynamic UI environments, not clearing timeouts can lead to memory leaks. If a timeout is set on a component and the component is unmounted before the timeout fires, the callback function may still be executed, which can lead to errors or unintended behavior, especially if the callback modifies component state.
> 
> Resource Management: Timers hold onto resources until they execute. Clearing them when no longer needed frees up these resources, which is beneficial for performance, especially in environments with limited resources like mobile devices.
> 
> Control Over Asynchronous Actions: In complex applications where multiple timeouts might be set and their behavior could overlap or interfere with each other, managing timeouts with clearTimeout ensures that only the desired timers are running at any given time.

Cali93 (approved):
Nice one Nacho 💪

PR #4649, opened 2024-06-25
Experiment/session bundler url

devceline:
# Proof Of Concept
Do not merge

PR #5171, opened 2024-07-10
feat: fetch call status via bundler url

ganchoradkov:
## Description
Implemented functionality in `universal-provider` where if bundler name is provided, it will do a http call to get the status instead of sending the request to the wallet. If bundler name is not provided then the request is sent to the wallet 

based on https://github.com/WalletConnect/walletconnect-monorepo/pull/4649

## Type of change

## How has this been tested?
tests
## Checklist


providers/universal-provider/src/providers/eip155.ts
```diff
@@ -198,6 +201,41 @@ class Eip155Provider implements IProvider {
...
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify(
+        formatJsonRpcRequest("eth_getUserOperationReceipt", [args.request.params?.[0]]),
+      ),
+    });
+
+    return await response.json();
```
Cali93, line 246: Consider checking if the response is successful before parsing it as a json.

```suggestion
    if(!response.ok){
      throw new Error(`Failed to fetch user operation receipt - ${response.status}`)
    }
    return await response.json();
```

providers/universal-provider/src/providers/eip155.ts
```diff
@@ -45,6 +46,8 @@ class Eip155Provider implements IProvider {
         return parseInt(this.getDefaultChain()) as unknown as T;
       case "wallet_getCapabilities":
         return (await this.getCapabilities(args)) as unknown as T;
+      case "wallet_getCallsStatus":
+        return (await this.getCallStatus(args)) as unknown as T;
```
Cali93, line 50: Wouldn't this result in a unhandled promise rejection if the call throws ?

Cali93 (approved):
LGTM, left some minor suggestions.

arein (commented):
how was this tested?

arein (commented):
This doesn't work with strict content-security-policies where arbitrary Bundler URLs cannot be called. That's why we need the Wallet Service. I'd expect some mention around these problems somewhere

providers/universal-provider/src/providers/eip155.ts
ganchoradkov: no, it as http calls are try/catched. Only if the request can't be sent to thte wallet and/or is rejected by the wallet it would throw

PR #5182, opened 2024-07-12
feat: generic provider

ganchoradkov:
## Description
Added generic subProvider to `universal-provider` to be used when custom one doesn't exist

## Type of change

## How has this been tested?
tests

## Checklist

### DIFF
--- a/providers/universal-provider/src/UniversalProvider.ts
+++ b/providers/universal-provider/src/UniversalProvider.ts
@@ -20,6 +20,7 @@ import CardanoProvider from "./providers/cardano";
 import ElrondProvider from "./providers/elrond";
 import MultiversXProvider from "./providers/multiversx";
 import NearProvider from "./providers/near";
+import GenericProvider from "./providers/generic";
 
 import {
   IUniversalProvider,
@@ -231,7 +232,13 @@ export class UniversalProvider implements IUniversalProvider {
       // ignore without active session
       if (!this.session) return;
       const [namespace, chainId] = this.validateChain(chain);
-      this.getProvider(namespace).setDefaultChain(chainId, rpcUrl);
+      const provider = this.getProvider(namespace);
+      // @ts-expect-error
+      if (provider.name === "generic") {
+        provider.setDefaultChain(`${namespace}:${chainId}`, rpcUrl);
+      } else {
+        provider.setDefaultChain(chainId, rpcUrl);
+      }
     } catch (error) {
       // ignore the error if the fx is used prematurely before namespaces are set
       if (!/Please call connect/.test((error as Error).message)) throw error;
@@ -367,6 +374,10 @@ export class UniversalProvider implements IUniversalProvider {
             namespace: combinedNamespace,
           });
           break;
+        default:
+          this.rpcProviders.generic = new GenericProvider({
+            namespace: combinedNamespace,
+          });
       }
     });
   }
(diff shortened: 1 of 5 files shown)


providers/universal-provider/src/providers/generic.ts
```diff
@@ -0,0 +1,133 @@
...
+  IProvider,
+  RequestParams,
+  RpcProvidersMap,
+  SessionNamespace,
+  SubProviderOpts,
+} from "../types";
+import { getGlobal, getRpcUrl } from "../utils";
+import { parseChainId } from "@walletconnect/utils";
+
+// Old wallet connect provider for Elrond
```
ganchoradkov:
```suggestion
```
rm comment

providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -231,7 +232,13 @@ export class UniversalProvider implements IUniversalProvider {
       // ignore without active session
       if (!this.session) return;
       const [namespace, chainId] = this.validateChain(chain);
-      this.getProvider(namespace).setDefaultChain(chainId, rpcUrl);
+      const provider = this.getProvider(namespace);
+      // @ts-expect-error
```
Cali93, line 236: Why do we need an `@ts-expect-error` here ?

providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -231,7 +232,13 @@ export class UniversalProvider implements IUniversalProvider {
...
+      if (provider.name === "generic") {
```
Cali93: Why do we use CAIP2 format for the generic provider and jus the `chainId` in the other case ?
On another note, is this used elsewhere ? If so it might be better to use a constant.
```suggestion
      if (provider.name === GENERIC_PROVIDER_NAME) {
```

providers/universal-provider/src/providers/generic.ts
```diff
@@ -0,0 +1,132 @@
...
+      return [];
+    }
+
+    return [
+      ...new Set(
+        accounts
+          // get the accounts from the active chain
+          .filter((account) => account.split(":")[1] === this.chainId.toString())
+          // remove namespace & chainId from the string
+          .map((account) => account.split(":")[2]),
```
Cali93, line 93: Don't we have utils for this ? Eg: `getChainIdFromCaip10Account` etc

providers/universal-provider/src/providers/generic.ts
```diff
@@ -0,0 +1,132 @@
...
+  private createHttpProviders(): RpcProvidersMap {
+    const http = {};
+    this.namespace?.accounts?.forEach((account) => {
+      const chain = parseChainId(account);
+      http[`${chain.namespace}:${chain.reference}`] = this.createHttpProvider(account);
+    });
+    return http;
+  }
+
+  private getHttpProvider(chain: string): JsonRpcProvider {
```
Cali93, line 107: nit naming
```suggestion
  private getHttpProvider(caip2NetworkId: string): JsonRpcProvider {
```

Cali93 (approved):
LGTM left minor comments.
Would have been nice to get more context over why we need this and relevant docs.

providers/universal-provider/src/UniversalProvider.ts
ganchoradkov: we're accesing private prop. I can make it public so ts doesn't complain

providers/universal-provider/src/UniversalProvider.ts
ganchoradkov: >Why do we use CAIP2 format for the generic provider and jus the chainId in the other case ?

because custom providers set the `namespace` part of caip2 internally. This generic provider doesn't know any namespace and it actually supports multiple at the same time so it needs the whole caip2 as param unlike the others

PR #5188, opened 2024-07-16
feat: events sdk

ganchoradkov:
## Description
Implemented Events according [to spec ](https://www.notion.so/walletconnect/Events-SDK-tech-doc-910ed56d90834b79943a754fec013aac)

## Type of change

## How has this been tested?
dogfooding, tests 

## Checklist


packages/core/src/controllers/events.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    if (eventsToSend.length === 0) return;
+
+    try {
+      const response = await fetch(EVENTS_CLIENT_API_URL, {
+        method: "POST",
+        body: JSON.stringify(eventsToSend),
+        headers: {
+          "x-project-id": `${this.core.projectId}`,
```
chris13524: Isn't this going to introduce an `OPTIONS` request? Can we change the API to send these fields in the URL instead?

packages/core/src/controllers/events.ts
```diff
@@ -0,0 +1,200 @@
...
+  };
+
+  private setMethods = (eventId: string) => {
+    return {
+      addTrace: (trace: string) => this.addTrace(eventId, trace),
+      setError: (errorType: string) => this.setError(eventId, errorType),
+    };
+  };
+
+  private addTrace = async (eventId: string, trace: string) => {
```
chris13524: Why are `addTrace` and `setError` marked as `async`? This makes me think they might be long-running. I think we can remove this to guarantee they don't block the flows.

packages/core/src/controllers/events.ts
```diff
@@ -0,0 +1,200 @@
...
+    for (const [_, event] of this.events) {
+      if (event.props.type) {
+        eventsToSend.push(event);
+      }
+    }
+
+    if (eventsToSend.length === 0) return;
+
+    try {
+      const response = await fetch(EVENTS_CLIENT_API_URL, {
```
chris13524: [`priority`](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#priority) should be set to `low`

packages/core/src/controllers/events.ts
```diff
@@ -0,0 +1,200 @@
...
+import { HEARTBEAT_EVENTS } from "@walletconnect/heartbeat";
+import { fromMiliseconds } from "@walletconnect/time";
+
+export class EventClient extends IEventClient {
+  public readonly context = EVENTS_STORAGE_CONTEXT;
+  private readonly storagePrefix = CORE_STORAGE_PREFIX;
+  private readonly storageVersion = EVENTS_STORAGE_VERSION;
+
+  private events = new Map<string, EventClientTypes.Event>();
+  private toPersist = false;
```
Cali93: Does this determine if we should persist events or not ? How come we mutate this variable in functions such as `addTrace` and `setError` ?

nit boolean naming should favor `isTerm`, `hasTerm`, `shouldTerm`
```suggestion
  private hasPersistence = false;
```

packages/core/src/controllers/events.ts
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  public createEvent: IEventClient["createEvent"] = (params) => {
+    const {
+      event = "ERROR",
+      type = "",
+      properties: { topic, trace },
+    } = params;
+    const eventId = uuidv4();
+    const bundleId = this.core.projectId || "";
```
Cali93, line 49: `bundleId` is for mobile platfroms, in your case you should send the `domain` of the web app using the client.

packages/core/src/controllers/events.ts
Cali93: I don't understand why those are returning promises either.

packages/core/src/controllers/events.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+  };
+
+  private submit = async () => {
+    if (!this.telemetryEnabled) return;
+
+    if (this.events.size === 0) return;
+
+    const eventsToSend = [];
+    // exclude events without type as they can be considered `in progress`
```
Cali93, line 165: Could you clarify what you mean by "in progress" ?

packages/core/src/controllers/events.ts
Cali93: Correct, [PR](https://github.com/WalletConnect/pulse-analytics-api/pull/40) to accept query params is now live on prod so that can be used in favour of headers.

packages/core/src/controllers/events.ts
Cali93: You should use the API url depending on the environment, eg if you're developing it should be using the [staging API](https://github.com/WalletConnect/pulse-analytics-api?tab=readme-ov-file#staging).

packages/core/src/controllers/events.ts
```diff
@@ -0,0 +1,200 @@
...
+        headers: {
+          "x-project-id": `${this.core.projectId}`,
+          "x-sdk-type": "events_sdk",
+          "x-sdk-version": `js-${RELAYER_SDK_VERSION}`,
+        },
+      });
+      if (response.ok) {
+        for (const event of eventsToSend) {
+          this.events.delete(event.eventId);
+          this.toPersist = true;
```
Cali93: I'm worry of race conditions with this variable being mutated in various places.

packages/core/src/controllers/pairing.ts
```diff
@@ -111,15 +114,38 @@ export class Pairing implements IPairing {
...
+      properties: {
+        topic: params?.uri,
+        trace: [EVENT_CLIENT_PAIRING_TRACES.pairing_started],
+      },
+    });
+
+    try {
+      this.isValidPair(params, event);
+    } catch (error) {
+      event.setError(EVENT_CLIENT_PAIRING_ERRORS.malformed_pairing_uri);
```
Cali93: Wouldn't this end up in a duplicated event since `isValidPair` already takes care of setting the error event ?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1107,6 +1231,13 @@ export class Engine extends IEngine {
   };
 
   private deleteProposal: EnginePrivate["deleteProposal"] = async (id, expirerHasDeleted) => {
+    if (expirerHasDeleted) {
+      try {
+        const proposal = this.client.proposal.get(id);
+        const event = this.client.core.eventClient.getEvent({ topic: proposal.pairingTopic });
+        event?.setError(EVENT_CLIENT_SESSION_ERRORS.proposal_expired);
+      } catch (error) {}
```
Cali93, line 1233: Curious where could this throw ? Should we just add a simple log to eventually catch weird cases ?

Cali93 (commented):
Looking forward to the results that will come out of this.
Left some comments, good job so far 💪

packages/core/src/controllers/events.ts
ganchoradkov: we're recording traces for a potential error that might or might not occur. If it doesn't occur the recorded traces are deleted

packages/sign-client/src/controllers/engine.ts
ganchoradkov: `const proposal = this.client.proposal.get(id);`
the storage throws if the `id` is not found. Imo logging is not nessesary because its not actionable by users and doesn't interrupt the flow. The method tries to delete a proposal that is already deleted so no issues there

packages/core/src/controllers/events.ts
ganchoradkov: will note it to update at later point. The SDK doesn't have dev environment currently and is always in `prod`

packages/core/src/controllers/events.ts
ganchoradkov: I will leave it as is because we also cater to node environments where domain is not available

packages/core/src/controllers/events.ts
ganchoradkov: `eventsToSend` shouldn't mutate because it is declared before the http call https://github.com/WalletConnect/walletconnect-monorepo/blob/fa4d6c9ddd83fe79df2dc1093831ba5bd814ccd5/packages/core/src/controllers/events.ts#L170
and even if `this.events` mutate during that, it wont remove the new values

chris13524 (approved):
LGTM

PR #5194, opened 2024-07-17
refactor: reduce reconnect timeout

ganchoradkov:
## Description
- Reduced reconnect timeout to 0.1 seconds
- Removed await from `subscriber.start` start as it blocks the client initialization during the batch fetch & subscriptions

Both these changes would make the client perform and feel faster especially when the client has many persisted topics

## Type of change

## How has this been tested?
tests

benchmarked with 500 topics/no mailbox msgs - 
before(wait for subscribe): `1106 ms`
after(don't wait for subscribe): `155 ms`

## Checklist

### DIFF
--- a/packages/core/src/constants/relayer.ts
+++ b/packages/core/src/constants/relayer.ts
@@ -1,5 +1,3 @@
-import { ONE_SECOND } from "@walletconnect/time";
-
 export const RELAYER_DEFAULT_PROTOCOL = "irn";
 
 export const RELAYER_DEFAULT_LOGGER = "error";
@@ -28,7 +26,7 @@ export const RELAYER_PROVIDER_EVENTS = {
   error: "error",
 };
 
-export const RELAYER_RECONNECT_TIMEOUT = ONE_SECOND;
+export const RELAYER_RECONNECT_TIMEOUT = 0.1;
 
 export const RELAYER_STORAGE_OPTIONS = {
   database: ":memory:",
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -299,7 +299,7 @@ export class Relayer extends IRelayer {
           reject(e);
         });
         this.subscriber.start().catch((error) => {
-          this.logger.warn(error);
+          this.logger.error(error);
           this.onDisconnectHandler();
         });
         this.hasExperiencedNetworkDisruption = false;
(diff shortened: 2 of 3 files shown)


packages/core/src/controllers/relayer.ts
```diff
@@ -298,7 +298,10 @@ export class Relayer extends IRelayer {
         ).catch((e) => {
           reject(e);
         });
-        await this.subscriber.start();
+        this.subscriber.start().catch((error) => {
```
bkrem, line 301: Are there possible race conditions or downstream failures here if we resolve `transportOpen` but the subscriber fails to init for some reason?

I think this qualifies as `logger.error()` in the catch since the client will be in a broken state without the subscriber initialized.

packages/core/src/controllers/relayer.ts
```diff
@@ -474,11 +477,13 @@ export class Relayer extends IRelayer {
   };
 
   private onConnectHandler = () => {
+    this.logger.trace("wss connected");
```
bkrem: Concept of wss vs http as transport is abstracted in the relayer, would be good to avoid references to concrete transport layer:
```suggestion
    this.logger.trace("relayer connected");
```

packages/core/src/controllers/relayer.ts
```diff
@@ -474,11 +477,13 @@ export class Relayer extends IRelayer {
...
     this.startPingTimeout();
     this.events.emit(RELAYER_EVENTS.connect);
   };
 
   private onDisconnectHandler = () => {
+    this.logger.trace("wss disconnected");
```
```suggestion
    this.logger.trace("relayer disconnected");
```

bkrem (approved):

packages/core/src/controllers/relayer.ts
ganchoradkov: in bad scenario(subscriber failing), the only difference in the behaviour is that the client will initialize instead of staying in a pending promise while reconnect is in progress

packages/core/src/constants/relayer.ts
```diff
@@ -28,7 +26,7 @@ export const RELAYER_PROVIDER_EVENTS = {
   error: "error",
 };
 
-export const RELAYER_RECONNECT_TIMEOUT = ONE_SECOND;
+export const RELAYER_RECONNECT_TIMEOUT = 0.1;
```
chris13524, line 29: Can we please remove it entirely for the first reconnect or two? After the first one, it can back off and retry every second or 5s like Kotlin does

PR #5229, opened 2024-07-30
fix: avoid deeplinking if document isn't in focus

ganchoradkov:
## Description
Avoid deeplinking if document is not in focus. This resolves an issue where deeplinks initiated when app was in background would trigger when the app is back in focus leading to confusing behaviour

## Type of change

## How has this been tested?
canary `2.15.2-canary-9bbce7c`

## Checklist

### DIFF
--- a/packages/utils/src/misc.ts
+++ b/packages/utils/src/misc.ts
@@ -379,6 +379,11 @@ export async function handleDeeplinkRedirect({
     const env = getEnvironment();
 
     if (env === ENV_MAP.browser) {
+      if (!getDocument()?.hasFocus()) {
+        console.warn("Document does not have focus, skipping deeplink.");
+        return;
+      }
+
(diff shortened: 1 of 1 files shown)


arein (commented):
this is a bit more risky. Are we sure there's no edge cases where the browser is in background and then comes to the foreground shortly after?

packages/utils/src/misc.ts
```diff
@@ -379,6 +379,11 @@ export async function handleDeeplinkRedirect({
     const env = getEnvironment();
 
     if (env === ENV_MAP.browser) {
+      if (!getDocument()?.hasFocus()) {
+        console.warn("Document does not have focus, skipping deeplink.");
```
Cali93, line 383: I think no log or just `console.log` would be more appropriate since this is the expected behaviour.

Cali93 (approved):
Can we add unit tests for the function and cover this case ? 🙏

PR #5236, opened 2024-08-01
feat: verify v2

ganchoradkov:
## Description
Implemented verify v2 according to spec https://www.notion.so/walletconnect/Verify-Server-rewrite-tech-doc-Verify-API-V2-30e207c77f8447d8af37a186799b7c9e?pvs=4#fbc0ca6d67c444118599ba49d8b358f9

## Type of change

## How has this been tested?
https://github.com/WalletConnect/web-examples/pull/662
## Checklist

### DIFF
--- a/packages/core/src/constants/verify.ts
+++ b/packages/core/src/constants/verify.ts
@@ -3,5 +3,6 @@ export const VERIFY_CONTEXT = "verify-api";
 const VERIFY_SERVER_COM = "https://verify.walletconnect.com";
 const VERIFY_SERVER_ORG = "https://verify.walletconnect.org";
 export const VERIFY_SERVER = VERIFY_SERVER_ORG;
+export const VERIFY_SERVER_V2 = "https://verify.walletconnect.com/v2";
 
 export const TRUSTED_VERIFY_URLS = [VERIFY_SERVER_COM, VERIFY_SERVER_ORG];
--- a/packages/core/src/constants/verify.ts
+++ b/packages/core/src/constants/verify.ts
@@ -3,6 +3,6 @@ export const VERIFY_CONTEXT = "verify-api";
 const VERIFY_SERVER_COM = "https://verify.walletconnect.com";
 const VERIFY_SERVER_ORG = "https://verify.walletconnect.org";
 export const VERIFY_SERVER = VERIFY_SERVER_ORG;
-export const VERIFY_SERVER_V2 = "https://verify.walletconnect.com/v2";
+export const VERIFY_SERVER_V2 = "https://verify.walletconnect.org/v2";
 
 export const TRUSTED_VERIFY_URLS = [VERIFY_SERVER_COM, VERIFY_SERVER_ORG];
--- a/packages/core/src/constants/verify.ts
+++ b/packages/core/src/constants/verify.ts
@@ -3,6 +3,6 @@ export const VERIFY_CONTEXT = "verify-api";
 const VERIFY_SERVER_COM = "https://verify.walletconnect.com";
 const VERIFY_SERVER_ORG = "https://verify.walletconnect.org";
 export const VERIFY_SERVER = VERIFY_SERVER_ORG;
-export const VERIFY_SERVER_V2 = "https://verify.walletconnect.org/v2";
+export const VERIFY_SERVER_V2 = `${VERIFY_SERVER}/v2`;
 
 export const TRUSTED_VERIFY_URLS = [VERIFY_SERVER_COM, VERIFY_SERVER_ORG];
--- a/packages/core/src/controllers/verify.ts
+++ b/packages/core/src/controllers/verify.ts
@@ -1,50 +1,86 @@
+/* eslint-disable no-console */
 import { generateChildLogger, getLoggerContext, Logger } from "@walletconnect/logger";
 import { IVerify } from "@walletconnect/types";
-import { isBrowser, isNode, isReactNative } from "@walletconnect/utils";
+import {
+  getCryptoKeyFromKeyData,
+  isNode,
+  P256KeyDataType,
+  verifyP256Jwt,
+} from "@walletconnect/utils";
 import { FIVE_SECONDS, ONE_SECOND, toMiliseconds } from "@walletconnect/time";
 
 import { TRUSTED_VERIFY_URLS, VERIFY_CONTEXT, VERIFY_SERVER, VERIFY_SERVER_V2 } from "../constants";
+import { IKeyValueStorage } from "@walletconnect/keyvaluestorage";
 
+type jwk = {
+  publicKey: P256KeyDataType;
+  expiresAt: number;
+};
 export class Verify extends IVerify {
   public name = VERIFY_CONTEXT;
-  private verifyUrl: string;
-  private iframe?: HTMLIFrameElement;
-  private initialized = false;
   private abortController: AbortController;
   private isDevEnv;
-  // the queue is only used during the loading phase of the iframe to ensure all attestations are posted
-  private queue: string[] = [];
-  // flag to disable verify when the iframe fails to load on main & fallback urls.
-  // this means Verify API is not enabled for the current projectId and there's no point in trying to initialize it again.
-  private verifyDisabled = false;
   private verifyUrlV2 = VERIFY_SERVER_V2;
+  private publicKey?: jwk;
 
-  constructor(public projectId: string, public logger: Logger) {
-    super(projectId, logger);
+  constructor(public projectId: string, public logger: Logger, public store: IKeyValueStorage) {
+    super(projectId, logger, store);
     this.logger = generateChildLogger(logger, this.name);
-    this.verifyUrl = VERIFY_SERVER;
     this.abortController = new AbortController();
     this.isDevEnv = isNode() && process.env.IS_VITEST;
     console.log("Verify v2 init", this.verifyUrlV2);
+    this.init();
   }
+
+  get storeKey(): string {
+    return `verify:public:key`;
+  }
+
+  public init = async () => {
+    this.publicKey = await this.store.getItem(this.storeKey);
+    console.log("persistedKey", this.publicKey);
+    if (this.publicKey && toMiliseconds(this.publicKey?.expiresAt) < Date.now()) {
+      console.log("public key expired");
+      await this.removePublicKey();
+    }
+    if (this.publicKey) return;
+    const key = await this.fetchPublicKey();
+    console.log("public key", key);
+    await this.persistPublicKey(key);
+  };
 
   public register: IVerify["register"] = async (params) => {
     console.log("register", params);
     const { id, decryptedId } = params;
     const url = `${this.verifyUrlV2}/attestation?projectId=${this.projectId}`;
-    const response = await fetch(url, {
-      method: "POST",
-      body: JSON.stringify({ id, decryptedId }),
-      headers: {
-        origin: "https://8951-78-130-198-143.ngrok-free.app/",
-      },
-    });
-    const { srcdoc } = await response.json();
+    let src = "";
+    try {
+      const response = await fetch(url, {
+        method: "POST",
+        body: JSON.stringify({ id, decryptedId }),
+        headers: {
+          origin: "https://8951-78-130-198-143.ngrok-free.app/",
+        },
+      });
+      const { srcdoc } = await response.json();
+      src = srcdoc;
+    } catch (e) {
+      console.error("error", e);
+      return;
+    }
+    console.log("srcdoc", src);
 
-    console.log("srcdoc", srcdoc);
+    const abortTimeout = this.startAbortTimer(ONE_SECOND * 2);
     const attestatiatonJwt = await new Promise((resolve) => {
+      const abortListener = () => {
+        document.body.removeChild(iframe);
+        window.removeEventListener("message", listener);
+        this.abortController.signal.removeEventListener("abort", abortListener);
+      };
+
+      this.abortController.signal.addEventListener("abort", abortListener);
       const iframe = document.createElement("iframe");
-      iframe.srcdoc = srcdoc;
+      iframe.srcdoc = src;
       iframe.src = "https://verify.walletconnect.com";
       iframe.style.display = "none";
       const listener = (event: MessageEvent) => {
@@ -53,10 +89,12 @@ export class Verify extends IVerify {
         const data = JSON.parse(event.data);
         if (data.type === "verify_attestation") {
           // best-practice field
-          // window.removeEventListener("message", listener);
-          // document.body.removeChild(iframe);
+          clearInterval(abortTimeout);
+          window.removeEventListener("message", listener);
+          document.body.removeChild(iframe);
+          this.abortController.signal.removeEventListener("abort", abortListener);
           console.log("attestation", data.attestation);
-          resolve(data.attestation);
+          resolve(data.attestation === null ? "" : data.attestation);
         }
       };
       document.body.appendChild(iframe);
@@ -68,9 +106,35 @@ export class Verify extends IVerify {
 
   public resolve: IVerify["resolve"] = async (params) => {
     if (this.isDevEnv) return "";
+    const { attestationId, hash } = params;
 
+    console.log("resolve attestation", params);
+
+    if (attestationId === "") {
+      console.log("resolve: attestationId is empty string");
+      return;
+    }
+
+    if (attestationId) {
+      const data = await this.isValidJwtAttestation(attestationId);
+      console.log("resolve data", data);
+
+      if (data?.hasExpired) {
+        console.log("resolve: jwt attestation expired");
+        return;
+      }
+
+      if (data?.valid) {
+        return {
+          origin: data.payload.origin,
+          isScam: data.payload.isScam,
+        };
+      }
+    }
+    if (!hash) return;
+    console.log("resolve hash", hash);
     const verifyUrl = this.getVerifyUrl(params?.verifyUrl);
-    return this.fetchAttestation(params.attestationId, verifyUrl);
+    return this.fetchAttestation(hash, verifyUrl);
   };
 
   get context(): string {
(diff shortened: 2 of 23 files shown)


packages/core/src/constants/verify.ts
```diff
@@ -3,5 +3,6 @@ export const VERIFY_CONTEXT = "verify-api";
 const VERIFY_SERVER_COM = "https://verify.walletconnect.com";
 const VERIFY_SERVER_ORG = "https://verify.walletconnect.org";
 export const VERIFY_SERVER = VERIFY_SERVER_ORG;
+export const VERIFY_SERVER_V2 = "https://verify.walletconnect.com/v2";
```
chris13524:
```suggestion
export const VERIFY_SERVER_V2 = "https://verify.walletconnect.org/v2";
```

packages/core/src/controllers/verify.ts
```diff
@@ -1,68 +1,140 @@
...
+      const abortListener = () => {
+        document.body.removeChild(iframe);
+        window.removeEventListener("message", listener);
+        this.abortController.signal.removeEventListener("abort", abortListener);
+      };
+
+      this.abortController.signal.addEventListener("abort", abortListener);
+      const iframe = document.createElement("iframe");
+      iframe.srcdoc = src;
+      iframe.src = "https://verify.walletconnect.com";
```
chris13524: Shouldn't need this?

packages/core/src/controllers/verify.ts
```diff
@@ -1,68 +1,140 @@
...
-      this.verifyDisabled = true;
+  public init = async () => {
+    this.publicKey = await this.store.getItem(this.storeKey);
+    console.log("persistedKey", this.publicKey);
+    if (this.publicKey && toMiliseconds(this.publicKey?.expiresAt) < Date.now()) {
+      console.log("public key expired");
+      await this.removePublicKey();
     }
+    if (this.publicKey) return;
+    const key = await this.fetchPublicKey();
```
chris13524: Fetching the key should:
1. Not block SDK initialization, it should run in-parallel (looks like already the case)
2. Be awaited by the signature verification code, looks like right now there is a race condition where if `publicKey` hasn't resolved yet then it won't verify
3. Not run until there is an intention to use the public key. I.e. a pairing was just created on the wallet-side

packages/core/src/controllers/verify.ts
```diff
@@ -1,68 +1,140 @@
...
+      src = srcdoc;
+    } catch (e) {
+      console.error("error", e);
+      return;
     }
+    console.log("srcdoc", src);
+
+    const abortTimeout = this.startAbortTimer(ONE_SECOND * 2);
+    const attestatiatonJwt = await new Promise((resolve) => {
+      const abortListener = () => {
```
chris13524: Is this perhaps a simpler way to do this vs all these add and remove listener calls? https://x.com/mattpocockuk/status/1817958766756733278

packages/core/src/controllers/verify.ts
```diff
@@ -161,4 +167,46 @@ export class Verify extends IVerify {
...
+    await this.store.removeItem(this.storeKey);
+    this.publicKey = undefined;
+  };
+
+  private isValidJwtAttestation = async (attestation: string) => {
+    if (!this.publicKey) {
+      console.log("public key not found");
+      return;
+    }
+    const cryptoKey = await getCryptoKeyFromKeyData(this.publicKey.publicKey);
```
chris13524: If this fails verification it should request the latest key: https://www.notion.so/walletconnect/Verify-Server-rewrite-tech-doc-Verify-API-V2-30e207c77f8447d8af37a186799b7c9e?pvs=4#b681b45f31384899b46985bff30266b4

packages/core/src/controllers/verify.ts
```diff
@@ -1,68 +1,140 @@
...
   public resolve: IVerify["resolve"] = async (params) => {
     if (this.isDevEnv) return "";
+    const { attestationId, hash } = params;
+
+    console.log("resolve attestation", params);
+
+    if (attestationId === "") {
+      console.log("resolve: attestationId is empty string");
+      return;
+    }
```
chris13524: This can be removed as per this conversation: https://walletconnect.slack.com/archives/C03TFK9BSGJ/p1722442807055809

packages/core/src/controllers/verify.ts
```diff
@@ -1,68 +1,140 @@
...
+        if (!event.data) return;
+        const data = JSON.parse(event.data);
+        if (data.type === "verify_attestation") {
+          // best-practice field
+          clearInterval(abortTimeout);
+          window.removeEventListener("message", listener);
+          document.body.removeChild(iframe);
+          this.abortController.signal.removeEventListener("abort", abortListener);
+          console.log("attestation", data.attestation);
+          resolve(data.attestation === null ? "" : data.attestation);
```
chris13524: This mapping can be removed as per this conversation: https://walletconnect.slack.com/archives/C03TFK9BSGJ/p1722442807055809

packages/core/src/controllers/verify.ts
```diff
@@ -1,76 +1,120 @@
...
+    try {
+      const response = await fetch(url, {
+        method: "POST",
+        body: JSON.stringify({ id, decryptedId }),
+      });
+      const { srcdoc } = await response.json();
+      src = srcdoc;
+      this.logger.debug("srcdoc fetched", src);
+    } catch (e) {
+      return;
```
chris13524: An error here shouldn't be suppressed. While previously verify was error-prone, we should tend towards reliability now i.e. the message is verified or it's not sent at all.

packages/core/src/controllers/verify.ts
```diff
@@ -1,76 +1,120 @@
...
+      iframe.srcdoc = src;
+      iframe.style.display = "none";
+      const listener = (event: MessageEvent) => {
+        if (!event.data) return;
+        const data = JSON.parse(event.data);
+        if (data.type === "verify_attestation") {
+          clearInterval(abortTimeout);
+          document.body.removeChild(iframe);
+          this.abortController.signal.removeEventListener("abort", abortListener);
+          window.removeEventListener("message", listener);
```
chris13524, line 85: Should this also be inside `abortListener()`?

packages/core/src/controllers/verify.ts
```diff
@@ -1,76 +1,120 @@
...
+        signal: this.abortController.signal,
+      });
+      const iframe = document.createElement("iframe");
+      iframe.srcdoc = src;
+      iframe.style.display = "none";
+      const listener = (event: MessageEvent) => {
+        if (!event.data) return;
+        const data = JSON.parse(event.data);
+        if (data.type === "verify_attestation") {
+          clearInterval(abortTimeout);
```
chris13524, line 80: Is this still needed?

packages/core/src/controllers/verify.ts
```diff
@@ -1,76 +1,120 @@
...
+          document.body.removeChild(iframe);
```
chris13524, line 81: Could we call `abortListener()` in non-abort cases too, and that way we don't need to duplicate this cleanup code?

packages/core/src/controllers/verify.ts
```diff
@@ -1,76 +1,120 @@
...
+      src = srcdoc;
+      this.logger.debug("srcdoc fetched", src);
+    } catch (e) {
+      return;
     }
+    const document = getDocument() as Document;
+    const abortTimeout = this.startAbortTimer(ONE_SECOND * 2);
+    const attestatiatonJwt = await new Promise((resolve) => {
+      const abortListener = () => {
+        document.body.removeChild(iframe);
```
chris13524: Do we need to reject the promise here?

packages/core/src/controllers/verify.ts
```diff
@@ -161,4 +139,83 @@ export class Verify extends IVerify {
...
+  };
+
+  private removePublicKey = async () => {
+    this.logger.debug(`removing verify v2 public key from storage`);
+    await this.store.removeItem(this.storeKey);
+    this.publicKey = undefined;
+  };
+
+  private isValidJwtAttestation = async (attestation: string) => {
+    const key = await this.getPublicKey();
```
chris13524, line 194: Fetching the initial key when one doesn't exist yet should occur [at the same time as connecting to the relay](https://www.notion.so/walletconnect/Verify-Server-rewrite-tech-doc-Verify-API-V2-30e207c77f8447d8af37a186799b7c9e?pvs=4#64d9804559e74a988ab8540b5b508c6c), not after the message to verify is received. This avoids extra latency to lookup the key by doing it in-parallel.

packages/core/src/controllers/verify.ts
ganchoradkov: I will add a log just in case but we should not block sending requests or normal sdk function just because there was an exception there, additionally its not really actionable by users to fix it should there be an exception there

packages/core/src/controllers/verify.ts
ganchoradkov: started rejecting so we can diff between success/failure cases

packages/core/src/controllers/verify.ts
ganchoradkov: this was implemented according to your previous comment above

> 3\. Not run until there is an intention to use the public key. I.e. a pairing was just created on the wallet-side

will update

packages/core/src/constants/verify.ts
```diff
@@ -3,5 +3,6 @@ export const VERIFY_CONTEXT = "verify-api";
 const VERIFY_SERVER_COM = "https://verify.walletconnect.com";
 const VERIFY_SERVER_ORG = "https://verify.walletconnect.org";
 export const VERIFY_SERVER = VERIFY_SERVER_ORG;
+export const VERIFY_SERVER_V2 = "https://verify.walletconnect.org/v2";
```
Cali93: nit 
```suggestion
export const VERIFY_SERVER_V2 = `${VERIFY_SERVER_ORG}/v2`;
```

packages/core/src/controllers/verify.ts
```diff
@@ -1,76 +1,130 @@
...
+  verifyP256Jwt,
+} from "@walletconnect/utils";
 import { FIVE_SECONDS, ONE_SECOND, toMiliseconds } from "@walletconnect/time";
+import { getDocument } from "@walletconnect/window-getters";
 
-import { TRUSTED_VERIFY_URLS, VERIFY_CONTEXT, VERIFY_SERVER } from "../constants";
+import { TRUSTED_VERIFY_URLS, VERIFY_CONTEXT, VERIFY_SERVER, VERIFY_SERVER_V2 } from "../constants";
+import { IKeyValueStorage } from "@walletconnect/keyvaluestorage";
 
+type jwk = {
```
Cali93: nit type should be uppercased
```suggestion
type Jwk = {
```

packages/core/src/controllers/verify.ts
```diff
@@ -1,76 +1,130 @@
...
-  // the queue is only used during the loading phase of the iframe to ensure all attestations are posted
-  private queue: string[] = [];
-  // flag to disable verify when the iframe fails to load on main & fallback urls.
-  // this means Verify API is not enabled for the current projectId and there's no point in trying to initialize it again.
-  private verifyDisabled = false;
-
-  constructor(public projectId: string, public logger: Logger) {
-    super(projectId, logger);
+  private verifyUrlV2 = VERIFY_SERVER_V2;
+  private publicKey?: jwk;
```
Cali93:
```suggestion
  private jwk?: Jwk;
```

packages/core/src/controllers/verify.ts
```diff
@@ -1,76 +1,130 @@
...
+      this.logger.warn(e);
+      return;
     }
+    try {
+      const document = getDocument() as Document;
+      const abortTimeout = this.startAbortTimer(ONE_SECOND * 3);
+      const attestatiatonJwt = await new Promise((resolve) => {
+        const abortListener = () => {
+          window.removeEventListener("message", listener);
+          document.body.removeChild(iframe);
```
Cali93, line 86: `iframe` isn't defined yet at this point ?

packages/core/src/controllers/verify.ts
```diff
@@ -1,76 +1,130 @@
...
+      src = srcdoc;
+      this.logger.debug("srcdoc fetched", src);
+    } catch (e) {
+      this.logger.warn(e);
+      return;
     }
+    try {
+      const document = getDocument() as Document;
+      const abortTimeout = this.startAbortTimer(ONE_SECOND * 3);
+      const attestatiatonJwt = await new Promise((resolve) => {
```
ignaciosantise: `attestationJwt` ?

ganchoradkov: lmao

packages/core/src/controllers/verify.ts
```diff
@@ -1,76 +1,130 @@
...
+        iframe.style.display = "none";
+        const listener = (event: MessageEvent) => {
+          if (!event.data) return;
+          const data = JSON.parse(event.data);
+          if (data.type === "verify_attestation") {
+            clearInterval(abortTimeout);
+            document.body.removeChild(iframe);
+            this.abortController.signal.removeEventListener("abort", abortListener);
+            window.removeEventListener("message", listener);
+            resolve(data.attestation === null ? "" : data.attestation);
```
ignaciosantise, line 103: which are the possible values of `data.attestation`? 

would `resolve(data.attestation || '');` work?

ganchoradkov: its the intended handling. if attestation is `null` it means that verify api is disabled which is different case then undefined

chris13524: I think `resolve(data.attestation ?? '');` would work then but personally I'd prefer to be a little more explicit here

packages/core/src/controllers/verify.ts
chris13524: It's the same type of error if publishing to the relay fails: the network is non-functional. The user's action is to retry when they've reconnected to the internet.

packages/core/src/controllers/verify.ts
chris13524: My request is a bit nuanced:
- Not block SDK startup
- Not run on SDK startup
- Not block attestation validation
- Only run when a pairing is created (i.e. it will be used)
- **Run concurrently with relay connection when a pairing is created**

So I'd see it going like this:
- Pairing is requested
- Relay connection is established
- (concurrently) Public key is obtained if it currently is not fresh

Then when verification takes place, it should likely already have the public key. And no bandwidth is wasted on getting the public key when it's not going to be used by a pairing.

chris13524 (approved):

chris13524:
There is a race condition that I'd like to see resolved, if not before merging then let's create an issue for it.

The logic is if there isn't a saved public key in the verification step then it will request the public key in a blocking manner. In the happy case the public key will already be requested and stored during SDK initialization, but if the network request isn't complete when the verification step comes, then it will result in a second request and additional latency.

The race condition can be avoided by awaiting the original request in the verification step.

```
[ key fetch in SDK init      ]
  [receive paring req ][ key fetch again ]
```

packages/core/src/controllers/verify.ts
```diff
@@ -1,76 +1,130 @@
...
+        };
+        document.body.appendChild(iframe);
+        window.addEventListener("message", listener, { signal: this.abortController.signal });
+      });
+      this.logger.debug("jwt attestation", attestatiatonJwt);
+      return attestatiatonJwt as string;
+    } catch (e) {
+      this.logger.warn(e);
+    }
+    return "";
```
Cali93, line 114: Why are we retruning an empty string ?

packages/core/src/controllers/verify.ts
```diff
@@ -1,76 +1,130 @@
...
+    if (attestationId === "") {
+      this.logger.debug("resolve: attestationId is empty, skipping");
+      return;
+    }
+
+    if (attestationId) {
+      const validation = await this.isValidJwtAttestation(attestationId);
+      if (validation) return validation;
+    }
+    if (!hash) return;
```
Cali93, line 130: Shouldn't we make this check before checking the validity ?

packages/core/src/controllers/verify.ts
```diff
@@ -80,85 +134,108 @@ export class Verify extends IVerify {
...
+    try {
+      if (newKey) {
+        const validation = await this.validateAttestation(attestation, newKey);
+        return validation;
+      }
+    } catch (e) {
+      this.logger.warn(e);
+      this.logger.warn("error validating attestation");
+    }
+    return undefined;
```
Cali93: This could be simplified since `getPublicKey` already calls `fetchAndPersistPublicKey`
```suggestion
    const key = await this.getPublicKey();
    try {
      const validation = await this.validateAttestation(attestation, key);
      return validation;
    } catch (e) {
      this.logger.warn(e);
      this.logger.warn("error validating attestation");
    }
   
    return undefined;
```

packages/core/src/controllers/verify.ts
```diff
@@ -80,85 +134,108 @@ export class Verify extends IVerify {
...
-      );
-      url = VERIFY_SERVER;
+  private validateAttestation = async (attestation: string, key: jwk) => {
+    const cryptoKey = await getCryptoKeyFromKeyData(key?.publicKey);
+    const result = await verifyP256Jwt<{
+      exp: number;
+      id: string;
+      origin: string;
+      isScam: boolean;
+    }>(attestation, cryptoKey);
```
Cali93: not super readable, can we extract the type ?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1462,8 +1467,11 @@ export class Engine extends IEngine {
         params.expiryTimestamp || calcExpiry(ENGINE_RPC_OPTS.wc_sessionPropose.req.ttl);
       const proposal = { id, pairingTopic: topic, expiryTimestamp, ...params };
       await this.setProposal(id, proposal);
-      const hash = hashMessage(JSON.stringify(payload));
-      const verifyContext = await this.getVerifyContext(hash, proposal.proposer.metadata);
+      const verifyContext = await this.getVerifyContext({
+        attestationId: attestation,
+        hash: hashMessage(JSON.stringify(payload)),
```
Cali93, line 1472: Nit I'd consider running some validation against the payload.

packages/utils/test/crypto.spec.ts
```diff
@@ -108,4 +112,28 @@ describe("Crypto", () => {
   it("calls generateRandomBytes32", () => {
     expect(generateRandomBytes32()).toBeTruthy();
   });
+  it("should validate verify v2 jwt", async () => {
```
Cali93, line 117: Would love to see more test cases here, specially for unhappy paths and also for the timeout logic.

Cali93 (commented):
Looks solid 💪 
A few more tests and it should be good to go

packages/core/src/controllers/verify.ts
chris13524: The current design is intentional in order to refresh the key if a different signing key is provided on the JWT. Perhaps a comment explaining that is needed

https://www.notion.so/walletconnect/Verify-Server-rewrite-tech-doc-Verify-API-V2-30e207c77f8447d8af37a186799b7c9e?pvs=4#7cffe1a834c944e6943f774d934fc925

packages/core/src/controllers/verify.ts
```diff
@@ -1,76 +1,130 @@
...
   get context(): string {
     return getLoggerContext(this.logger);
   }
 
   private fetchAttestation = async (attestationId: string, url: string) => {
-    this.logger.info(`resolving attestation: ${attestationId} from url: ${url}`);
+    this.logger.debug(`resolving attestation: ${attestationId} from url: ${url}`);
     // set artificial timeout to prevent hanging
     const timeout = this.startAbortTimer(ONE_SECOND * 5);
     const result = await fetch(`${url}/attestation/${attestationId}`, {
```
chris13524: Missing the new `v2Supported=true` flag

packages/core/src/controllers/verify.ts
```diff
@@ -214,28 +208,29 @@ export class Verify extends IVerify {
...
       isScam: boolean;
-    }>(attestation, cryptoKey);
+      isVerified: boolean;
+    }>(attestation, key.publicKey);
+    console.log("decoded result", result);
     const validation = {
-      valid: result.verified,
-      hasExpired: toMiliseconds(result.payload.payload.exp) < Date.now(),
-      payload: result.payload.payload,
+      valid: true,
```
chris13524: I assume this needs to be reverted

packages/sign-client/src/controllers/engine.ts
ganchoradkov: the payload is already validated at the top of the method

packages/core/src/controllers/verify.ts
ganchoradkov: js is correctly scoping it

packages/core/src/controllers/verify.ts
ganchoradkov: nope, because `hash` is used for verify v1, while verify v2 uses the `attestation` param. Since we first try with v2, `hash` is used only if `attestation` isn't available

Cali93: Got it, thanks for the context. Would be nice to have a comment about this.

packages/core/src/controllers/verify.ts
Cali93: Hmm I understand the subtle difference now. So if the first case fails we explicitly request a new key.

Cali93 (approved):
LFG 🚀

PR #5241, opened 2024-08-04
utils: switch deps from stablelib to noble

paulmillr:
This has been discussed in https://github.com/orgs/WalletConnect/discussions/4035.

noble is:

- audited (more audits coming soon)
- faster
- smaller
- used by other libraries (ethers, metamask), which can make them reuse same noble deps
- popular, issues will be discovered faster

Second part of the change: https://github.com/WalletConnect/walletconnect-utils/pull/185

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -4,5 +4,5 @@
     "packages/*",
     "providers/*"
   ],
-  "version": "2.15.2"
+  "version": "2.15.3"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -27515,7 +27515,7 @@
     },
     "packages/core": {
       "name": "@walletconnect/core",
-      "version": "2.15.2",
+      "version": "2.15.3",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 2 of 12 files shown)


chris13524:
Thanks for the PR!

Can we do a test to ensure that the keys are definitely compatible between these two libraries so we don't break any existing sessions when the library is upgraded?

paulmillr:
Of course, but I will need your cooperation here with regards to these compilation errors.

It happens because the build tooling is too old.

paulmillr:
@chris13524 can you clarify how you want the tests to look like? Should I manually create a key with old lib and put it into new impl?

paulmillr:
@chris13524 ping. Can you tell what to do here:

a. Create static JSON vectors from old implementation, commit them to repo, test with new code
b. Add dependency on old implementation and cross-test random vectors

chris13524:
@paulmillr I think of you just do a manual test showing they are interoperable that's fine. Input/output of these functions would be good, don't need to test the whole app

paulmillr:
Continuing in #5333

PR #5259, opened 2024-08-09
refactor: session settle response 

ganchoradkov:
## Description
Moved session settle response publish after the `session_connect` event is emitted. This update would reduce the connect latency by full `publish+relay ack` ms and improvement would be greater the slower the network is

context https://walletconnect.slack.com/archives/C04DB2EAHE3/p1723135790360629 

## Type of change

## How has this been tested?
n/a
## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -1555,18 +1555,18 @@ export class Engine extends IEngine {
         ...(sessionProperties && { sessionProperties }),
         ...(sessionConfig && { sessionConfig }),
       };
-      await this.sendResult<"wc_sessionSettle">({
-        id: payload.id,
-        topic,
-        result: true,
-        throwOnFailedPublish: true,
-      });
       const target = engineEvent("session_connect");
       const listeners = this.events.listenerCount(target);
       if (listeners === 0) {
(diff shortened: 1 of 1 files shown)


packages/sign-client/src/controllers/engine.ts
```diff
@@ -1555,18 +1555,18 @@ export class Engine extends IEngine {
...
-        result: true,
-        throwOnFailedPublish: true,
-      });
       const target = engineEvent("session_connect");
       const listeners = this.events.listenerCount(target);
       if (listeners === 0) {
         throw new Error(`emitting ${target} without any listeners 997`);
       }
       this.events.emit(engineEvent("session_connect"), { session });
+      await this.sendResult<"wc_sessionSettle">({
```
chris13524, line 1564: This is technically still blocking the method with the `await` here. Do you think this is of concern (i.e. remove the `await`)? Or would it not block the UX flow?

chris13524 (approved):

packages/sign-client/src/controllers/engine.ts
ganchoradkov: it makes no ux difference whether the method awaits the result or not because the event is emitted before that and it triggers the session completion

PR #5285, opened 2024-08-19
fix: Verify V3

chris13524:
## Description

Upgrades the Verify attestation style as per the [v3 spec](https://www.notion.so/walletconnect/Verify-V3-tech-doc-ca95f7df4415422e875fb9b4e5863920?pvs=4).

Remaining work:
- [x] Fix bug in the iframe load timeout
- [x] Merge https://github.com/WalletConnect/walletconnect-monorepo/pull/5288#pullrequestreview-2255230087
- [x] Rebase on revert commit
- [x] Merge https://github.com/WalletConnect/cloud-app/pull/1338

## Type of change

## How has this been tested?

Existing automated tests

## Checklist

### DIFF
--- a/packages/core/src/controllers/verify.ts
+++ b/packages/core/src/controllers/verify.ts
@@ -69,7 +69,7 @@ export class Verify extends IVerify {
     try {
       const document = getDocument() as Document;
       const abortTimeout = this.startAbortTimer(ONE_SECOND * 5);
-      const attestationJwt = await new Promise(async (resolve, reject) => {
+      const attestationJwt = await new Promise((resolve, reject) => {
         const abortListener = () => {
           window.removeEventListener("message", listener);
           document.body.removeChild(iframe);
--- a/packages/core/src/controllers/verify.ts
+++ b/packages/core/src/controllers/verify.ts
@@ -69,11 +69,11 @@ export class Verify extends IVerify {
     try {
       const document = getDocument() as Document;
       const abortTimeout = this.startAbortTimer(ONE_SECOND * 5);
-      const attestationJwt = await new Promise((resolve) => {
+      const attestationJwt = await new Promise(async (resolve, reject) => {
         const abortListener = () => {
           window.removeEventListener("message", listener);
           document.body.removeChild(iframe);
-          throw new Error("attestation aborted");
+          reject("attestation aborted");
         };
         this.abortController.signal.addEventListener("abort", abortListener);
         const iframe = document.createElement("iframe");
--- a/packages/core/src/controllers/verify.ts
+++ b/packages/core/src/controllers/verify.ts
@@ -79,6 +79,7 @@ export class Verify extends IVerify {
         const iframe = document.createElement("iframe");
         iframe.src = src;
         iframe.style.display = "none";
+        iframe.addEventListener("error", abortListener, { signal: this.abortController.signal });
         const listener = (event: MessageEvent) => {
           if (!event.data) return;
           const data = JSON.parse(event.data);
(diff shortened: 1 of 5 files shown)


packages/core/src/controllers/verify.ts
```diff
@@ -62,35 +62,21 @@ export class Verify extends IVerify {
...
-      this.logger.debug("srcdoc fetched", src);
-    } catch (e) {
-      this.logger.warn(e);
-      return;
-    }
+    const src = `${this.verifyUrlV3}/attestation?projectId=${this.core.projectId}&origin=${origin}&id=${id}&decryptedId=${decryptedId}`;
     try {
       const document = getDocument() as Document;
-      const abortTimeout = this.startAbortTimer(ONE_SECOND * 3);
+      const abortTimeout = this.startAbortTimer(ONE_SECOND * 5);
```
ganchoradkov, line 71:
```suggestion
      const abortTimeout = this.startAbortTimer(ONE_SECOND * 5);
```
this would block the publish for 5s in worst case scenario where verify iframe doesn't load, just noting it

ganchoradkov (approved):
💯

packages/core/src/controllers/verify.ts
chris13524: I almost think it should be higher, especially in bad network conditions. Perhaps we should add an onerror handler to the iframe?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1470,6 +1477,7 @@ export class Engine extends IEngine {
       const verifyContext = await this.getVerifyContext({
         attestationId: attestation,
         hash: hashMessage(JSON.stringify(payload)),
+        encryptedId,
```
chris13524, line 1480: @ganchoradkov isn't one of these (`attestationId`, `hash`, `encryptedId`) redundant with the other?

packages/core/src/controllers/verify.ts
```diff
@@ -102,16 +106,18 @@ export class Verify extends IVerify {
...
     }
 
     if (attestationId) {
-      const validation = await this.isValidJwtAttestation(attestationId);
-      if (validation) return validation;
+      const decoded = decodeJWT(attestationId) as unknown as { payload: JwkPayload };
+      if (decoded.payload.id === encryptedId) {
+        const validation = await this.isValidJwtAttestation(attestationId);
+        if (validation) return validation;
+      }
```
chris13524: @ganchoradkov 

If JWT is invalid this should result in UNKNOWN not an attempt to verify with v1

```suggestion
      } else {
        return;
      }
```

packages/sign-client/src/controllers/engine.ts
ganchoradkov: nope because 
- `attestationId` = the jwt
- `hash` = the unecnrypted payload used for verify v1
- `encryptedId` = hash of the encrypted message used for verify v2

`encryptedId` is used for the check against `jwt.payload.id`

packages/core/src/controllers/verify.ts
```diff
@@ -62,40 +63,29 @@ export class Verify extends IVerify {
...
+    const src = `${this.verifyUrlV3}/attestation?projectId=${this.core.projectId}&origin=${origin}&id=${id}&decryptedId=${decryptedId}`;
     try {
       const document = getDocument() as Document;
-      const abortTimeout = this.startAbortTimer(ONE_SECOND * 3);
+      const abortTimeout = this.startAbortTimer(ONE_SECOND * 5);
       const attestationJwt = await new Promise((resolve) => {
         const abortListener = () => {
           window.removeEventListener("message", listener);
           document.body.removeChild(iframe);
           throw new Error("attestation aborted");
```
chris13524: @ganchoradkov don't we need to `reject()` the promise here not just throw?

packages/core/src/controllers/verify.ts
chris13524: Added `onerror` handler in 47e3471d815aab083bd81e0ea4c75a0500e776ff

packages/core/src/controllers/verify.ts
ganchoradkov: ~The behaviour is the same whether is there is an error or the promise is rejected, just semantics~
tested and i'm wrong, updated with `reject` 👍

PR #5288, opened 2024-08-20
Revert "chore: reverts verify v2"

ganchoradkov:
Reverts WalletConnect/walletconnect-monorepo#5286

## testing canary
`2.15.1-canary.0`
`import { randomBytes } from '@stablelib/random';` <- randomBytes is imported correctly as named import
https://www.npmjs.com/package/@walletconnect/utils/v/2.15.1-canary.0?activeTab=code

Problematic build 
`import Bo,{randomBytes as Ei}from"@stablelib/random";` <- stablelib/random doesn't have a default export thus causing an exception
https://www.npmjs.com/package/@walletconnect/utils/v/2.15.0?activeTab=code


chris13524 (approved):

packages/core/src/controllers/verify.ts
```diff
@@ -1,156 +1,149 @@
...
     }
   };
 
   public register: IVerify["register"] = async (params) => {
-    if (!this.initialized) {
-      this.addToQueue(params.attestationId);
-      await this.init();
-    } else {
-      this.sendPost(params.attestationId);
+    if (!isBrowser()) return;
```
Cali93, line 65: Just curious but assuming a NextJS app just uses RSC and SSR then Verify would never be running, right ?

Cali93 (commented):
LGTM

PR #5340, opened 2024-09-08
fix: unchecked access to `process` global variable

### LINKED ISSUE #5339: ReferenceError: process is not defined
This error appeared after `@walletconnect/universal-provider` v2.16.0

This was caused by unchecked access to global variable `process`:

```js
if ((e2 = process == null ? void 0 : process.env) != null && e2.IS_VITEST || rr() && (t = this.client.metadata.redirect) != null && t.linkMode)
```

tien:
Fixes #5339

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -2914,7 +2914,10 @@ export class Engine extends IEngine {
   };
 
   private registerLinkModeListeners = async () => {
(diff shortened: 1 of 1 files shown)


ganchoradkov (approved):
thank you for patching! 💯

tien:
Thanks, @ganchoradkov 🙏

Can this be released soon as part of a hot fix? As it's a fatal error.

PR #5343, opened 2024-09-09
feat: init event

ganchoradkov:
## Description
context: https://www.notion.so/walletconnect/Walletkit-events-sdk-init-event-08b51439dd8b42c5bbe993c63d2ce54e

## Type of change

## How has this been tested?
tests
## Checklist

### DIFF
--- a/packages/core/src/controllers/events.ts
+++ b/packages/core/src/controllers/events.ts
@@ -1,6 +1,6 @@
 import { generateChildLogger, Logger } from "@walletconnect/logger";
 import { ICore, IEventClient, EventClientTypes } from "@walletconnect/types";
-import { uuidv4 } from "@walletconnect/utils";
+import { formatUA, uuidv4 } from "@walletconnect/utils";
 import {
   CORE_STORAGE_PREFIX,
   EVENTS_CLIENT_API_URL,
@@ -16,12 +16,12 @@ export class EventClient extends IEventClient {
   public readonly context = EVENTS_STORAGE_CONTEXT;
   private readonly storagePrefix = CORE_STORAGE_PREFIX;
   private readonly storageVersion = EVENTS_STORAGE_VERSION;
-
   private events = new Map<string, EventClientTypes.Event>();
   private shouldPersist = false;
   constructor(public core: ICore, public logger: Logger, telemetryEnabled = true) {
     super(core, logger, telemetryEnabled);
     this.logger = generateChildLogger(logger, this.context);
+    this.telemetryEnabled = telemetryEnabled;
     if (telemetryEnabled) {
       this.restore().then(async () => {
         await this.submit();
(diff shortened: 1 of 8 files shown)


packages/core/src/controllers/events.ts
```diff
@@ -39,6 +39,32 @@ export class EventClient extends IEventClient {
     );
   }
 
+  public init: IEventClient["init"] = async () => {
+    if (typeof process !== "undefined" && process.env.IS_VITEST === "true") return;
```
bkrem: Why is this needed? 

Also I believe we have this exact check somewhere else already so if we need it in multiple places it's safer to abstract it into a util fn; avoids e.g. forgetting the `typeof` check and/or updating lots of inline checks if we ever need to change this.

bkrem (approved):

packages/core/src/controllers/events.ts
ganchoradkov: good idea to make it an util.
 >Why is this needed?

so we don't spam logs on dev/ci runs

packages/core/src/controllers/events.ts
```diff
@@ -39,6 +39,32 @@ export class EventClient extends IEventClient {
...
+    try {
+      const initEvent = {
+        eventId: uuidv4(),
+        bundleId: this.core.projectId || "",
```
Cali93: `bundleId` should be used by mobile with app's bundle ID or package Name.
`domain` should be used for web with the current app's domain.

If none is available to you, you can remove both `bundleId` and `domain` and we'll generate one at the API level.

packages/core/src/controllers/events.ts
```diff
@@ -39,6 +39,32 @@ export class EventClient extends IEventClient {
...
+        timestamp: Date.now(),
+        props: {
+          event: "INIT",
+          type: "",
```
Cali93, line 50: cc @dnul if you're relying on the type here, what should the value be ?

packages/core/src/controllers/events.ts
```diff
@@ -189,4 +209,15 @@ export class EventClient extends IEventClient {
       this.logger.warn(error);
     }
   };
+
+  private sendPost = async (events: EventClientTypes.Event[]) => {
```
Cali93: nit
```suggestion
  private sendEvent = async (events: EventClientTypes.Event[]) => {
```

Cali93 (approved):
LGTM given that we address the `bundleId` comment. Did you align on the format with the data team?

packages/core/src/controllers/events.ts
ganchoradkov: as per the specs https://www.notion.so/walletconnect/Walletkit-events-sdk-init-event-08b51439dd8b42c5bbe993c63d2ce54e?pvs=4#49045ac3ba8d4f36900a24047b9cc87b
```
val event: String = "INIT"
val type: String, // None
```

PR #5393, opened 2024-09-26
chore(dot-org): use walletconnect.org for the relay and rpc urls

Cali93:
## Description

We should use walletconnect.org everywhere.

## Type of change

- [X] Chore (non-breaking change that addresses non-functional tasks, maintenance, or code quality improvements)

## How has this been tested?

Covered by existing tests

## Fixes/Resolves (Optional)

Users in some regions getting blocked by their ISPs

## Checklist

- [X] I have performed a self-review of my own code
- [X] My changes generate no new warnings
- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/packages/utils/test/cacao.spec.ts
+++ b/packages/utils/test/cacao.spec.ts
@@ -1,13 +1,10 @@
 import { describe, expect, it } from "vitest";
 import {
-  base64Encode,
-  buildNamespacesFromAuth,
   createEncodedRecap,
   createRecap,
   decodeRecap,
   encodeRecap,
   formatMessage,
-  formatStatementFromRecap,
   getChainsFromRecap,
   getCommonValuesInArrays,
   getDecodedRecapFromResources,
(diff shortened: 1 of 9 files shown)


ganchoradkov (approved):
💯

packages/utils/test/cacao.spec.ts
```diff
@@ -22,7 +19,7 @@ describe("URI", () => {
     it("should merge recaps", () => {
       const recap1 = {
         att: {
-          "https://notify.walletconnect.com": { "manage/all-apps-notifications": [{}] },
```
ganchoradkov: btw these updates might case tests to fail

ganchoradkov (approved):

packages/utils/test/cacao.spec.ts
Cali93: Curious why ? I checked that notify server also runs under notify.walletconnect.org

Cali93: Ah I just noticed it's because the recap test includes a JWT with the old url. Will revert these back then

bkrem (approved):
🙏

PR #5403, opened 2024-09-30
feat: tg mini apps support

ganchoradkov:
## Description
Fixed deep links when in telegram mini apps context
- base64 payload encoding when targetted to https://t.me universal links
- `formatUriFromPairing` util to regenerate pairing uri from pairing to be used as deeplink
- `parseUri` util to handle incoming pairing uris in base64 encoding
- updated `@walletconnect/modal` to latest

## Type of change

## How has this been tested?
in telegram mini apps
https://walletconnect.slack.com/archives/C04DB2EAHE3/p1726817747215279
`2.16.3-rc-tma-1`

## Checklist

### DIFF
--- a/packages/utils/src/misc.ts
+++ b/packages/utils/src/misc.ts
@@ -412,6 +412,7 @@ export async function getDeepLink(storage: IKeyValueStorage, key: string) {
   try {
     if (isBrowser()) {
       link = localStorage.getItem(key) as string;
+      if (link) return link;
     }
     link = await storage.getItem(key);
   } catch (err) {
--- a/packages/utils/src/misc.ts
+++ b/packages/utils/src/misc.ts
@@ -9,6 +9,7 @@ import {
 import { getDocument, getLocation, getNavigator } from "@walletconnect/window-getters";
 import { getWindowMetadata } from "@walletconnect/window-metadata";
 import { ErrorResponse } from "@walletconnect/jsonrpc-utils";
+import { IKeyValueStorage } from "@walletconnect/keyvaluestorage";
 import * as qs from "query-string";
 
 // -- constants -----------------------------------------//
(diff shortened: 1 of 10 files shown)


packages/utils/src/misc.ts
```diff
@@ -369,12 +369,17 @@ export async function handleDeeplinkRedirect({
...
     const json = typeof wcDeepLink === "string" ? JSON.parse(wcDeepLink) : wcDeepLink;
     let deeplink = json?.href;
-
     if (typeof deeplink !== "string") return;
 
+    const payload = `requestId=${id}&sessionTopic=${topic}`;
     if (deeplink.endsWith("/")) deeplink = deeplink.slice(0, -1);
-
-    const link = `${deeplink}/wc?requestId=${id}&sessionTopic=${topic}`;
+    let link = `${deeplink}`;
```
Cali93: Curious why do we need to interpolate this ?
```suggestion
    let link = deeplink;
```

packages/utils/src/misc.ts
```diff
@@ -369,12 +369,17 @@ export async function handleDeeplinkRedirect({
...
+    if (deeplink.startsWith("https://t.me")) {
+      const startApp = deeplink.includes("?") ? "&startapp=" : "?startapp=";
+      link = `${link}${startApp}${toBase64(payload, true)}`;
+    } else {
+      link = `${link}/wc?${payload}`;
+    }
```
Cali93: I'd suggest to use the URL constructor as @enesozturk suggested on the other PR

packages/utils/src/uri.ts
```diff
@@ -17,6 +18,13 @@ export function parseRelayParams(params: any, delimiter = "-"): RelayerTypes.Pro
 }
 
 export function parseUri(str: string): EngineTypes.UriParameters {
+  if (!str.includes("wc:")) {
```
Cali93, line 21: Super nit and not sure if it suits this use case but assuming the `str` always starts with `wc://...` you could also use the URL constructor here and check if the protocol is `wc:`

Cali93 (approved):
LGTM left minor comments

packages/utils/src/misc.ts
ganchoradkov: leftover from previous behaviour

packages/utils/src/misc.ts
ganchoradkov: this url approach doesn't seem to work well with native links, maybe i'm doing something wrong but assigning `/wc?` path results in tripple dashes e.g. 
```
const url = new URL('trust://');
url.pathname = "wc";
-> trust:///wc?

ganchoradkov: will keep the behaviour but added tests to confirm it works as expected

PR #5404, opened 2024-09-30
fix:  json parse unhandled

ganchoradkov:
## Description
Fixed unhandled exception while trying to `json.parse` incoming window messages. Now we try to parse only strings and additional try/catch was added

## Type of change

## How has this been tested?
dogfooding 

## Checklist

### DIFF
--- a/packages/core/src/controllers/verify.ts
+++ b/packages/core/src/controllers/verify.ts
@@ -80,16 +80,21 @@ export class Verify extends IVerify {
         iframe.addEventListener("error", abortListener, { signal: this.abortController.signal });
         const listener = (event: MessageEvent) => {
           if (!event.data) return;
-          const data = JSON.parse(event.data);
-          if (data.type === "verify_attestation") {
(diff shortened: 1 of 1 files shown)


packages/core/src/controllers/verify.ts
```diff
@@ -80,16 +80,21 @@ export class Verify extends IVerify {
...
-            if (decoded.payload.id !== id) return;
-
-            clearInterval(abortTimeout);
-            document.body.removeChild(iframe);
-            this.abortController.signal.removeEventListener("abort", abortListener);
-            window.removeEventListener("message", listener);
-            resolve(data.attestation === null ? "" : data.attestation);
+          if (typeof event.data !== "string") return;
+          try {
+            const data = JSON.parse(event.data);
```
Cali93, line 85: This will throw if the data is a string but not parsable so alternatively you could also have a try catch just around the JSON.parse()
And log event.data with something like, `invalid event payload -> even.data`

Cali93 (approved):
Thanks for the super quick fix 🏎️

PR #5428, opened 2024-10-08
Fix: 1271 sig validation chainId validation

ganchoradkov:
## Description
Fixed a bug where we were not passing CAIP2 formatted `chainId` for 1271 signature validation 
## Type of change

## How has this been tested?
tests

## Checklist

### DIFF
--- a/packages/utils/src/signatures.ts
+++ b/packages/utils/src/signatures.ts
@@ -1,6 +1,7 @@
 import { hashMessage } from "@ethersproject/hash";
 import { recoverAddress } from "@ethersproject/transactions";
 import { AuthTypes } from "@walletconnect/types";
+import { parseChainId } from "./caip";
 const DEFAULT_RPC_URL = "https://rpc.walletconnect.org/v1";
 
 export async function verifySignature(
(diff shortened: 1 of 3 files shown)


packages/utils/src/signatures.ts
```diff
@@ -49,6 +50,12 @@ export async function isValidEip1271Signature(
   projectId: string,
   baseRpcUrl?: string,
 ) {
+  const parsedChain = parseChainId(chainId);
+  if (!parsedChain.namespace || !parsedChain.reference) {
+    throw new Error(
+      `isValidEip1271Signature failed: chainId must be in CAIP-2 format, received: ${chainId}`,
```
Cali93, line 56: Why should it be in CAIP-2 format since EIP-1271 signature is specific to EVM ?

Cali93 (commented):
Curious why this is needed in the first place ? Shouldn't we process the verification even if we don't have a namespace ?

packages/utils/src/signatures.ts
ganchoradkov: blockchain api rejects the request if it isnt. Probably we can update there it there.

Cali93: Wouldn't this be a breaking change if someone provides a custom RPC url and is sending the non compliant CAIP-2 chain ID ?

Cali93 (approved):

packages/utils/src/signatures.ts
ganchoradkov: it shouldn't. These params are opinionated for our rpc api. Direct node urls do not use them

PR #5433, opened 2024-10-09
fix: subscriber publish fail

ganchoradkov:
## Description
fixed an issue where failed subscription attempts were retried instead of throwing an error

## Type of change

## How has this been tested?
tests and in canary - `2.17.1-canary-0`

## Checklist

### DIFF
--- a/packages/core/src/controllers/subscriber.ts
+++ b/packages/core/src/controllers/subscriber.ts
@@ -236,6 +236,7 @@ export class Subscriber extends ISubscriber {
     };
     this.logger.debug(`Outgoing Relay Payload`);
     this.logger.trace({ type: "payload", direction: "outgoing", request });
+    const shouldThrow = opts?.internal?.throwOnFailedPublish;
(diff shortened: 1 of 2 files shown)


packages/core/src/controllers/subscriber.ts
```diff
@@ -236,6 +236,7 @@ export class Subscriber extends ISubscriber {
     };
     this.logger.debug(`Outgoing Relay Payload`);
     this.logger.trace({ type: "payload", direction: "outgoing", request });
+    const shouldThrow = opts?.internal?.throwOnFailedPublish;
```
ignaciosantise, line 239: this is true by default now, right? Would it cause issues in wallets that upgrade to this version? 🤔

ganchoradkov: nope. It would raise early feedback if there are network issues

ignaciosantise (approved):

PR #5515, opened 2024-11-22
fix:  connection robustness and flaky tests

ganchoradkov:
## Description
Reworked several parts of the client to address general stability issues and to increase robustness of the socket connections
- Implemented multi attempt connection logic in the relayer
- Implemented graceful handling of disconnects mid publish
- Reworked subscriber to attempt initial sub and let heartbeat retry to subscribe until success or reaching a timeout
- Reworked publisher to attempt initial publish and let heartbeat retry to publish until success or reaching a timeout
- Fixed several flaky tests that failed in bad network conditions with many disconnects 
- Improved error logging

## Type of change

## How has this been tested?
canary - `2.17.2-canary-rcnt-3`
deployment - https://react-dapp-v2-git-chore-connection-test-reown-com.vercel.app/

## Checklist

### DIFF
--- a/packages/core/src/controllers/publisher.ts
+++ b/packages/core/src/controllers/publisher.ts
@@ -114,6 +114,8 @@ export class Publisher extends IPublisher {
       });
 
       await createExpiringPromise(publishPromise, this.publishTimeout, failedPublishMessage);
+      // @ts-expect-error
+      global.setSentTopic(topic);
     } catch (e) {
       this.logger.debug(`Failed to Publish Payload`);
       this.logger.error(e as any);
--- a/packages/core/src/controllers/publisher.ts
+++ b/packages/core/src/controllers/publisher.ts
@@ -115,7 +115,7 @@ export class Publisher extends IPublisher {
 
       await createExpiringPromise(publishPromise, this.publishTimeout, failedPublishMessage);
       // @ts-expect-error
-      global?.setSentTopic(topic);
+      global?.setSentTopic?.(topic);
     } catch (e) {
       this.logger.debug(`Failed to Publish Payload`);
       this.logger.error(e as any);
--- a/packages/core/src/controllers/publisher.ts
+++ b/packages/core/src/controllers/publisher.ts
@@ -115,7 +115,7 @@ export class Publisher extends IPublisher {
 
       await createExpiringPromise(publishPromise, this.publishTimeout, failedPublishMessage);
       // @ts-expect-error
-      global.setSentTopic(topic);
+      global?.setSentTopic(topic);
     } catch (e) {
       this.logger.debug(`Failed to Publish Payload`);
       this.logger.error(e as any);
--- a/packages/core/src/controllers/publisher.ts
+++ b/packages/core/src/controllers/publisher.ts
@@ -88,24 +88,22 @@ export class Publisher extends IPublisher {
               id,
               attestation: opts?.attestation,
             })
+              .then(resolve)
               .catch((e) => {
                 this.logger.warn(e, e?.message);
                 reject(e);
-              })
-              .then(resolve);
+              });
           }),
           this.initialPublishTimeout,
           `Failed initial publish, retrying.... id:${id} tag:${tag}`,
         );
-        await initialPublish
-          .then((result) => {
-            this.events.removeListener(RELAYER_EVENTS.publish, onPublish);
-            resolve(result);
-          })
-          .catch((e) => {
-            this.queue.set(id, { ...params, attempt: 1 });
-            this.logger.warn(e, e?.message);
-          });
+        try {
+          await initialPublish;
+          this.events.removeListener(RELAYER_EVENTS.publish, onPublish);
+        } catch (e) {
+          this.queue.set(id, { ...params, attempt: 1 });
+          this.logger.warn(e, (e as Error)?.message);
+        }
       });
       this.logger.trace({
         type: "method",
--- a/packages/core/src/controllers/publisher.ts
+++ b/packages/core/src/controllers/publisher.ts
@@ -78,17 +78,24 @@ export class Publisher extends IPublisher {
         };
         this.relayer.events.on(RELAYER_EVENTS.publish, onPublish);
         const initialPublish = createExpiringPromise(
-          this.rpcPublish({
-            topic,
-            message,
-            ttl,
-            prompt,
-            tag,
-            id,
-            attestation: opts?.attestation,
-          }).catch((e) => this.logger.warn(e, e?.message)),
+          new Promise((resolve, reject) => {
+            this.rpcPublish({
+              topic,
+              message,
+              ttl,
+              prompt,
+              tag,
+              id,
+              attestation: opts?.attestation,
+            })
+              .catch((e) => {
+                this.logger.warn(e, e?.message);
+                reject(e);
+              })
+              .then(resolve);
+          }),
           this.initialPublishTimeout,
-          `Failed to publish payload, please try again. id:${id} tag:${tag}`,
+          `Failed initial publish, retrying.... id:${id} tag:${tag}`,
         );
         await initialPublish
           .then((result) => {
@@ -178,7 +185,10 @@ export class Publisher extends IPublisher {
       const attempt = params.attempt + 1;
       this.queue.set(id, { ...params, attempt });
       const { topic, message, opts, attestation } = params;
-      this.logger.trace({}, `Publisher: queue->publishing: ${params.opts.id}, attempt: ${attempt}`);
+      this.logger.warn(
+        {},
+        `Publisher: queue->publishing: ${params.opts.id}, tag: ${params.opts.tag}, attempt: ${attempt}`,
+      );
       await this.rpcPublish({
         topic,
         message,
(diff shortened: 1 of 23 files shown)


packages/core/src/controllers/relayer.ts
```diff
@@ -400,6 +322,78 @@ export class Relayer extends IRelayer {
...
+
+        await new Promise<void>(async (resolve, reject) => {
+          const onDisconnect = () => {
+            reject(new Error(`Connection interrupted while trying to subscribe`));
+          };
+          this.provider.once(RELAYER_PROVIDER_EVENTS.disconnect, onDisconnect);
+
+          await createExpiringPromise(
+            this.provider.connect(),
+            15_000,
```
ignaciosantise: why `15_000` and not `15000`? I remember having some issues in RN with numbers with underscore

ganchoradkov: updated

packages/core/test/relayer.spec.ts
```diff
@@ -350,7 +301,7 @@ describe("Relayer", () => {
         await Promise.all(
           Array.from(Array(disconnectsToEmit).keys()).map(() => relayer.onDisconnectHandler()),
         );
-        await throttle(1000);
+        await throttle(5_000);
```
ignaciosantise, line 304: same for all underscore numbers. Well, the tests doesn't count but to keep consistency?

packages/sign-client/src/client.ts
```diff
@@ -252,7 +252,9 @@ export class SignClient extends ISignClient {
       await this.auth.init();
       await this.engine.init();
       this.logger.info(`SignClient Initialization Success`);
-      this.engine.processRelayMessageCache();
+      setTimeout(() => {
+        this.engine.processRelayMessageCache();
+      }, 1_000);
```
ignaciosantise:
```suggestion
      }, 1000);
```

packages/core/test/relayer.spec.ts
ignaciosantise, line 304:
```suggestion
        await throttle(5000);
```

packages/core/test/relayer.spec.ts
```diff
@@ -315,25 +259,32 @@ describe("Relayer", () => {
...
       it("should restart transport after connection drop", async () => {
         const randomSessionIdentifier = relayer.core.crypto.randomSessionIdentifier;
-        await relayer.provider.connection.close();
-        expect(relayer.connected).to.be.false;
-        await throttle(1000);
-        expect(relayer.connected).to.be.true;
+        await relayer.transportOpen();
+        const timeout = setTimeout(() => {
+          throw new Error("Connection did not restart after disconnect");
+        }, 5_001);
```
ignaciosantise, line 271:
```suggestion
        }, 5001);
```

packages/core/src/controllers/publisher.ts
```diff
@@ -57,52 +61,67 @@ export class Publisher extends IPublisher {
...
+      const publishPromise = new Promise(async (resolve) => {
+        const onPublish = ({ id }: { id: string }) => {
+          if (params.opts.id === id) {
+            this.removeRequestFromQueue(id);
+            this.relayer.events.removeListener(RELAYER_EVENTS.publish, onPublish);
+            resolve(params);
+          }
+        };
+        this.relayer.events.on(RELAYER_EVENTS.publish, onPublish);
+        const initialPublish = createExpiringPromise(
```
Cali93, line 80: Not directly related to this PR but `createExpiringPromise` could use `Promise.race` to be simplified and avoid potential race conditions.
This approach eliminates the need for manual `clearTimeout` calls because once the Promise settles, the other Promise in the race is ignored, and the timer will be garbage collected.
```ts
export function createExpiringPromise<T>(
  promise: Promise<T>,
  expiry: number,
  expireErrorMessage?: string,
): Promise<T> {
  const timeoutPromise = new Promise<T>((_, reject) =>
    setTimeout(() => reject(new Error(expireErrorMessage)), expiry),
  );

  return Promise.race([promise, timeoutPromise]);
}
```

packages/core/src/controllers/publisher.ts
```diff
@@ -57,52 +61,67 @@ export class Publisher extends IPublisher {
...
+          new Promise((resolve, reject) => {
```
Cali93, line 81: Why do you have to wrap your promise ? You should be able to pass this.rpcPublish directly

packages/core/src/controllers/publisher.ts
```diff
@@ -57,52 +61,67 @@ export class Publisher extends IPublisher {
...
+              prompt,
+              tag,
+              id,
+              attestation: opts?.attestation,
+            })
+              .catch((e) => {
+                this.logger.warn(e, e?.message);
+                reject(e);
+              })
+              .then(resolve);
```
Cali93: nit `.then` should be used before `.catch`
```suggestion
              .then(resolve);
              .catch((e) => {
                this.logger.warn(e, e?.message);
                reject(e);
              })
```

packages/core/test/relayer.spec.ts
ganchoradkov: I will leave tests as is

packages/core/src/controllers/publisher.ts
```diff
@@ -57,52 +61,67 @@ export class Publisher extends IPublisher {
...
-      this.logger.debug(`Successfully Published Payload`);
+        await initialPublish
+          .then((result) => {
+            this.events.removeListener(RELAYER_EVENTS.publish, onPublish);
+            resolve(result);
+          })
+          .catch((e) => {
+            this.queue.set(id, { ...params, attempt: 1 });
+            this.logger.warn(e, e?.message);
+          });
```
Cali93: Not a big fan of how we're combining the use of the promise constructor with .then .catch and async/await + try/catch
I think the code could be easier to reason about if we were just using the promise constructor with async await and the try/catch block.

packages/core/src/controllers/publisher.ts
```diff
@@ -57,52 +61,67 @@ export class Publisher extends IPublisher {
...
+      await createExpiringPromise(publishPromise, this.publishTimeout, failedPublishMessage);
     } catch (e) {
       this.logger.debug(`Failed to Publish Payload`);
       this.logger.error(e as any);
       if (opts?.internal?.throwOnFailedPublish) {
         throw e;
       }
-      this.queue.set(id, params);
+    } finally {
+      this.queue.delete(id);
```
Cali93, line 122: So previously we were setting the request in the queue so that it's retired but now we're deleting it from the queue because the retry logic is part of the `initialPublish` right ?

packages/core/src/controllers/publisher.ts
```diff
@@ -151,17 +170,35 @@ export class Publisher extends IPublisher {
...
   private checkQueue() {
-    this.queue.forEach(async (params) => {
-      const { topic, message, opts } = params;
-      await this.publish(topic, message, opts);
+    this.queue.forEach(async (params, id) => {
+      const attempt = params.attempt + 1;
+      this.queue.set(id, { ...params, attempt });
+      const { topic, message, opts, attestation } = params;
+      this.logger.warn(
+        {},
```
Cali93, line 187: Why are we passing an empty object here ?

packages/core/src/controllers/subscriber.ts
```diff
@@ -450,46 +492,49 @@ export class Subscriber extends ISubscriber {
...
     this.logger.trace(`Fetching batch messages for ${subscriptions.length} subscriptions`);
     const response = await this.rpcBatchFetchMessages(subscriptions);
     if (response && response.messages) {
-      this.pendingBatchMessages = this.pendingBatchMessages.concat(response.messages);
+      await new Promise<void>((resolve) => {
+        setTimeout(async () => {
+          await this.relayer.handleBatchMessageEvents(response.messages);
+          resolve();
+        }, 1000);
+      });
```
Cali93: I feel like this could be way more concise and readable
```suggestion
await delay(1000); // or sleep(1000)
await this.relayer.handleBatchMessageEvents(response.messages);
```

Implementation of the sleep or delay fn
```ts
function delay(ms: number) {
  return new Promise<void>((resolve) => setTimeout(resolve, ms));
}
```

Cali93 (approved):
LGTM, left comments mainly around readability.

packages/core/src/controllers/publisher.ts
ganchoradkov: correct. Now the request is set into the queue only after `initialPublish` fails

packages/core/src/controllers/publisher.ts
ganchoradkov: its just for the logger, there is no error object to pass

packages/core/src/controllers/publisher.ts
ganchoradkov: technically this way its safer because `this.rpcPublish` is only invoked when `initialPublish` variable is awaited

packages/core/src/controllers/publisher.ts
ganchoradkov: will test it out duing tech debt week

PR #5537, opened 2024-12-06
calling resetPingTimeout on each ping

paranormal:

## Description

We built a web wallet app and noticed an unusually high number of `relayer disconnected` messages in logs. The relayer was not resetting the ping timeout on every ping, just on the first one. This MR fixes this. Also, I've noticed that increasing the wait interval to an additional 4 seconds made the error disappear completely.

## Type of change

## How has this been tested?

After the fix, I see only `Disconnecting for load balancing reasons` disconnects in server logs.

Use this table template to show examples of your changes:

## Checklist

### DIFF
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -19,7 +19,7 @@ import {
   Logger,
 } from "@walletconnect/logger";
 import { RelayJsonRpc } from "@walletconnect/relay-api";
-import { FIVE_MINUTES, ONE_SECOND, THIRTY_SECONDS, toMiliseconds } from "@walletconnect/time";
+import { FIVE_MINUTES, FIVE_SECONDS, THIRTY_SECONDS, toMiliseconds } from "@walletconnect/time";
 import {
   ICore,
   IMessageTracker,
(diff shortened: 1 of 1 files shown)


packages/core/src/controllers/relayer.ts
```diff
@@ -19,7 +19,7 @@ import {
   Logger,
 } from "@walletconnect/logger";
 import { RelayJsonRpc } from "@walletconnect/relay-api";
-import { FIVE_MINUTES, ONE_SECOND, THIRTY_SECONDS, toMiliseconds } from "@walletconnect/time";
+import { FIVE_MINUTES, FIVE_SECONDS, THIRTY_SECONDS, toMiliseconds } from "@walletconnect/time";
```
```suggestion
import { FIVE_MINUTES, ONE_SECOND, FIVE_SECONDS, THIRTY_SECONDS, toMiliseconds } from "@walletconnect/time";
```
could you re-add the `ONE_SECOND` constant as its used in other places around the relayer

ganchoradkov (changes requested):
Thank you for the patch!  Please re-add the missing const as its breaking the build

ganchoradkov (approved):
💯

PR #5582, opened 2025-01-08
chore: adds deprecation notice on pairing methods and reduces expiry …

ganchoradkov:
## Description
Supersedes https://github.com/WalletConnect/walletconnect-monorepo/pull/5310
- Added deprecation warnings on pairings methods
- Reduced expiry of activated pairings to `five minutes` instead of `thirty days` essentially removing stale pairings without introing breaking changes

## Type of change

## How has this been tested?
- addtional tests to confirm the new behaviour
- the old tests to confirm no breaking changes

## Checklist


packages/core/src/controllers/pairing.ts
```diff
@@ -187,7 +187,7 @@ export class Pairing implements IPairing {
 
   public activate: IPairing["activate"] = async ({ topic }) => {
     this.isInitialized();
-    const expiry = calcExpiry(THIRTY_DAYS);
+    const expiry = calcExpiry(ONE_MINUTE);
```
llbartekll: shouldn't be 5 minutes? also you don't need an activate method anymore
look into the doc https://www.notion.so/walletconnect/Improving-the-Sign-SDK-Pairing-Expiry-Reduction-d4cf787ecd9f4169a6887d45bb5d5069

```Deprecation: The concepts of active and inactive pairings will no longer apply. Pairings will only exist until the first session is created. The pairing_delete, pairing_extend, and pairing_ping methods will be deprecated.```

File: packages/core/src/controllers/pairing.ts [hunk 2]
```diff
@@ -35,7 +35,7 @@ import {
   isJsonRpcResult,
   isJsonRpcError,
 } from "@walletconnect/jsonrpc-utils";
-import { FIVE_MINUTES, THIRTY_DAYS, toMiliseconds } from "@walletconnect/time";
+import { FIVE_MINUTES, ONE_MINUTE, toMiliseconds } from "@walletconnect/time";
```
llbartekll:
```
 After the session proposal is delivered from the dapp, the wallet will persist the pairing for 5 minutes if the proposal is not responded and then remove the pairing from local storage. ```

**Review by llbartekll — CHANGES_REQUESTED** [review]

File: packages/core/src/controllers/pairing.ts [hunk 3]
```diff
@@ -187,14 +187,15 @@ export class Pairing implements IPairing {
 
   public activate: IPairing["activate"] = async ({ topic }) => {
     this.isInitialized();
-    const expiry = calcExpiry(THIRTY_DAYS);
+    const expiry = calcExpiry(FIVE_MINUTES);
     this.core.expirer.set(topic, expiry);
     await this.pairings.update(topic, { active: true, expiry });
   };
 
   public ping: IPairing["ping"] = async (params) => {
```
devceline, line 198: We need to add JSDoc so people see it in their IDEs.
```suggestion
 /**
 * @deprecated Ping will be removed in the next major release.
 */
  public ping: IPairing["ping"] = async (params) => {
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -681,6 +683,9 @@ export class Engine extends IEngine {
         done(),
       ]);
     } else if (this.client.core.pairing.pairings.keys.includes(topic)) {
+      this.client.logger.warn(
+        "ping() on pairing topic is deprecated and will be removed in the next major release.",
```
devceline, line 687: Should it say ping here?

Cali93 (commented):
Do we have docs where we can point folks to for more context on the deprecation ?

packages/sign-client/src/controllers/engine.ts
ganchoradkov: yes, since we have single `PING()` for both session & pairing topics, only when pairing is provided we log the warning

ganchoradkov:
>Do we have docs where we can point folks to for more context on the deprecation ?
Not yet, these are obscure behaviours that are not really documented

devceline (approved):
LGTM

PR #5625, opened 2025-01-31
chore: remove `defineChain`

0xmkh:


providers/ethereum-provider/src/wcmToAppKit.ts
```diff
@@ -38,9 +38,13 @@ const mapCaipIdToAppKitCaipNetwork = (caipId: CaipNetworkId): CaipNetwork => {
     rpcUrls: {
       default: { http: ["https://rpc.walletconnect.org/v1"] },
     },
-  });
+    // Come from defineChain in @reown/appkit/networks
```
tomiir, line 41: won't this cause viem to be bundled inside of appkit?

PR #5630, opened 2025-02-03
feat: tvf

ganchoradkov:
## Description
Implemented automatic parsing and adding of TVF params to relay message payloads on both peer sides. These params are assigned on a few selected rpc methods as per [specs](https://www.notion.so/walletconnect/TVF-tracking-in-WalletKit-tech-doc-14a3a661771e8094a056db28c0005e6e?pvs=4)

## Type of change

## How has this been tested?
tests

## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -2996,7 +2996,7 @@ export class Engine extends IEngine {
   ) => {
     try {
       const tvf: RelayerTypes.ITVF = {
-        correlationId: id.toString(),
+        correlationId: id,
         rpcMethods: [params.request.method],
         chainId: params.chainId,
         ...(this.isValidContractData(params.request.params) && {
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -597,7 +597,6 @@ export class Engine extends IEngine {
       chainId,
     };
     const shouldSetTVF = this.shouldSetTVF(protocolMethod, protocolRequestParams);
-    console.log("shouldSetTVF", shouldSetTVF, protocolMethod, chainId, request.method);
 
     return await Promise.all([
       new Promise<void>(async (resolve) => {
@@ -1494,9 +1493,9 @@ export class Engine extends IEngine {
       if (expiry) opts.ttl = expiry;
       if (relayRpcId) opts.id = relayRpcId;
 
-      opts.internal = {
-        ...opts.internal,
-        tvf,
+      opts.tvf = {
+        ...tvf,
+        correlationId: payload.id,
       };
 
       if (throwOnFailedPublish) {
@@ -1540,11 +1539,7 @@ export class Engine extends IEngine {
       const request = record.request;
       try {
         if (this.shouldSetTVF(request.method as JsonRpcTypes.WcMethod, request.params)) {
-          tvf = this.getTVFParams(
-            id,
-            request.params,
-            isValidArray(result) ? result : ([result] as any),
-          );
+          tvf = this.getTVFParams(id, request.params, result);
         }
       } catch (error) {
         this.client.logger.warn(`sendResult() -> getTVFParams() failed`, error);
@@ -1559,10 +1554,12 @@ export class Engine extends IEngine {
       await (global as any).Linking.openURL(redirectURL, this.client.name);
     } else {
       const opts = ENGINE_RPC_OPTS[record.request.method].res;
-      opts.internal = {
-        ...opts.internal,
-        tvf,
+
+      opts.tvf = {
+        ...tvf,
+        correlationId: id,
       };
+
       if (throwOnFailedPublish) {
         opts.internal = {
           ...opts.internal,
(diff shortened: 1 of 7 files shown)


packages/sign-client/src/controllers/engine.ts
```diff
@@ -587,22 +589,29 @@ export class Engine extends IEngine {
...
 
+    const protocolRequestParams: JsonRpcTypes.RequestParams["wc_sessionRequest"] = {
+      request: {
+        ...request,
+        expiryTimestamp: calcExpiry(expiry),
+      },
+      chainId,
+    };
+    const shouldSetTVF = this.shouldSetTVF(protocolMethod, protocolRequestParams);
+    console.log("shouldSetTVF", shouldSetTVF, protocolMethod, chainId, request.method);
```
Cali93:
```suggestion
```

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1518,8 +1534,21 @@ export class Engine extends IEngine {
       throw error;
     }
     let record;
+    let tvf;
     try {
       record = await this.client.core.history.get(topic, id);
+      const request = record.request;
```
Cali93, line 1539: Should we validate/parse the request to avoid having to cast the method to `JsonRpcTypes.WcMethod`?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -2941,4 +2974,56 @@ export class Engine extends IEngine {
...
+    params: JsonRpcTypes.RequestParams["wc_sessionRequest"],
+  ) => {
+    if (!params) return false;
+    if (protocolMethod !== "wc_sessionRequest") return false;
+    const { request, chainId } = params;
+    const namespace = parseChainId(chainId).namespace;
+    if (!(namespace in TVF_METHODS)) return false;
+    const tvfMethods = TVF_METHODS[namespace as keyof typeof TVF_METHODS];
+    if (!tvfMethods) return false;
+    return tvfMethods.includes(request.method);
```
Cali93: Do you think it'd make sense to introduce a validation library with clear schemas for handling validation ?

packages/sign-client/src/controllers/engine.ts
```diff
@@ -2941,4 +2974,56 @@ export class Engine extends IEngine {
...
+          // initially only get contractAddresses from EVM txs
+          contractAddresses: [params.request.params?.[0]?.to],
+        }),
+        txHashes,
+      };
+      return tvf;
+    } catch (e) {
+      this.client.logger.warn("Error getting TVF params", e);
+    }
+    return {};
```
Cali93, line 3007: Curious why you're adopting this pattern of returning outside of the catch ?

packages/sign-client/test/sdk/client.spec.ts
```diff
@@ -493,6 +493,91 @@ describe("Sign Client Integration", () => {
           await throttle(1_000);
           await deleteClients(clients);
         });
+
+        it("can set tvf params", async () => {
```
Cali93, line 497: Is that the only test case we need ? What about checking validation errors etc ?

Cali93 (approved):
LGTM, feel free to merge when all comments are resolved, specially the one for more test cases for the unhappy path.

packages/sign-client/src/controllers/engine.ts
```diff
@@ -2941,4 +2974,56 @@ export class Engine extends IEngine {
...
+  };
+
+  private getTVFParams = (
+    id: number,
+    params: JsonRpcTypes.RequestParams["wc_sessionRequest"],
+    txHashes?: string[],
+  ) => {
+    try {
+      const tvf: RelayerTypes.ITVF = {
+        correlationId: id,
```
jakubuid, line 2994: `correlationId` must be set on all Sign tags not only on wc_sessionRequest (1108)

packages/sign-client/src/controllers/engine.ts
```diff
@@ -2941,4 +2974,56 @@ export class Engine extends IEngine {
...
+    if (!(namespace in TVF_METHODS)) return false;
+    const tvfMethods = TVF_METHODS[namespace as keyof typeof TVF_METHODS];
+    if (!tvfMethods) return false;
+    return tvfMethods.includes(request.method);
+  };
+
+  private getTVFParams = (
+    id: number,
+    params: JsonRpcTypes.RequestParams["wc_sessionRequest"],
+    txHashes?: string[],
```
jakubuid: how do you set txHashes? can you point to the logic that parses these valuse for Solana RPCs?

ganchoradkov: this line provides the txHashes https://github.com/WalletConnect/walletconnect-monorepo/blob/7449929e08f2b2d87916f2b1e9076afbf7c6eb7d/packages/sign-client/src/controllers/engine.ts#L1546
when the call is response to a session request that passes this check
https://github.com/WalletConnect/walletconnect-monorepo/blob/7449929e08f2b2d87916f2b1e9076afbf7c6eb7d/packages/sign-client/src/controllers/engine.ts#L1542
it means the method is one of the following 
https://github.com/WalletConnect/walletconnect-monorepo/blob/7449929e08f2b2d87916f2b1e9076afbf7c6eb7d/packages/sign-client/src/constants/engine.ts#L148

jakubuid: Solana RPCs we support encapsulates the result with `transactions` e.g.
see:
https://docs.reown.com/advanced/multichain/rpc-reference/solana-rpc#solana_signalltransactions

jakubuid: Check: https://github.com/reown-com/reown-kotlin/pull/46/files#diff-d87d8f12338566a38f75a24780006abad01dd5865be0e803b842c36be0c5012aR33

ganchoradkov: updated!

packages/sign-client/src/controllers/engine.ts
ganchoradkov: the parsing is simplified now so probably overkill to export it into its own library. if we expand the handling at later point, why not

packages/sign-client/src/controllers/engine.ts
ganchoradkov: `request.method`type is `string` so i'm casting it to predefined set of protocol methods to be more intuitive of the values. I can try to refactor the type `JsonRpcRecord` that history returns but it might cause unexpected breaking changes in the types

packages/sign-client/src/controllers/engine.ts
ganchoradkov: seems easier to read whats going on in the fx

packages/sign-client/test/sdk/client.spec.ts
ganchoradkov: added extra validations for the different namespaces and result structures. Validation errors are tested with the other tests where different methods/values are used and don't assign tvf

PR #5654, opened 2025-02-18
fix: relayer would only connect if there are topics present in the cl…

ganchoradkov:
## Description
Implemented two changes based on https://www.notion.so/walletconnect/Core-SDK-Relay-reconnects-optimisation-1903a661771e80cebb47ccfae9fd905f
- `transportOpen` will only open ws connection if there are topics for the client to subscribe
- reconnection logic e.g. after ws drop, also checks for topics before reopening the connection, otherwise skips  

## Type of change

## How has this been tested?
tests

## Checklist


packages/core/src/controllers/relayer.ts
```diff
@@ -268,6 +272,13 @@ export class Relayer extends IRelayer {
   }
 
   async transportOpen(relayUrl?: string) {
+    if (!this.subscriber.hasAnyTopics) {
+      this.logger.warn(
+        "Starting WS connection skipped because the client has no topics to work with.",
```
arein, line 277: should this be `warn` level or `debug`?

ganchoradkov: it should be an error and the call should be rejected technically but we want to avoid breaking changes. Warn specifies something is not behaving as expected and I think it fits best. Debug will rarely be used by devs as its quite noisy

arein (commented):
looks good - can you link me to the release process please? I suggest we test it in the lab

packages/core/src/controllers/relayer.ts
arein: got it. So we don't want folks to interact with this API? Can we test on AppKit just to make sure we don't get loads of warnings now unexpectedly?

ganchoradkov:
> looks good - can you link me to the release process please? I suggest we test it in the lab

the release checklist
https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/.github/pr_template_release.md

arein (approved):
LGTM let's make sure to do some thorough testing on AppKit and on the sample app

PR #5668, opened 2025-02-24
fix: problematic session completion

ganchoradkov:
## Description
Fixed an annoying `Missing or invalid. Record was recently deleted - pairing` issue [reported here](https://reown-inc.slack.com/archives/C07BNE1C39C/p1740145763566439?thread_ts=1740059889.947619&cid=C07BNE1C39C) which happens because Sign Client engine used single non unique event for completing a session - `session_connect`. When the client had multiple pending proposals, establishing a session was triggering  the `session_connect` listener for all others resulting in the error log. This also means that the other pending proposals can no longer finalize a session because their `session_connect` listener was wrongly triggered. 

I've implemented
- unique `session_connect`+`proposalId`  listener for each proposal allowing multiple sessions to be established at the same time
- refactored the session settle behaviour to not rely on event listeners. This allows the sdk to complete a session even after restart
- added new client event `session_connect` to allow apps to subscribe for new sessions completed after client restart 
 

## Type of change

## How has this been tested?
tests

## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -147,6 +147,15 @@ export class Engine extends IEngine {
 
   private recentlyDeletedLimit = 200;
   private relayMessageCache: RelayerTypes.MessageEvent[] = [];
+  private pendingSessions: Map<
+    number,
+    {
+      sessionTopic: string;
+      pairingTopic: string;
+      proposalId: number;
+      publicKey: string;
+    }
+  > = new Map();
 
   constructor(client: IEngine["client"]) {
     super(client);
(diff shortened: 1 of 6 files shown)


packages/sign-client/test/sdk/persistence.spec.ts
```diff
@@ -206,6 +207,63 @@ describe("Sign Client Persistence", () => {
           // delete
           await deleteClients(clients);
         });
+
+        it("should complete a session after dapp restart", async () => {
```
bkrem, line 211: Can we also add a test to pin the new desired behaviour? 🙏 

E.g. 
1. send proposal 1
2. send proposal 2
3. accept 1
4. can still accept 2

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1918,18 +1920,28 @@ export class Engine extends IEngine {
       this.isValidSessionSettleRequest(params);
       const { relay, controller, expiry, namespaces, sessionProperties, sessionConfig } =
         payload.params;
-      const session = {
+      const pendingSession = [...this.pendingSessions.values()].find(
+        (s) => s.sessionTopic === topic,
+      );
+
+      if (!pendingSession) {
+        return this.client.logger.error(`Pending session not found for topic ${topic}`);
```
bkrem, line 1928: Is it safe for us to return `client.logger`?

Only other place I see us doing this is here and it's not for an error: https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/packages/sign-client/src/controllers/engine.ts#L1763

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1918,18 +1920,28 @@ export class Engine extends IEngine {
...
+      const session: SessionTypes.Struct = {
         topic,
         relay,
         expiry,
         namespaces,
         acknowledged: true,
-        pairingTopic: "", // pairingTopic will be set in the `session_connect` handler
-        requiredNamespaces: {},
-        optionalNamespaces: {},
+        pairingTopic: pendingSession.pairingTopic,
```
bkrem, line 1939: Nice, this was always an awkward workaround to deal with the chicken-egg problem here prior 💯

packages/sign-client/src/controllers/engine.ts
```diff
@@ -1940,12 +1952,22 @@ export class Engine extends IEngine {
...
+        topic: pendingSession.pairingTopic,
+        metadata: session.peer.metadata,
+      });
+
+      this.client.events.emit("session_connect", { session });
+      this.events.emit(engineEvent("session_connect", pendingSession.proposalId), { session });
+
+      this.pendingSessions.delete(pendingSession.proposalId);
+      this.deleteProposal(pendingSession.proposalId, false);
+      this.cleanupDuplicatePairings(session);
```
bkrem, line 1969: How come we add the `cleanupDuplicatePairings` call here?

bkrem (approved):
Thank you for fixing, improving the flow here and overall cleaning up the code 🙏 💯

packages/sign-client/src/controllers/engine.ts
ganchoradkov: its fine since there is no usage of `onRelayEventRequest`'s return. Thought it look tidy in 1 line, will update to be safe

packages/sign-client/src/controllers/engine.ts
ganchoradkov: It was moved from the `session_connect` event handler because now all session completion logic is done in session settle handler

packages/sign-client/test/sdk/persistence.spec.ts
ganchoradkov: Added https://github.com/WalletConnect/walletconnect-monorepo/pull/5668/commits/7ff2824683a3e2dbeb78a924e2fcbbadae167ea1

PR #5673, opened 2025-03-04
chore: solana tx

ganchoradkov:
## Description
Added custom parsing for `solana_signAllTransactions`

## Type of change

## How has this been tested?
tests

## Checklist

### DIFF
--- a/packages/utils/src/signatures.ts
+++ b/packages/utils/src/signatures.ts
@@ -1,7 +1,10 @@
 import { keccak_256 } from "@noble/hashes/sha3";
 import { recoverAddress } from "viem";
 import { AuthTypes } from "@walletconnect/types";
+import bs58 from "bs58";
+
 import { parseChainId } from "./caip";
+
 const DEFAULT_RPC_URL = "https://rpc.walletconnect.org/v1";
 
 export function hashEthereumMessage(message: string) {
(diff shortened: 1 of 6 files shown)


packages/utils/src/signatures.ts
```diff
@@ -106,3 +109,35 @@ export async function isValidEip1271Signature(
 function generateJsonRpcId() {
   return Date.now() + Math.floor(Math.random() * 1000);
 }
+
+export function extractSolanaTransactionId(solanaTransaction: string): string {
+  const binary = atob(solanaTransaction);
```
jakubuid, line 114: what's `atob`?

in Kotlin i'm trying to decode base64 if fails base58

jakubuid (approved):

packages/utils/src/signatures.ts
ganchoradkov: `atob` decodes base64 string. I added a bit extra validation just in case to avoid handling dummy data. 
For example, atob throws if the string is not valid base64 encoding, and the validation after that will filer base64 encoded non transaction values

jakubuid: got it
phantom docs says that these signed transactions can be base58 encoded but our solana rpc docs say that they can be base64
i think we should add decoding for both

PR #5799, opened 2025-03-24
feat: conditional qr modal ep

ganchoradkov:
## Description
Added new util method to `ethereum-provider`, `updateConfigOptions` which lets devs update the provider configuration after its initialization. This allows for conditional QR modal behaviour where devs might choose to enable the modal even if they had disabled it during init. This wasn't possible before and required complex work-arounds such as running duplicate provider instances - one with modal enabled and one with the modal disabled 

## Type of change

## How has this been tested?
tests, dogfooding 

## Checklist


providers/ethereum-provider/src/EthereumProvider.ts
```diff
@@ -290,6 +290,7 @@ export class EthereumProvider implements IEthereumProvider {
       const session = await new Promise<SessionTypes.Struct | undefined>(
         async (resolve, reject) => {
           if (this.rpc.showQrModal) {
+            await this.loadWalletConnectModal();
             this.modal?.subscribeModal((state: { open: boolean }) => {
```
chris13524: Maybe refactor `loadWalletConnectModal` to be a "get instance" method and return the modal instance? Then you don't need to manually load the modal when you need it, because the only way to obtain it is to load it.

chris13524 (approved):

PR #5807, opened 2025-03-26
fix: auto metadata

ganchoradkov:
## Description
Implemented auto population of parially missing `metadata` fields. 
Context https://reown-inc.slack.com/archives/C04DB2EAHE3/p1742872367620239

## Type of change

## How has this been tested?
tests, dogfooding

## Checklist

### DIFF
--- a/packages/utils/src/misc.ts
+++ b/packages/utils/src/misc.ts
@@ -113,6 +113,37 @@ export function appendToQueryString(
 
 // -- metadata ----------------------------------------------//
 
+export function populateAppMetadata(metadata?: SignClientTypes.Metadata): SignClientTypes.Metadata {
+  const appMetadata = getAppMetadata();
+  try {
+    if (metadata?.url && appMetadata.url) {
+      if (metadata.url !== appMetadata.url) {
+        console.warn(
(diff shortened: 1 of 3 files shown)


packages/utils/src/misc.ts
```diff
@@ -113,6 +113,37 @@ export function appendToQueryString(
...
+      ...appMetadata,
+      ...metadata,
+      url: metadata?.url || appMetadata.url,
+      name: metadata?.name || appMetadata.name,
+      description: metadata?.description || appMetadata.description,
+      icons:
+        metadata?.icons?.length && metadata.icons.length > 0 ? metadata.icons : appMetadata.icons,
+    };
+  } catch (error) {
+    console.warn("Error populating app metadata", error);
```
lukaisailovic, line 142: Should we use the console directly or the logger package?

ganchoradkov: there is no logger instance in this context, I can pass it as an argument but then it can be silenced via logger levels. While this is not an error, should be visible so devs can take action

lukaisailovic (approved):

PR #6272, opened 2025-04-29
[DO NOT MERGE] fix: prevent network modal from opening randomly on reconnect

0xmkh:
## Description

When attempting to reconnect after disconnecting, the EP opens the appkit account modal view. This happens because the appkit state is not reset after disconnection.

## Type of change

## How has this been tested?

Tested on Uniswap and locally

https://github.com/user-attachments/assets/afdf678d-6c6e-474c-bce3-56676438b31c

## Checklist

### DIFF
--- a/providers/ethereum-provider/src/EthereumProvider.ts
+++ b/providers/ethereum-provider/src/EthereumProvider.ts
@@ -619,6 +619,7 @@ export class EthereumProvider implements IEthereumProvider {
           ...options,
           universalProvider: this.signer as any,
           manualWCControl: true,
+          enableNetworkSwitch: false,
         });
       } catch (e) {
         throw new Error("To use QR modal, please install @reown/appkit package");


tomiir (approved):

tomiir:
Missing changesets 🙏

providers/ethereum-provider/src/EthereumProvider.ts
```diff
@@ -618,6 +632,7 @@ export class EthereumProvider implements IEthereumProvider {
         appKit = createAppKit({
           ...options,
           universalProvider: this.signer as any,
+          enableNetworkSwitch: false,
```
magiziz, line 635: This is needed to be extra careful that we don't open "Network Switch" modal if users switch to an unsupported network

providers/ethereum-provider/src/EthereumProvider.ts
```diff
@@ -406,10 +406,24 @@ export class EthereumProvider implements IEthereumProvider {
...
   }
 
+  private resetModalState(namespaces: string[]) {
+    const deduplicatedNamespaces = [...new Set(namespaces)];
+    Object.keys(deduplicatedNamespaces).forEach((namespace) => {
+      this.modal?.resetNetwork(namespace);
+      this.modal?.resetAccount(namespace);
+    });
+    this.modal?.resetWcConnection();
+  }
```
magiziz, line 416: Calling `this.modal.disconnect()` caused several issues because the appkit account state wasn’t fully synchronized which led to strange errors. This solution works for now to at least fix the bug.

0xmkh:
Will try to do this on the appkit level

providers/ethereum-provider/src/EthereumProvider.ts
```diff
@@ -619,6 +619,7 @@ export class EthereumProvider implements IEthereumProvider {
           ...options,
           universalProvider: this.signer as any,
           manualWCControl: true,
+          enableNetworkSwitch: false,
```
0xmkh, line 622: The fix is done on the appkit side. but setting `enableNetworkSwitch` to false will make sure the network modal doesn’t open just in case appkit has any issues again

PR #6778, opened 2025-06-24
fix: duplicate `session_request` after client restart

ganchoradkov:
## Description
Fixed a bug where duplicate `session_request` was emitted after client restart affecting `sign-client` instances with disabled session request queue. 

Also implemented additional safe guards against emitting the same request multiple times 

## Type of change

## How has this been tested?
tests

## Checklist

### DIFF
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -2476,7 +2476,6 @@ export class Engine extends IEngine {
 
   private emitSessionRequest = (request: PendingRequestTypes.Struct) => {
     if (this.emittedSessionRequests.has(request.id)) {
-      console.log("skipping session_request", request.id);
       return;
     }
     this.emittedSessionRequests.add(request.id);
--- a/packages/sign-client/src/controllers/engine.ts
+++ b/packages/sign-client/src/controllers/engine.ts
@@ -105,6 +105,7 @@ import {
   getAlgorandTransactionId,
   buildSignedExtrinsicHash,
   getSignDirectHash,
+  LimitedSet,
 } from "@walletconnect/utils";
 import EventEmmiter from "events";
 import {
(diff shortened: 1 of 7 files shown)


packages/sign-client/src/controllers/engine.ts
```diff
@@ -2466,6 +2475,11 @@ export class Engine extends IEngine {
   };
 
   private emitSessionRequest = (request: PendingRequestTypes.Struct) => {
+    if (this.emittedSessionRequests.has(request.id)) {
+      console.log("skipping session_request", request.id);
```
ganchoradkov:
```suggestion
```
rm log

packages/sign-client/src/controllers/engine.ts
```diff
@@ -2466,6 +2475,10 @@ export class Engine extends IEngine {
   };
 
   private emitSessionRequest = (request: PendingRequestTypes.Struct) => {
+    if (this.emittedSessionRequests.has(request.id)) {
+      return;
```
chris13524, line 2484: Is it expected that the same request is attempted to be emitted multiple times? Maybe a WARN log would be useful here

packages/utils/src/misc.ts
```diff
@@ -560,3 +560,31 @@ export function fromBase64(encodedString: string): string {
...
+    this.limit = limit;
+    this.set = new Set<T>();
+  }
+
+  add(item: T) {
+    if (this.set.has(item)) return;
+
+    if (this.set.size >= this.limit) {
+      // Remove the oldest entry (FIFO)
+      const firstKey = this.set.values().next().value;
```
chris13524, line 578: > The values() method ... [returns] each element in this set in insertion order.

Wow, crazy that JS retains order of set members!

packages/utils/src/misc.ts
```diff
@@ -560,3 +560,31 @@ export function fromBase64(encodedString: string): string {
...
+
+  constructor({ limit }: { limit: number }) {
+    this.limit = limit;
+    this.set = new Set<T>();
+  }
+
+  add(item: T) {
+    if (this.set.has(item)) return;
+
+    if (this.set.size >= this.limit) {
```
chris13524, line 576: You're doing `>` here, so why not `while` as well? Hopefully only `==` should pass anyway, but still :D

chris13524 (approved):
Thanks!!!

packages/sign-client/src/controllers/engine.ts
ganchoradkov: added

packages/utils/src/misc.ts
ganchoradkov: probably `while` is an overkill for what it does currently

PR #6814, opened 2025-07-23
fix: set default chain unhandled exception

ganchoradkov:
## Description
Fixe unhandled exception when setting default chain of a generic sub provider. It was caused due to incorrect usage of the `this.name` variable that holds correct namespace name in other subproviders but not in the `generic` one.
context: https://reown-inc.slack.com/archives/C04DB2EAHE3/p1753257987706779

Also reworked the general generic provider usage such that new instance is created for each namespace rather than reusing a single instance for all namespaces missing a dedicated subprovider - it simplifies usage and maintainability 

## Type of change

## How has this been tested?
tests
## Checklist

### DIFF
--- a/providers/universal-provider/src/UniversalProvider.ts
+++ b/providers/universal-provider/src/UniversalProvider.ts
@@ -218,12 +218,7 @@ export class UniversalProvider implements IUniversalProvider {
       if (!this.session) return;
       const [namespace, chainId] = this.validateChain(chain);
       const provider = this.getProvider(namespace);
-      // @ts-expect-error
-      if (provider.name === GENERIC_SUBPROVIDER_NAME) {
-        provider.setDefaultChain(`${namespace}:${chainId}`, rpcUrl);
-      } else {
-        provider.setDefaultChain(chainId, rpcUrl);
-      }
+      provider.setDefaultChain(chainId, rpcUrl);
     } catch (error) {
       // ignore the error if the fx is used prematurely before namespaces are set
       if (!/Please call connect/.test((error as Error).message)) throw error;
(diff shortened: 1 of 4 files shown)


providers/universal-provider/src/providers/generic.ts
```diff
@@ -27,6 +27,7 @@ class GenericProvider implements IProvider {
     this.events = getGlobal("events");
     this.client = getGlobal("client");
     this.chainId = this.getDefaultChain();
+    this.name = this.getNamespaceName();
```
tomiir, line 30: how is this different from `opts.namespace`?

providers/universal-provider/src/UniversalProvider.ts
```diff
@@ -385,13 +380,9 @@ export class UniversalProvider implements IUniversalProvider {
...
-          if (!this.rpcProviders[GENERIC_SUBPROVIDER_NAME]) {
-            this.rpcProviders[GENERIC_SUBPROVIDER_NAME] = new GenericProvider({
-              namespace: combinedNamespace,
-            });
-          } else {
-            this.rpcProviders[GENERIC_SUBPROVIDER_NAME].updateNamespace(combinedNamespace);
-          }
+          this.rpcProviders[namespace] = new GenericProvider({
+            namespace: combinedNamespace,
+          });
```
tomiir, line 385: will this break backwards compatibility?

tomiir (approved):

providers/universal-provider/src/UniversalProvider.ts
ganchoradkov: nope, its internal usage

PR #6926, opened 2025-09-12
chore: replaced viem dependency with ox

ignaciosantise:
## Description
Viem is a required dependency of the utils package, but we only use it for its `recoverAddress` function. `ox` package also contains this function same function.

What we win by replacing it:
* `ox` is a lighter package (and also internally used by Viem)
  * NPM Unpacked Size comparison: 17MB vs 5.56MB
* Reduce versioning conflicts with other packages that also use Viem. (e.g. AppKit already has a fixed version of Viem, so finally the devs will have 2 or more versions of Viem installed)

Note: I tried moving Viem as a devDependency only to use its types, but this increases the bundle size for UMD consumers. So I just copied the used types and removed it completely

## Type of change

## How has this been tested?

Tests are passing -> `cd packages/utils/ && npm run test:run -- --reporter=verbose test/signatures.spec.ts -t "EIP-191"`

## Fixes/Resolves (Optional)

* Version conflict issues on React Native and Web

## Checklist


ganchoradkov:
>Note2: A EIP-1271 signatures test is failing, but it's not related to this PR as it also fails on main branch

where I can see the failure? The test _should_ be passing just fine

ganchoradkov (approved):
💯 💯 💯

PR #7163, opened 2026-02-10
fix: handle batch subscribe failure and trigger transport restart

ganchoradkov:
## Summary

- Fix silent error swallowing in `rpcBatchSubscribe` that caused the SDK to mark topics as subscribed when the RPC actually failed
- Keep failed topics in `pending` for automatic retry via heartbeat
- Add `connection_stalled` listener in the relayer to trigger `restartTransport` with exponential backoff

## Problem

When `rpcBatchSubscribe` failed (dead socket, relay error, timeout), the error was silently swallowed by a `.catch().then(resolve)` promise chain. `batchSubscribe` then unconditionally called `onBatchSubscribe`, which:
1. Added topics to `subscriptions` and `topicMap` as if they were active
2. Removed them from `pending`, preventing heartbeat-based retry
3. Made `isSubscribed(topic)` return `true` for topics never subscribed on the relay

## Solution

```mermaid
sequenceDiagram
    participant Sub as Subscriber
    participant Rel as Relayer
    participant Relay as RelayServer

    Sub->>Rel: batchSubscribe -> relayer.request()
    Rel->>Relay: irn_batchSubscribe
    Relay-->>Rel: ERROR
    Rel-->>Sub: rejects

    Note over Sub: rpcBatchSubscribe returns false
    Note over Sub: Topics added to pending (not subscriptions)

    Sub->>Rel: emit connection_stalled
    Rel->>Rel: restartTransport() with backoff
    Rel->>Relay: resetTransport (disconnect + subscriber.stop) then transportOpen
    Relay-->>Rel: connected
    Note over Rel: Backoff persists across reconnects

    Rel->>Sub: subscriber.start() -> resubscribe cached
    Note over Sub: Heartbeat also retries pending topics
```

### Changes

**`subscriber.ts`**
- `rpcBatchSubscribe` returns `Promise<boolean>` -- properly rejects on RPC failure instead of silently resolving
- `batchSubscribe` checks return value -- on failure, adds topics to `pending` for retry instead of registering them as subscribed

**`relayer.ts`**
- New `connection_stalled` event listener triggers `restartTransport()` with true exponential backoff
- Backoff: 0ms (immediate) -> 2s -> 4s -> 8s -> 16s -> 30s cap
- `stalledRestartInProgress` guard prevents concurrent restarts -- stays active until `restartTransport` completes (set in `finally` block)
- Backoff intentionally persists across successful reconnects -- resetting on connect would cause an infinite restart loop when subscriptions keep failing after socket reconnection
- Backoff resets only on explicit `transportClose` (clean lifecycle boundary)
- Cleanup in `transportClose` clears pending timeout, guard, and backoff counter
- Extracted `resetTransport()` -- shared disconnect + cleanup used by both `transportClose` and `restartTransport`, separating explicit-close state resets from internal restart path
- `restartTransport` calls `resetTransport()` (not `transportClose()`) to preserve stalled-restart backoff state across restart cycles
- `resetTransport` sets `reconnectInProgress = true` before `transportDisconnect` to prevent `onProviderDisconnect` from scheduling a competing reconnect
- `resetTransport` calls `subscriber.stop()` explicitly to cache active subscriptions and clear stale state, since the `reconnectInProgress` guard prevents `onProviderDisconnect` from doing so

## Test plan

- [x] Batch subscribe failure does NOT mark topics as subscribed
- [x] Batch subscribe failure adds topics to `pending` for retry
- [x] Batch subscribe failure emits `connection_stalled`
- [x] Batch subscribe success still registers subscriptions correctly
- [x] All tests validate `irn_batchSubscribe` RPC was actually called with correct params
- [x] First `connection_stalled` triggers immediate restart
- [x] Transport explicitly closed ignores `connection_stalled`
- [x] Rapid emissions only trigger one restart (guard)
- [x] Exponential backoff on subsequent emissions (2s delay verified)
- [x] Backoff does NOT reset on successful connection (prevents infinite loop)
- [x] Backoff resets after explicit `transportClose`
- [x] Test cleanup calls `transportClose` unconditionally to flush pending timers

---

> [!NOTE]
> **Medium Risk**
> Touches relayer transport lifecycle and reconnection behavior; incorrect backoff/guarding could cause reconnect loops, missed reconnects, or delayed message delivery under network instability.
> 
> **Overview**
> Fixes `Subscriber` batch subscribe so RPC failures are no longer silently swallowed: `rpcBatchSubscribe` now fails explicitly, and `batchSubscribe` keeps failed topics in `pending` (instead of registering them as subscribed) so heartbeat can retry.
> 
> Adds `Relayer` handling for `RELAYER_EVENTS.connection_stalled` to restart the transport with exponential backoff and a single-restart guard; transport shutdown/restart is refactored via a shared `resetTransport()` and `transportClose()` now clears stalled-restart timers/backoff. Tests are added to cover batch-subscribe failure/success behavior and the stalled-restart backoff semantics.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 41d210b1629b6e5be3c7bdb74b8a5db7b3306272. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/fix-batch-subscribe-error-handling.md
+++ b/.changeset/fix-batch-subscribe-error-handling.md
@@ -1,0 +1,5 @@
+---
+"@walletconnect/core": patch
+---
+
+Fixed batch subscribe silently swallowing errors and marking topics as subscribed when the RPC failed. Failed topics are now kept in pending for retry, and connection_stalled triggers a transport restart with exponential backoff.
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -319,11 +319,14 @@ export class Relayer extends IRelayer {
   }
 
   private async resetTransport() {
+    // Guards must be set synchronously before any async work:
+    // - reconnectInProgress prevents onProviderDisconnect from scheduling a competing reconnect
+    // - clearTimeout prevents a pending reconnectTimeout from firing during transportDisconnect
     this.reconnectInProgress = true;
-    await this.transportDisconnect();
-    await this.subscriber.stop();
     clearTimeout(this.reconnectTimeout);
     this.reconnectTimeout = undefined;
+    await this.transportDisconnect();
+    await this.subscriber.stop();
     this.reconnectInProgress = false;
   }
 
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -97,6 +97,11 @@ export class Relayer extends IRelayer {
   private reconnectInProgress = false;
   private requestsInFlight: string[] = [];
   private connectTimeout = toMiliseconds(ONE_SECOND * 15);
+  private stalledRestartInProgress = false;
+  private stalledRestartTimeout: NodeJS.Timeout | undefined;
+  private stalledRestartBackoff = 0;
+  private readonly stalledRestartBaseInterval = toMiliseconds(ONE_SECOND * 2);
+  private readonly stalledRestartMaxInterval = toMiliseconds(THIRTY_SECONDS);
   constructor(opts: RelayerOptions) {
     super(opts);
     this.core = opts.core;
@@ -273,6 +278,8 @@ export class Relayer extends IRelayer {
     clearTimeout(this.reconnectTimeout);
     this.reconnectTimeout = undefined;
     this.reconnectInProgress = false;
+    clearTimeout(this.stalledRestartTimeout);
+    this.stalledRestartInProgress = false;
     await this.transportDisconnect();
   }
 
@@ -567,6 +574,7 @@ export class Relayer extends IRelayer {
 
   private onConnectHandler = () => {
     this.logger.warn({}, "Relayer connected 🛜");
+    this.stalledRestartBackoff = 0;
     this.startPingTimeout();
     this.events.emit(RELAYER_EVENTS.connect);
   };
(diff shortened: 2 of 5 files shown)


ganchoradkov:
@claude review

ganchoradkov:
## Addressed all review feedback

### Copilot comments

| Comment | Status |
|---------|--------|
| `stalledRestartInProgress` reset before `restartTransport` finishes | Fixed in `bb73456` — moved reset to `finally` block |
| `transportClose` doesn't reset `stalledRestartBackoff` | Fixed in `bb73456` — backoff resets in `transportClose` |
| `registerEventListeners` is async but `init` doesn't await it | Pre-existing pattern — not introduced by this PR, out of scope |
| Tests use real `setTimeout` delays, prefer fake timers | Acknowledged — the 2.1s wait is the only slow test and validates real timing behavior; can optimize in a follow-up |

### cursor[bot] comments

| Comment | Severity | Status |
|---------|----------|--------|
| Linear backoff instead of exponential | Medium | Fixed in `51c33bd` — now uses `Math.pow(2, backoff)` for true exponential (0→2s→4s→8s→16s→30s cap) |
| Backoff reset on connect causes infinite restart loop | High | Fixed in `bb73456` — removed backoff reset from `onConnectHandler` |
| Backoff state reset mid-restart via `transportClose` call | High | Fixed in `af48f39` — extracted `resetTransport()` so `restartTransport` no longer calls `transportClose()` |
| Missing guard allows spurious reconnect during restart | Medium | Fixed in `9294d0c` — `resetTransport` sets `reconnectInProgress = true` before disconnect |
| `resetTransport` skips `subscriber.stop()` | High | Fixed in `95c1ee1` — `resetTransport` now calls `subscriber.stop()` explicitly |

### github-actions[bot] (Claude Auto Review) comments

| Comment | Severity | Status |
|---------|----------|--------|
| Linear backoff instead of exponential | Medium | Fixed in `51c33bd` (same as cursor[bot] finding) |
| Missing guard causes spurious reconnect | High | Fixed in `9294d0c` + `95c1ee1` (same as cursor[bot] finding) |
| Subscriber state cleanup skipped during restart | High | Fixed in `95c1ee1` (same as cursor[bot] finding) |

All high-severity issues have been resolved. The only items not addressed are pre-existing patterns (async `registerEventListeners`) and style preferences (fake timers), both of which are out of scope for this PR.

ganchoradkov:
@claude review

PR #7195, opened 2026-03-12
Feat: Change log message

dsshard:
In React Native, using warn triggers a toast that appears at the bottom of the screen. During connect, this becomes very annoying. Changing logger.warn to logger.info will remove unnecessary noise from the screens during development.


dsshard:

dsshard:
recheck

PR #7206, opened 2026-03-26
fix: prevent infinite loop and memory exhaustion in relayer reconnection

### LINKED ISSUE #7131: Infinite loop when reconnecting causes memory exhaustion
In [Cashonize](https://cashonize.com/) Bitcoin cash wallet which uses WalletConnect with a custom network for BCH, after some combination of the tab being backgrounded and network being interrupted, an infinite loop causes a rapid build up of errors until memory is exhausted with the browser tab exceeding 4GB and millions event listerners.

**SDK Version (if relevant)**
 - Client: Javascript
 - Version: 2.21.8

**To Reproduce**
Working on reproduction.

**Expected behavior**
Wallet and app reconnect after a network interruption without excessive resource utilisation.

**Video and screenshot**

https://github.com/user-attachments/assets/ab86743d-7624-4498-9754-786d2e717706

Image

 - OS: MacOS 14.7.4
 - Browsers: Brave 1.85.20, Chrome 143.0.7499

Debugging indicated a flood of calls to `Relayer.connect()` in [`relayer.ts`](https://github.com/WalletConnect/walletconnect-monorepo/blob/5f5702bd9ad87ef998dbc32d823dffe4764a264b/packages/core/src/controllers/relayer.ts#L356). This class has complex retry logic with loops, recursion and circular dependencies which have the potentional to create large numbers of websocket connections and events listeners.

The `core.crypto.signJWT(this.relayUrl)` makes each attempt expensive and there are hundreds to thousands of attempts per second.

Image

Image

ganchoradkov:
## Summary

Fixes #7131 — Infinite loop when reconnecting causes memory exhaustion.

When the network is unreachable but `isOnline()` returns `true` (e.g. WiFi connected but broadband down, or Brave blocking the API), multiple interacting bugs in the relayer's `connect()` method combine to create **exponential growth** of concurrent connection attempts (5 → 25 → 125 → 625+), each calling the expensive `signJWT()`, until the browser tab exceeds 4GB and millions of event listeners accumulate.

### Root causes fixed

- **`new Promise(async executor)` antipattern** — After `reject()`, `.catch()` swallows the error and execution continues to `subscriber.start()`, which runs as an unsupervised background task that eventually triggers more `connect()` calls via `connection_stalled`
- **`connectionAttemptInProgress` reset inside retry loop** — The `finally` block reset the guard on every iteration instead of after the loop, allowing `restartTransport()` to spawn concurrent `connect()` calls during the sleep between retries
- **`reconnectInProgress` stuck forever** — Early returns in `onProviderDisconnect()` (no topics / transport closed) exited without resetting the flag, permanently blocking all future automatic reconnection
- **WebSocket leak in `createProvider()`** — Old provider connections were never closed before creating new ones, leaking a WebSocket + 4 event listeners per retry attempt
- **`toEstablishConnection()` bypassed serialization** — Called `connect()` directly instead of going through `transportOpen()` which has the `connectPromise` guard, enabling concurrent unserialized connections from subscriber internals

```mermaid
flowchart TD
    A[connect attempt fails] -->|"Bug 1: async executor"| B[subscriber.start runs in background]
    B --> C[subscriber request fails]
    C --> D[emits connection_stalled]
    D -->|"Bug 2: guard reset"| E[restartTransport passes guard]
    E --> F[new connect call spawned]
    F --> A
    A -->|"Bug 4: WS leak"| G[old socket leaked]
    F -->|"Bug 5: no serialization"| H[concurrent connect calls]
    
    style A fill:#ff6b6b
    style F fill:#ff6b6b
    style H fill:#ff6b6b
```

### Changes

| Fix | File | Description |
|-----|------|-------------|
| 1 | `relayer.ts` | Split nested async executor into two sequential properly-constructed promises |
| 1b | `relayer.ts` | Cleaned up `transportOpen()` — removed unnecessary `new Promise(async ...)` wrapper |
| 2 | `relayer.ts` | Moved `connectionAttemptInProgress = false` to outer `finally` after the while loop |
| 3 | `relayer.ts` | Reset `reconnectInProgress` before early returns in `onProviderDisconnect()` |
| 4 | `relayer.ts` | Close old provider connection in `createProvider()` before creating new one |
| 5 | `relayer.ts` | Route `toEstablishConnection()` through `transportOpen()` for `connectPromise` serialization |

### Tests added

- `reconnectInProgress` resets when no topics exist on disconnect
- `reconnectInProgress` resets when transport is explicitly closed
- Reconnection still works after early return from `onProviderDisconnect`
- `restartTransport` is blocked during active connection attempts

## Test plan

- [ ] Existing relayer tests pass
- [ ] New guard tests verify the concurrency fixes
- [ ] Manual test: background tab + network interruption no longer causes memory exhaustion
- [ ] Manual test: reconnection still works correctly after network recovery

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/fix-relayer-reconnect-infinite-loop.md
+++ b/.changeset/fix-relayer-reconnect-infinite-loop.md
@@ -1,0 +1,13 @@
+---
+"@walletconnect/core": patch
+---
+
+fix: prevent infinite loop and memory exhaustion in relayer reconnection
+
+- Eliminate `new Promise(async executor)` antipattern in `connect()` and `subscribe()` so background tasks no longer spawn unsupervised connection attempts
+- Serialize connection attempts via `connectPromise` in both `transportOpen()` and `toEstablishConnection()` to prevent concurrent `connect()` calls
+- Keep `connectionAttemptInProgress` guard up during the entire retry loop to block `restartTransport()` from spawning parallel connections
+- Reset `reconnectInProgress` on early returns in `onProviderDisconnect()` to prevent the flag from getting permanently stuck
+- Close old WebSocket in `createProvider()` before creating a new one to prevent connection and listener leaks
+- Wrap `subscriber.stop()` in try/catch in `onProviderDisconnect()` to prevent `reconnectInProgress` from getting stuck on throw
+- Reset `stalledRestartBackoff` on successful connection so recovery latency doesn't degrade after repeated stall/recover cycles
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -426,10 +426,7 @@ export class Relayer extends IRelayer {
       }
 
       if (this.connected) {
-        this.logger.debug(
-          {},
-          `Connected to ${this.relayUrl} successfully on attempt: ${attempt}`,
-        );
+        this.logger.debug({}, `Connected to ${this.relayUrl} successfully on attempt: ${attempt}`);
         break;
       }
 
--- a/packages/core/src/controllers/relayer.ts
+++ b/packages/core/src/controllers/relayer.ts
@@ -147,8 +147,7 @@ export class Relayer extends IRelayer {
   get connecting() {
     return (
       // @ts-expect-error
-      this.provider?.connection?.socket?.readyState === 0 ||
-      this.connectPromise !== undefined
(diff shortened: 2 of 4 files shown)


ganchoradkov:
@claude review

ganchoradkov:
@claude review

ganchoradkov:
@claude review

PR #7216, opened 2026-04-06
fix(ethereum-provider): validate projectId on init

Zelys-DFKH:
## Summary

Apps that misconfigure projectId as an empty string or whitespace fail silently. The QR modal never appears, with no error message. This PR adds early validation in EthereumProvider.init() to throw a descriptive error immediately, pointing developers to cloud.walletconnect.com to get a projectId.

## Changes

- Validate projectId in EthereumProvider.static init() — throw if empty, null, or whitespace-only
- Add test cases for empty string, whitespace, and valid projectId
- Add changeset for patch version bump

## Why

The silent failure mode is confusing to debug. A developer misconfigures the env var, the QR modal doesn't show, and they spend 30 minutes wondering why without any error in the console. This change catches the mistake at init time with clear guidance.

### DIFF
--- a/.changeset/ethereum-provider-projectid-validation.md
+++ b/.changeset/ethereum-provider-projectid-validation.md
@@ -1,0 +1,5 @@
+---
+@walletconnect/ethereum-provider: patch
+---
+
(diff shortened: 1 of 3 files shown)


Zelys-DFKH:

PR #7247, opened 2026-05-27
fix: reject SIWE statements containing line breaks per EIP-4361

ganchoradkov:
## Summary

`formatMessage` in `packages/utils/src/cacao.ts` builds the SIWE/CAIP-122 message by placing the `statement` between the wallet-address block and the `URI:`/`Version:`/`Chain ID:`/`Nonce:`/… fields, then joining the whole array with `\n`. Per [EIP-4361](https://eips.ethereum.org/EIPS/eip-4361) the `statement` is a single line (ABNF: `reserved / unreserved / " "`), but `cacao.statement` was never validated for line breaks.

A statement containing `\r`/`\n` could therefore **forge additional SIWE fields** in the message that gets signed/verified.

### The vector

```
statement: "I accept\nURI: https://evil.com\nNonce: attacker"
```

…would inject forged `URI:`/`Nonce:` lines directly into the signed message.

### Fix

Guard added right after reading `cacao.statement`, before any field assembly — throws if the statement contains `\r` or `\n`.

```mermaid
flowchart TD
    A[cacao.statement] --> B{contains \r or \n?}
    B -- yes --> C[throw Error]
    B -- no --> D[assemble message lines]
    D --> E[join with \n -> signed message]
```

## Changes
- `packages/utils/src/cacao.ts` — reject line breaks in statement
- `packages/utils/test/cacao.spec.ts` — tests for `\n`, `\r\n` injection + valid single-line statement
- changeset (`@walletconnect/utils` patch)

## Test plan
- [x] `npx vitest run --dir test cacao` in `packages/utils` — 21/21 pass

## Note for reviewers
`validateSignedCacao` reconstructs the message via `formatMessage` to verify signatures, so an already-stored cacao with a newline statement will now **throw** during verification rather than returning `false`. That's the safe outcome (such a cacao should never have been signed), but happy to wrap the call site to fail gracefully instead if preferred.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/reject-statement-line-breaks.md
+++ b/.changeset/reject-statement-line-breaks.md
@@ -1,0 +1,5 @@
+---
+"@walletconnect/utils": patch
+---
+
+Reject SIWE/CAIP-122 statements containing line breaks in `formatMessage`. Per EIP-4361 the statement is a single line, and embedded `\r`/`\n` could forge other fields (URI, Nonce, etc.) in the signed message.
--- a/.changeset/reject-statement-line-breaks.md
+++ b/.changeset/reject-statement-line-breaks.md
@@ -2,4 +2,4 @@
 "@walletconnect/utils": patch
 ---
 
-Reject SIWE/CAIP-122 statements containing line breaks in `formatMessage`. Per EIP-4361 the statement is a single line, and embedded `\r`/`\n` could forge other fields (URI, Nonce, etc.) in the signed message.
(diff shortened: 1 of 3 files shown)


ganchoradkov:
Thanks for the reviews — addressed all four points in `8857209c9`:

**Copilot — recap-derived newlines bypass the guard** ✅
Correct catch. `isValidRecap` doesn't reject `\r`/`\n` in resource/ability names, so `formatStatementFromRecap` could reintroduce line breaks *after* the original guard ran. Moved the check to validate the **final** statement, after recap formatting. A regression test now feeds a recap whose resource name contains `\n` and asserts `formatMessage` throws.

**Claude — `validateSignedCacao` now throws instead of returning a boolean** ✅
Wrapped the `formatMessage` call in `validateSignedCacao` in a `try/catch` that returns `false`, preserving its boolean contract for legacy/attacker-crafted CACAOs.

**Claude / Copilot — missing test coverage** ✅
Added two tests: `validateSignedCacao` returns `false` for a newline statement, and the recap-derived newline case above.

Changeset expanded to note the recap path and the `validateSignedCacao` behavior. 23/23 tests passing.

PR #7311, opened 2026-08-09
fix(utils): reject CR/LF in CACAO single-line SIWE fields

SashaMIT:
## Summary
- After #7247, `formatMessage` rejects line breaks in the final statement, but other EIP-4361 single-line fields (`domain`, `aud`/`uri`, `version`, `nonce`, `iat`, optional `exp`/`nbf`/`requestId`) were still interpolated unchanged.
- A crafted `domain` like `example.com\nURI: https://evil.com` forges additional message lines (same class as statement smuggling).
- Fail closed: reject `\r`/`\n` in those fields before building the message.

## Test plan
- [x] Extended `packages/utils/test/cacao.spec.ts` for domain / aud / nonce CR/LF rejects
- [x] Existing statement + recap linebreak tests still match the shared error prefix
- [ ] CI `@walletconnect/utils` vitest

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/utils/src/cacao.ts
+++ b/packages/utils/src/cacao.ts
@@ -104,12 +104,25 @@ export const formatMessage = (cacao: AuthTypes.FormatMessageParams, iss: string)
     statement = formatStatementFromRecap(statement, decoded);
   }
 
-  // Per EIP-4361 the statement is a single line and must not contain line breaks.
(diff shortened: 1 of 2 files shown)


SashaMIT:

SashaMIT:
recheck

PR #7345, opened 2026-09-14
fix(pay): accept interface-typed objects in ConfirmPaymentParams.data

jakubuid:
## Why

`@walletconnect/pay@1.1.0` (#7344) introduced `ConfirmPaymentParams.data` typed as `(string | Record<string, unknown> | unknown[])[]`. TypeScript only grants the implicit index signature `Record<string, unknown>` requires to object-literal types — **not to interfaces** — so a wallet passing its own result type (e.g. TRON's `interface TronSignedTransaction { raw_data_hex: string; signature: string[] }`) gets `TS2345: Index signature for type 'string' is missing`. The web wallet hit exactly this in reown-com/web-examples#1049 and had to convert its interface to a type alias.

## What

- `data?: (string | object)[]` — `object` accepts interfaces, plain objects and arrays alike; strings stay first-class. **No runtime change**: `buildConfirmPaymentRequest` still forwards elements verbatim.
- Regression test passing an `interface`-typed value through `buildConfirmPaymentRequest` (proven at the type level with an ad-hoc `tsc --noEmit` over the test, since the package tsconfig only covers `src/`).
- Sync `SDK_VERSION` constant with the released 1.1.0 (the version-packages commit left it at 1.0.10; the build script re-syncs it anyway).
- Changeset: `@walletconnect/pay` **patch** → 1.1.1.

## Verification

- `npm run build` (tsc + rollup) clean; vitest 82/82 passing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jakubuid:
Closing for now — not needed functionally: pay 1.1.0 already forwards objects at runtime; this was only a TypeScript ergonomics improvement (interfaces vs `Record<string, unknown>`). Integrators can declare their result as a `type` alias or spread it. Can be revisited if it becomes a recurring support question.

